PR #12, opened 2021-11-03
fix: gradient clipping of FP16 optimizer

ver217:
All-reduce over tensor parallel nodes should be ignored when there are no tensor parallel parameters.


FrankLeeeee:
Re-PR with a different source branch

PR #71, opened 2021-12-15
add interleaved pipeline

ver217:
Add interleaved pipeline. Usage:

```python
from colossalai.builder import PipelineModelInitializer
from colossalai.engine.schedule import InterleavedPipelineSchedule

model = PipelineModelInitializer(model_cfg, num_model_chunks).initialize()
if isinstance(model, list):
        model = torch.nn.ModuleList(model)

schedule = InterleavedPipelineSchedule(num_microbatches, num_model_chunks)
```

### DIFF
--- a/colossalai/builder/pipeline.py
+++ b/colossalai/builder/pipeline.py
@@ -1,6 +1,7 @@
 import copy
 import heapq
 
+
 from colossalai.builder import build_model, build_layer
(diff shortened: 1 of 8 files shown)


FrankLeeeee:
@ver217 can convert list to module list inside the initialize function?

ver217:
> @ver217 can convert list to module list inside the initialize function?

I will do so later.

PR #72, opened 2021-12-16
add interleaved pipeline (#71)

ver217:
Usage:

```python
from colossalai.builder import PipelineModelInitializer
from colossalai.engine.schedule import InterleavedPipelineSchedule

model = PipelineModelInitializer(model_cfg, num_model_chunks).initialize()
if isinstance(model, list):
        model = torch.nn.ModuleList(model)

schedule = InterleavedPipelineSchedule(num_microbatches, num_model_chunks)
```

### DIFF
--- a/colossalai/amp/naive_amp/_fp16_optimizer.py
+++ b/colossalai/amp/naive_amp/_fp16_optimizer.py
@@ -58,7 +58,8 @@ class DynamicGradScaler:
                  backoff_factor,
                  growth_interval,
                  hysteresis,
-                 max_scale: int = None):
+                 max_scale: int = None,
+                 verbose: bool = False):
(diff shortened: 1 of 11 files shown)


FrankLeeeee:
I will take a look in the afternoon, but the example stills need convert to module list manually, wonder if this is a must?

ver217:
> I will take a look in the afternoon, but the example stills need convert to module list manually, wonder if this is a must?

Maybe yes. We need to keep compatibility, so that the model which engine receives shoud be a torch.nn.Module. And If not, users cannot call model.parameters() directly when initializing optimizer. For simplicity, I just convert the list of model to torch.nn.ModuleList. I will refactor the PipelineModelInitializer later, so that users never need to do the convert manually.

PR #73, opened 2021-12-16
overlap computation and communication in 2d operations

1SAA:
- Improved performance by overlaping computation and communication in 2d operations


FrankLeeeee (changes requested):
Hi Haichen,

In `_operation.py`, the size of the list is hard-coded as 2. This will raise errors if tensor parallel size is not 4.

Example: 
```python
line 94: A_list = [torch.empty_like(A) for _ in range(2)]
```

PR #83, opened 2021-12-21
Layer integration

kurisusnowdeng:
- Integrated parallel layers for ease of building models
- Renamed few apis
- Reworked metric (e.g. accuracy) and moved it into `colossalai.nn`
- Added throughput hook, log metric by step hook
- Reworked initialization
- Reworked `model_zoo`, with only one implementation for each model now.
- Added `benchmark`, including cifar10 and imagenet100 training scripts
- Updated related unit tests by replacing models from config with models in `model_zoo`
- Tested all parallel layers and cifar10 convergence


FrankLeeeee:
@kurisusnowdeng The checks are not passed

FrankLeeeee (changes requested):
Hi, @kurisusnowdeng ,  thanks for your contribution, I think there are quite a lot of issues worth discussion and I have listed them down below and welcome your opinions.

`benchmark/cifar/train.py` & `benchmark/imagenet100/train.py` & `benchmark/imagenet1k/train.py`

1. at line 54, if you want to leave commented `colossalai.launch ` code in this file, you should add a comment to explain why you to keep this part for clarity (this serves as an alternative for setting up the environment). I believe this could better help with readability. For example:
```python
args = colossalai.get_default_parser().parse_args()
# if you want to use torch.distributed.launch or torchrun
colossalai.launch_from_torch(config=args.config)

# if you are using a general launcher, you can use the code below
# colossalai.launch(config=args.config,
#                   rank=args.rank,
#                   world_size=args.world_size,
#                   local_rank=args.local_rank,
#                   host=args.host,
#                   port=args.port)
```

2. why is vit named as `vit_lite_7_patch4_32`? `7` ’s meaning is not clear.
3. When calculating the batch size of the dataloader by `gpc.config.BATCH_SIZE // gpc.data_parallel_size`, it ignores the gradient accumulation setting?

`colossalai/communication/__init__.py`
1. please adjust your vscode setting to max line width 120 as the format change is not  meaning commit

`colossalai/communication/collective.py`
1. this file has quite a large portion of commented code. If they are not used, they should be removed.

`colossalai/initialize.py`
1. I would discourage the addition of `MASTER_ADDR` and `MASTER_PORT` as these variables are not set by torch.distributed.launch but by the user instead. It is essentially the same for the user to input this parameter to `launch_from_torch` or export to environment. And this change will require refactoring markdown doc, web doc and examples folder.


`colossalai/nn/init.py`
1. Giving the argument `init_method` a default value None is not a good idea. As if the user does not provide a specific value for `init_method`, essentially nothing will do and the user may not be aware of it. The two init functions do not have assert statements too. The `init_method` parameter should be only within a list of choices.
2.  should not let user input `fan_in` and `fan_out` as most users do not know what they mean
3. I would prefer to split this function into different functions. E.g. `init.zero_`, `init.normal_`. Jax is not a method of initiailzation to me.


`colossalai/nn/layer/__init__.py`
1. for the linear layer, some parameters are missing, i.e. `skip_bias_add`. This parameter is for operation fusion to gain some performance. If using tensor_parallel is None, then the torch native linear does not support skip bias add, so better wrap the native linear again. So that regardless of `tensor_parallel` value, skip bias add can be always enabled if set to True.


`colossalai/nn/metric/__init__.py`
1. shifting metric to nn module will affect markdown and rst docs as well as the examples, please make sure they are refactored too.

`colossalai/nn/metric/accuracy_2d.py`
1. `reduce_by_batch` is an autograd function. calculating acc is not part of the forward & backward process, so no autograd function should be used. 
2. The same problem exists in other accuracy files.


`colossalai/trainer/_trainer.py`
1. many format changes due to vs code line width setting
2. not sure why changing `train-step` to `Train-step`? Seems trivial to me.

`colossalai/trainer/hooks/_log_hook.py`
1. quite a lot of format change without code change
2. This applies to other hook files as well

`colossalai/utils/memory.py`
1. the change to `full_log` at line 51 seems quite trivial?

`unit_testing`
1. The test dimensions in `common.py` should be kept small. The latest unit testing code (10-14 min)takes more than 2 times the time spent by the previous setting.
2. The origin small values look ok to me as it is enough to verify correctness

`tests/test_zero_tensor_parallel/test_vit_2d_level_3.py` & `tests/test_zero_tensor_parallel/test_vit_2d_level_2.py`
1. besides change in CrossEntropyLoss in line 65, the remaining change is only format change

FrankLeeeee:
And one more thing, `docstring` should be added to all the new classes and functions so that they can be captured by sphinx for generating web documentation.

kurisusnowdeng:
All fixed. @FrankLeeeee

FrankLeeeee:
Hi @kurisusnowdeng , I have reviewed your code and some issues remain.

`benchmark/cifar/configs/vit_1d.py`
1. the correct field name for `gradient_clipping` is `clip_grad_norm`. This applies to all other config files.


`colossalai/nn/init.py`
1. seems `fan_in` and `fan_out`  can be directly obtained from the weight shape, they should be hidden from the user.
2. `fan_in` should not be set to None as default because it cannot be default in the assert statement in `lecun` initialization.
```python
def lecun_uniform_():
    # adapted from jax.nn.initializers
    def initializer(tensor: Tensor, fan_in: int = None, fan_out: int = None):
        assert fan_in is not None, 'Fan_in is not provided.'

        var = 1.0 / fan_in
        bound = math.sqrt(3 * var)
        return nn.init.uniform_(tensor, -bound, bound)

    return initializer


def lecun_normal_():
    # adapted from jax.nn.initializers
    def initializer(tensor: Tensor, fan_in: int = None, fan_out: int = None):
        assert fan_in is not None, 'Fan_in is not provided.'

        std = math.sqrt(1.0 / fan_in)
        return nn.init.trunc_normal_(tensor, std=std / .87962566103423978)

    return initializer
```


`colossalai/nn/layer/__init__.py`
1. from line 6 to line 24, why are all the modules imported?
2. Are there reasons for setting the default initializer? e.g. kaiming for weight and xavier for bias. The user is not aware of our setting and my suggestion would be to fall back to default torch initialization.
3. I would suggest putting the unified layers into a single file instead of `__init__.py` as `__init__.py` serves for intialization/configuration purpose.

`colossalai/nn/layer/parallel_1d/_utils.py`
1. the utils functions here are for general use, not specific for 1D, I don’t think they should be put under 1D utils.

`colossalai/nn/metric/accuracy_2d.py`
1. accuracy is still using autograd function and this applies to other metrics as well.

kurisusnowdeng:
Fixed. Please check again. @FrankLeeeee

FrankLeeeee (approved):

PR #113, opened 2022-01-04
Added MoE parallel

1SAA:
- Added MoE parallel for colossal

### DIFF
--- a/colossalai/constants.py
+++ b/colossalai/constants.py
@@ -15,7 +15,8 @@ INITIALIZER_MAPPING = {
     '2.5d': 'Initializer_2p5D',
     '3d': 'Initializer_3D',
     'sequence': 'Initializer_Sequence',
-    'model': 'Initializer_Model'
+    'model': 'Initializer_Model',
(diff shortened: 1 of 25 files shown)


FrankLeeeee:
Hi @1SAA , I have reviewed your code and want raise some questions as shown below.
`colossalai/context/random/init_func.py`
1. `moe_set_seed` can be put into `colossalai/context/random/_helper.py`

`colossalai/context/parallel_mode.py`
1. is it better to name `MOE_TENSOR` as `MOE_EXPERT` as the common term in MoE is expert parallel.

`colossalai/global_variables.py`
1. Should use `expert_parallel_size` instead of `tensor_parallel_size` in MoeEnv?

`model_zoo/helper.py`
1. do not use all capitalized words for parameters and attributes

`colossalai/nn/layer/vanilla/__init__.py`
1. I would recommend to put these new layers into your moe model example in the model_zoo directly.

`README.md`
1. Can you provide a readme documentation on how the user can use moe parallel?

PR #130, opened 2022-01-07
Raise exception when missing required packages when using ZeRO

ver217:
Try import deepspeed when using zero


ver217:
Users may be confused if we don't throw an error if they want to use ZeRO and didn't install deepspeed.

FrankLeeeee (approved):

PR #137, opened 2022-01-09
resnet on imagenet dataset but trained on multiple GPU's

anushka192001:


FrankLeeeee:
Hi @anushka192001 , thanks for your pull request. We already have an example on resnet for cifar10 dataset. I think these two examples only differ in datasets so I think this pull request is not very meaningful.

FrankLeeeee:
Hi @anushka192001 , I will close this PR for now. If you have new examples, welcome to put up a PR in [ColossalAI-Examples](https://github.com/hpcaitech/ColossalAI-Examples). Feel free to open an issue or discussion if you have any query. :)

PR #153, opened 2022-01-17
add transpose_pad_fusion kernel for inference

dujiangsu:


colossalai/kernel/cuda_native/csrc/transpose_pad_fusion_kernel.cu
```diff
@@ -0,0 +1,93 @@
+// For eliminating redundant computation.
+// The core idea and code of kernel is from TurboTransformer.
```
feifeibear, line 2: Be careful. If you use the code you should use the copyright of Tencent.
You'd better polish the code and reimplement it by yourself.
You should also add the necessary unitest for your CUDA kernels.

feifeibear:
Did you benchmark the performance improvement of the MR.
benchmark results are welcome in comparison with Effective transformers or Faster Transformers?

colossalai/kernel/cuda_native/csrc/transpose_pad_fusion_kernel.cu
dujiangsu: I'll add the copyright of Tencent and convert my test into the unitest format.
Since I am working on the inference engine, I plan to apply the "transpose_pad" method as a demo policy as well as the first policy for dealing with redundant computation.

dujiangsu:
> Did you benchmark the performance improvement of the MR. benchmark results are welcome in comparison with Effective transformers or Faster Transformers?

My implementation can achieve better performance in samples with larger seq_len variance (e.g. datasets in GLUE). I am going to make a more detailed evaluation tomorrow.

feifeibear:
> > Did you benchmark the performance improvement of the MR. benchmark results are welcome in comparison with Effective transformers or Faster Transformers?
> 
> My implementation can achieve better performance in samples with larger seq_len variance (e.g. datasets in GLUE). I am going to make a more detailed evaluation tomorrow.

Thanks, post the result here if possible.

PR #178, opened 2022-01-21
Hotfix/vocab parallel layers

kurisusnowdeng:
- Moved env variables to global variables; added branch context
- Added vocab parallel layers
- Updated gpt model


FrankLeeeee (changes requested):
`colossalai/engine/schedule/_pipeline_schedule.py`
1. I am a bit confused about why TorchAMPModel added to the type check in line 99, 108, 387 and 391. I believe TorchAMP does not support pipeline yet and this PR is not related to schedule code.

`colossalai/nn/layer/colossalai_layer/embedding.py`
1. it causes confusion when patch embedding and vocab embedding are put under the same Embedding class.
2. Please add unit testing for these embedding layers to verify their correctness

`colossalai/nn/layer/colossalai_layer/linear.py`
1. This is the same as embedding

`colossalai/nn/layer/colossalai_layer/normalization.py`
1. From line 39 to 42, why only `nn.LayerNorm` is set to dtype and device?

```python
39  if tensor_parallel is None:
40              self.norm = nn.LayerNorm(normalized_shape, eps=eps).to(dtype).to(get_current_device())
41          else:
42              self.norm = _parallel_layernorm[tensor_parallel](normalized_shape, eps=eps, dtype=dtype)
```


`colossalai/nn/layer/parallel_*d/_operation.py`
1. There is a problem missed in the previous code review. Please use standard class naming convention for autograd function, i.e. camel case. This applies to operation file of all other tensor parallel code as well.
2. For the autograd function, the docstring for the forward static method is moved to the class level. I am a bit confused by this as the class-level docstring should correspond to `__init__` method.


`colossalai/nn/layer/utils/common.py`
1. In line 40, the docstring for arguments are deleted.


`colossalai/nn/loss/__init__.py`
1. I woud like to suggest that cross entropy and vocab cross entropy should not be put under the same unfied class to avoid confusion.
2. the unified loss should not be put in the `__init__.py` file as well as mentioned in previous code reviews. functions and classes in the `__init__.py` generally serve the purpose of setting up the code environment or build something upon importing this module.

FrankLeeeee:
Hi, @kurisusnowdeng . I have finished code review and listed my comments above. Meanwhile, this PR has many changes unrelated to the PR title and description such as schedule and hooks. Please split your PR into multiple PRs next time for standardisation and traceability of our code change, otherwise the PR might be rejected.

PR #244, opened 2022-02-18
add Chinese README

binmakeswell:


FrankLeeeee:
Hi, I feel the images are more like paper-style. Can you redraw the images using prettier theme for  table/plots. I think the table can be presented as a plot as well. I feel plots are generally more easy on the eyes.

PR #270, opened 2022-02-28
[WIP] initialize the shard param class

feifeibear:


ver217 (approved):

feifeibear:
A small step to reconfigure the zero hooks.
Wrap the Tensor rather than the whole module (Zero3ParameterManager). So we can shard and gather in a finer granularity!

PR #280, opened 2022-03-01
remove deepspeed implementation and refactor for the reconstructed zero module

FrankLeeeee:
I have removed deepspeed-related code and refactored unit testing and other modules which import zero.


ver217 (approved):

feifeibear (approved):
LGTM

PR #284, opened 2022-03-01
Add CPU Adam

Gy-Lu:
Added CPU Adam. For now, fp32 mode has passed the unit test, while the unit test of fp16 mode should be redesigned.


feifeibear:
merge to feature.zero not develop?

FrankLeeeee:
Hi, is this a duplicate PR of #288 . If so, I will close this PR.

PR #287, opened 2022-03-01
add sharded grad and refactor grad hooks

ver217:
Implemented a complete SharedModel FWD/BWD process
This PR uses the newly implemented hook mechanism. 
Specifically, OpHook is used for the management of param.data, and BaseParamHookMgr is used for the reduction and management of param.grad.

### DIFF
--- a/colossalai/engine/ophooks/_shard_grad_ophook.py
+++ b/colossalai/engine/ophooks/_shard_grad_ophook.py
@@ -1,0 +1,31 @@
+import torch
+from colossalai.registry import OPHOOKS
+
+from . import BaseOpHook
+
+
+@OPHOOKS.register_module
+class ShardGradHook(BaseOpHook):
+    """
+    A hook to process sharded param before and afther FWD and BWD operator executing.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def pre_fwd_exec(self, module: torch.nn.Module, *args):
+        pass
+
+    def post_fwd_exec(self, module: torch.nn.Module, *args):
+        pass
+
+    def pre_bwd_exec(self, module: torch.nn.Module, input, output):
+        for param in module.parameters():
+            assert hasattr(param, '_sharded_grad')
+            param._sharded_grad.setup()
+
+    def post_bwd_exec(self, module: torch.nn.Module, input):
+        pass
+
+    def post_iter(self):
+        pass
--- a/colossalai/zero/sharded_model/sharded_grad.py
+++ b/colossalai/zero/sharded_model/sharded_grad.py
@@ -19,6 +19,9 @@ class ShardedGradient:
                  process_group: Optional[ProcessGroup] = None,
                  offload_config: Optional[dict] = None
                  ) -> None:
+        assert hasattr(
+            param, 'ca_attr') and param.ca_attr.is_shared, 'ShardedGradient can only be initialized with sharded parameter'
+
         self.param = param
         self.process_group = process_group or gpc.get_group(ParallelMode.DATA)
         self.shard_idx = self.process_group.rank()
@@ -27,12 +30,9 @@ class ShardedGradient:
 
         self._cpu_offload = offload_config.get('device', None) == 'cpu' if offload_config else False
 
-        sharded_param, _ = get_shard(param.data, self.shard_idx, self.num_shards)
-
         self._is_sharded = self.num_shards > 1
-        self._orig_size = param.data.size()
-        self._shard_size = sharded_param.size()
-        del sharded_param
+        self._orig_size = param.ca_attr.origin_shape
+        self._shard_size = param.ca_attr.shard_shape
 
         self._saved_grad_shard: Optional[torch.Tensor] = None
         self._saved_full_grad: Optional[torch.Tensor] = None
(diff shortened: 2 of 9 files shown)


colossalai/engine/ophooks/_shard_grad_ophook.py
```diff
@@ -0,0 +1,31 @@
+import torch
+from colossalai.registry import OPHOOKS
+
+from . import BaseOpHook
+
+
+@OPHOOKS.register_module
+class ShardGradHook(BaseOpHook):
```
feifeibear, line 8: @feifeibear Need a mechanism to guarantee different execution orders. Not related to this PR

colossalai/zero/sharded_model/sharded_grad.py
```diff
@@ -0,0 +1,122 @@
...
+                 ) -> None:
+        self.param = param
+        self.process_group = process_group or gpc.get_group(ParallelMode.DATA)
+        self.shard_idx = self.process_group.rank()
+        self.num_shards = self.process_group.size()
+        self.offload_config = offload_config
+
+        self._cpu_offload = offload_config.get('device', None) == 'cpu' if offload_config else False
+
+        sharded_param, _ = get_shard(param.data, self.shard_idx, self.num_shards)
```
feifeibear, line 30: SharedParam is another Class. You do not need do it here?

colossalai/zero/sharded_model/sharded_grad.py
```diff
@@ -0,0 +1,122 @@
...
+        """This function will be called pre-backward
+
+        :raises AssertionError: Raise if grad shape is wrong
+        """
+        if self.param.grad is not None:
+            if self.param.grad.device != self.param.data.device:
+                # TODO: offload?
+                raise RuntimeError(
+                    'grad and param are on different device, grad {self.param.grad.device} vs. param {self.param.data.device}')
+                self.param.grad = None
```
feifeibear, line 56: Code is a bit redundant. move self.param.grad  = None out of if.
remove usless of branch.

colossalai/zero/sharded_model/sharded_grad.py
```diff
@@ -0,0 +1,122 @@
...
+        self._cpu_grad: Optional[torch.Tensor] = None
+
+        if self._cpu_offload:
+            # this buffer will be held and reused every iteration
+            self._cpu_grad = torch.zeros_like(param.data, device='cpu').pin_memory()
+
+    @torch.no_grad()
+    def setup(self) -> None:
+        """This function will be called pre-backward
+
```
feifeibear, line 48: Need more comments.
Is this function is used to clear the memory space of .grad before BWD?

colossalai/zero/sharded_model/sharded_grad.py
feifeibear: Also you shall not assume the parameter.data has been gathered! Therefore, use its size is not appropriate.

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -45,19 +60,106 @@ def __init__(self,
...
+        # However, if you set gradient_predivide_factor to None, we will set gradient_predivide_factor to a value >= 1.0 automatically
+        self.gradient_predivide_factor: float = gradient_predivide_factor if gradient_predivide_factor is not None else \
+            get_gradient_predivide_factor(self.world_size)
+        self.gradient_postdivide_factor: float = self.world_size / self.gradient_predivide_factor
+
+        self.comm_stream: torch.cuda.Stream = torch.cuda.Stream()
+        self.reducer = ReduceScatterBucketer(reduce_scatter_bucket_size_mb)
+        self._require_backward_grad_sync: bool = True
+
+        self._register_grad_post_backward_hooks()
```
feifeibear, line 83: Duplicated hooks? conflict with the ShardedGradient?

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -45,19 +60,106 @@ def __init__(self,
...
+            # Leave the gradient accumulation state as-is if not synchronizing this pass. This ensures p.grad
+            # remains the unsharded gradient accumulated from prior no-sync passes, and _saved_grad_shard
+            # remains the sharded gradient from the last synchronized pass. This also allows interleaved no-sync and
+            # sync passes, if desired.
+            if not self._require_backward_grad_sync:
+                continue
+            p._sharded_grad.write_back()
+
+    @torch.no_grad()
+    def _grad_post_backward_hook(self, param: Parameter, grad: torch.Tensor) -> Optional[torch.Tensor]:
```
feifeibear, line 116: do we still need this hook?

colossalai/zero/sharded_model/sharded_model_v2.py
ver217: This hook is registered on parameter and it's for gradient reduction. ShardedGradient just manages hooks about setting up gradients before backward.

colossalai/zero/sharded_model/sharded_model_v2.py
ver217: In my experiments, I find hooks registered on module will be fired more than once, if the module is used more than once. However, hooks registerd on parameter will be fired only once when gradient is computed, so I think this kind of hook is simpler.

feifeibear (commented):
Make sure the shared_model_v2 tests run in real distributed enviroment.

feifeibear (approved):
LGTM

PR #291, opened 2022-03-02
added buffer sync to naive amp model wrapper

### LINKED ISSUE #258: Enhance fp16 compatibility with SyncBN
## Feature Request

When using naive amp or zero, we need to convert the whole model to fp16 for better memory management. In data parallel training, we need to convert the vanilla batch normalization layer to sync batch normalization layer so that the running statistics are consistent across different ranks. However, the current fp16 implementation converts the `count` in batch norm to fp16 as well and some internal data type checks will fail. If no sync bn is used, models on different ranks will produce different results even given the same inputs. Thus, we should support sync bn with FP16.

There are several ways for implementation and open for discussion
1. write colossalai's own sync bn
2. adapt fp16 module to support a mixture of fp16 and fp32

## To Reproduce

```python
# create resnet
model = resnet18(num_classes=10).cuda()
model = torch.nn.SyncBatchNorm.convert_sync_batchnorm(model, process_group=gpc.get_group(ParallelMode.DATA))
model = model.half()

...

for img, label in dataloader:
      output = model(img)
      ...

```

This will raise exception `RuntimeError: Expected counts to have type Half but got Float`

## Screenshots

Screenshot 2022-02-24 at 5 58 16 PM

FrankLeeeee:
Fixed #258


FrankLeeeee:
I have tested this with Sharded Optimizer. For  Sharded Model, I think just need to wrap Sharded Model with NaiveAMPModel so that it can sync buffers. @ver217 do you have any comments?

ver217:
> I have tested this with Sharded Optimizer. For Sharded Model, I think just need to wrap Sharded Model with NaiveAMPModel so that it can sync buffers. @ver217 do you have any comments?

Agree.

tests/test_zero_data_parallel/test_sharded_optim_with_sync_bn.py
```diff
@@ -0,0 +1,87 @@
...
+    # need to make sure the batch norm stats are synchronized
+    # so that given the same input, the model will produce the same
+    # output on different ranks
+    engine.eval()
+    data = torch.rand(4, 3, 128, 128).cuda().half()
+    dist.broadcast(data, src=0, group=gpc.get_group(ParallelMode.DATA))
+
+    # predict
+    out = engine(data)
+    print(out)
```
feifeibear, line 58: remove this line?

tests/test_zero_data_parallel/test_sharded_optim_with_sync_bn.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # test if results are equal
+    tensor_list = [torch.empty_like(out) for _ in range(world_size - 1)]
```
feifeibear, line 61: why world_size - 1?

tests/test_zero_data_parallel/test_sharded_optim_with_sync_bn.py
FrankLeeeee: Sure.

FrankLeeeee:
Conflict resolved with develop.

PR #292, opened 2022-03-02
add a common util for hooks registered on parameter.

feifeibear:
Extract hook logic to common utils.
Provides a mechanism to register hooks to params so that users don't need to know how to use hooks (which requires reading a lot of documentation to gain knowledge).
double check with @ver217 , that the backward hook registered on a parameter can only be fired once during BWD.

### DIFF
--- a/colossalai/engine/paramhooks/_base_paramhook.py
+++ b/colossalai/engine/paramhooks/_base_paramhook.py
@@ -1,5 +1,6 @@
 from typing import Callable, List
 import torch
+import functools
 
 class BaseParamHook(object):
     def __init__(self, param_list: List[torch.nn.Parameter]) -> None:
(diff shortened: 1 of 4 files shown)


colossalai/engine/paramhooks/_base_paramhook.py
```diff
@@ -0,0 +1,37 @@
...
+        is computed. 
+        The hook should have the following signature:
+        ```
+        hook(grad) -> Tensor or None
+        ```
+        """
+        if not torch.is_grad_enabled():
+            return  # don't register grad hooks if grad isn't enabled
+        for p in self.param_list:
+            if p.requires_grad and not hasattr(p, 'zero_shard_bwd_hook'):
```
ver217, line 24: Hi, I think we should not use 'zero_shard_bwd_hook' as the attribute name. Maybe we should use a more general one.

colossalai/engine/paramhooks/_base_paramhook.py
```diff
@@ -0,0 +1,37 @@
...
+        self.param_list = param_list
+        self._register_backward_hooks(backward_hook_call)
+
+    def _register_backward_hooks(self, hook_call : Callable) -> None:
+        r"""
+        The hook_call will be called every time a gradient with respect to the a param in self.param_list 
+        is computed. 
+        The hook should have the following signature:
+        ```
+        hook(grad) -> Tensor or None
```
ver217, line 18: I think the hook may be parameter relevant. Can this hook receive a parameter as argument?

feifeibear: Sure. You define a hook function in this way and add parameters as many as you need in a wrapper class.
```
class HooKWrapper:
        def __init__(self) -> None:
            self.hook_triggered_times = 0

        def wrapper_func(self):
            def hook(grad) -> torch.Tensor or None:
                self.hook_triggered_times += 1
                return grad
            return hook
```

ver217 (approved):

PR #296, opened 2022-03-02
primary memory tracer

xcetos:
This primary memory tracer is used to collect memory usage during each iteration, and it works well on the standalone node. Adaptive sampling and timer synchronization have already been implemented but the unit test is not yet finished. There are some other design problems:

1. Shall we provide the `warmup` parameter? In most cases, only several iterations will be affected by non-training factors. 
2. I believe users are supposed to show the results based on iterations or time intervals, e.g. show the memory usage between the 10-th and 15-th iterations.

### DIFF
--- a/colossalai/engine/ophooks/_memtracer_ophook.py
+++ b/colossalai/engine/ophooks/_memtracer_ophook.py
@@ -5,18 +5,18 @@ from colossalai.registry import OPHOOKS
 from colossalai.logging import get_dist_logger
 from time import sleep, time
 import pickle
+from typing import Union, Optional
 
-
-def get_cuda_memory_used(device):
+def get_cuda_memory_used(device: Optional[torch.device]) -> int:
     """
     Get the free memory info of device.
     Notice that for CPU, this function will return 1/N of the total free memory,
     where N is the world size.
     """
-    ret = torch.cuda.memory_allocated()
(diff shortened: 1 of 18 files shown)


FrankLeeeee:
Hi Jie, the unit testing failed, can you check your test code again?

colossalai/engine/ophooks/_memtracer_ophook.py
```diff
@@ -74,22 +78,59 @@ def state_dict(self):
...
+        warmup (int): This parameter indicates how many iterations to truncate
+        before profiling, e.g. set to 5 and the data will start from 6-th iteration
+        refreshrate (int): This parameter decides the frequency of write file.
+        datafile(string): the name of the stats data file
+    Attributes:
+        _warmup (int): warmup iterations
+        _refreshrate(int): how many iterations we shall refresh the file
+        _logger (colossalai.logging.logger): output log file
+        _curiter (int): current iteration number
+        _count (int): the number of times the data file was written
```
feifeibear, line 102: rename it to writting_file_count_?

feifeibear (approved):
LGTM

colossalai/engine/ophooks/_memtracer_ophook.py
xcetos: Ok, I will fix it ASAP

PR #301, opened 2022-03-03
add sharded adam

ver217:


colossalai/zero/sharded_optim/sharded_adam.py
```diff
@@ -0,0 +1,163 @@
...
+        self.grad_scaler = DynamicGradScaler(initial_scale=initial_scale,
+                                             min_scale=min_scale,
+                                             growth_factor=growth_factor,
+                                             backoff_factor=backoff_factor,
+                                             growth_interval=growth_interval,
+                                             hysteresis=hysteresis,
+                                             max_scale=max_scale)
+        self._found_overflow: Tensor = torch.FloatTensor([0]).to(self.state_device)
+
+        # Early state initialization
```
ver217, line 56: @feifeibear I think the sharded Adam is easy to expand toa  general sharded optimizer by just removing early state initialization here. And in step(), we have to make sure param.data has correct shape, dtype and device.

colossalai/zero/sharded_optim/sharded_adam.py
```diff
@@ -0,0 +1,163 @@
...
+
+        found_inf = self._check_overflow()
+        self.grad_scaler.update(found_inf)
+
+        if found_inf:
+            self.zero_grad()
+            return
+
+        # Write payload back to p.data
+        for group in self.optim.param_groups:
```
ver217, line 97: @feifeibear I write payload back to param.data here. Also, this function can handle unsharded (normal) parameters as well.

colossalai/zero/sharded_optim/sharded_adam.py
```diff
@@ -0,0 +1,163 @@
...
+                if hasattr(p, 'ca_attr'):
+                    data = p.ca_attr.payload(self.state_device)
+                if torch.is_floating_point(data) and data.dtype != torch.float:
+                    data = data.to(torch.float)
+                p.data = data
+        ret = self.optim.step(*args, **kwargs)
+        # Set p.data to None
+        for group in self.optim.param_groups:
+            for p in group['params']:
+                p.data = None
```
ver217, line 109: @feifeibear I should write the updated weight to sharded param as well. Do we have a method to update it?

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -81,6 +83,10 @@ def backward(self, loss):
         loss.backward()
         self._final_backward_hook()
 
+    def backward_by_grad(self, tensor, grad):
```
feifeibear, line 86: Why we have is API? Did PyTorch nn.Module has one?

colossalai/zero/sharded_optim/sharded_adam.py
```diff
@@ -0,0 +1,163 @@
...
+class OptimState(Enum):
+    SCALED = 1
+    UNSCALED = 2
+
+
+class ShardedAdam(ColossalaiOptimizer):
+
+    def __init__(self,
+                 adam_optim: Optimizer,
+                 sharded_model: nn.Module,
```
feifeibear, line 28: nn.Module -> Union[nn.Module, ShardedModelV2]

colossalai/zero/sharded_optim/sharded_adam.py
feifeibear: I agree with you on moving OS memory allocation and init here.

FrankLeeeee:
Hi, sharded_model_V2 has conflict.

colossalai/zero/sharded_optim/sharded_adam.py
feifeibear: I have discussed with ver217. The following code snippet is wrong.

colossalai/zero/sharded_optim/sharded_adam.py
FrankLeeeee: Agree. This part is not necessary as many torch optimizers will do lazy init.

colossalai/zero/sharded_model/sharded_model_v2.py
FrankLeeeee: This API is in fact for pipeline parallel. The pipeline stage which is not the last stage has to backward with grad and tensor.

feifeibear (approved):
LGTM.
I will add grad fp16->fp32 and param fp32->fp16 copy in another PR.

ver217:
> 

In sharded_grad, I store a fp32 copy after reduction. So the grad is fp32 when calling `optim.step()`.

PR #312, opened 2022-03-04
[zero] add more shard strategy

FrankLeeeee:
- To support more strategies, shard and gather should return a list of tensors and they should reversed operation to each other
- Added two more strategies: TensorListShardStrategy shards the tensor list in a greedy fashion while keep each tensor in full and FlatShardStrategy flattens all tensors and shard.


colossalai/zero/shard_utils/flat_shard_strategy.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    FlatShardStategy flattens all tensors and shards the 1D tensor.
+    """
+
+    def __init__(self, process_group: Optional[dist.ProcessGroup] = None) -> None:
+        super().__init__(process_group)
+
+    def shard(self, tensor_list: List[Tensor]) -> List[Tensor]:
+        flat_tensor = _flatten_dense_tensors(tensor_list)
+        new_tensor_list = _unflatten_dense_tensors(flat_tensor, tensor_list)
```
feifeibear, line 24: why _unflatten_dense_tensors? will it turn flat_tensor back to tensor_list?

FrankLeeeee: Yes, flatten will create a new tensor. Need to unflatten and update the `.data` attribute of the original tensor.

colossalai/zero/shard_utils/flat_shard_strategy.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # need to update the parameter data
+        for old, new in zip(tensor_list, new_tensor_list):
+            old.data = new.data
+
+        shard, _ = get_shard(flat_tensor, rank=self.local_rank, world_size=self.world_size)
+        shard.orig_shape_list = tensor_shape_list
+        return [shard]
+
+    def gather(self, tensor_list: List[ShardedTensor]) -> List[ShardedTensor]:
```
feifeibear, line 35: The parameter type is diff from the abstract class.
def gather(self, tensor_list: List[Tensor]) -> List[Tensor]:

FrankLeeeee: Noted and will change soon.

colossalai/zero/shard_utils/tensor_list_shard_strategy.py
```diff
@@ -0,0 +1,46 @@
...
+
+        # partititon the parameters in a greedy fashion
+        sorted_params = sorted(tensor_list, key=lambda x: x.numel(), reverse=True)
+        for param in sorted_params:
+            # allocate this parameter to the rank with
+            # the smallest numel for load balancing purpose
+            rank_to_go = numel_per_rank.index(min(numel_per_rank))
+            params_per_rank[rank_to_go].append(param)
+            numel_per_rank[rank_to_go] += param.numel()
+        return params_per_rank[self.local_rank]
```
feifeibear, line 42: I think you may need to delete the memory of no-local tensors

colossalai/zero/shard_utils/base_shard_strategy.py
```diff
@@ -13,14 +27,14 @@ def __init__(self, process_group: Optional[dist.ProcessGroup] = None) -> None:
         super().__init__()
 
     @abstractmethod
-    def shard(self, tensor_list: List[ShardedTensor]):
+    def shard(self, tensor_list: List[Tensor]) -> List[Tensor]:
```
feifeibear, line 30: maybe we should change the payload of tensors in the tensor_list inline.
the function should not return a value?
The same as gather?

FrankLeeeee: I think this depends on the sharding granularity. If we shard a tensor, the sharded tensor will do. If we shard a tensor list, then we cannot store the sharded output into the payload.

feifeibear: I see. 
If so, the input tensor shall not be changed.
We can unify the design whether shard() can make impact on input tensors.

PR #315, opened 2022-03-04
add sharded optim v3

ver217:
Use `ShardedTensor` API. Support `ShardedModel` and normal model.

### DIFF
--- a/colossalai/zero/sharded_optim/sharded_optim_v3.py
+++ b/colossalai/zero/sharded_optim/sharded_optim_v3.py
@@ -1,0 +1,155 @@
+from enum import Enum
+from typing import Dict, Optional, Union
+
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+from colossalai.amp.naive_amp._fp16_optimizer import DynamicGradScaler
+from colossalai.context.parallel_mode import ParallelMode
+from colossalai.core import global_context as gpc
+from colossalai.nn.optimizer import ColossalaiOptimizer
+from colossalai.zero.sharded_model import ShardedModelV2
+from torch import Tensor
+from torch.distributed import ProcessGroup
+from torch.nn.parameter import Parameter
+from torch.optim import Optimizer
+
+from ..shard_utils import TensorShardStrategy
+from ..sharded_param import ShardedTensor
+from ._utils import has_inf_or_nan
+
+
+class OptimState(Enum):
+    SCALED = 1
+    UNSCALED = 2
+
+
+class ShardedOptimizerV2(ColossalaiOptimizer):
+
+    def __init__(self,
+                 optimizer: Optimizer,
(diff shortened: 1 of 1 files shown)


colossalai/zero/sharded_optim/sharded_optim_v3.py
```diff
@@ -0,0 +1,155 @@
...
+                 min_scale: float = 1,
+                 growth_factor: float = 2,
+                 backoff_factor: float = 0.5,
+                 growth_interval: float = 1000,
+                 hysteresis: float = 2,
+                 max_scale: int = 2**32,
+                 dp_process_group: Optional[ProcessGroup] = None,
+                 mp_process_group: Optional[ProcessGroup] = None) -> None:
+        super().__init__(optimizer)
+        self.model: Union[nn.Module, ShardedModelV2] = sharded_model
```
feifeibear, line 43: I just wonder if we can do it in this way.
If the class receive a Module type sharded_model. It just converts it into a ShardedModelV2

sharded_model = ShardedModelV2(sharded_model， ..)

colossalai/zero/sharded_optim/sharded_optim_v3.py
```diff
@@ -0,0 +1,155 @@
...
+                    p.data = self.master_params[p]
+                else:
+                    self.shard_strategy.shard([self.master_params[p]])
+                    p.data = self.master_params[p].payload
+        ret = self.optim.step(*args, **kwargs)
+        # Write master param to payload
+        for group in self.optim.param_groups:
+            for p in group['params']:
+                if hasattr(p, 'ca_attr'):
+                    p.ca_attr.data = p.data
```
feifeibear, line 99: Did you make sure the param copy is fp32 -> fp16?

feifeibear: If it is ZeRO3, it should copy cpu fp32 -> gpu fp16.

feifeibear: the setter of ShardedParam data is not good, you can add a new function as copy_data(self, t)?

ver217: OK

colossalai/zero/sharded_optim/sharded_optim_v3.py
```diff
@@ -0,0 +1,155 @@
...
+from ..sharded_param import ShardedTensor
+from ._utils import has_inf_or_nan
+
+
+class OptimState(Enum):
+    SCALED = 1
+    UNSCALED = 2
+
+
+class ShardedOptimizerV2(ColossalaiOptimizer):
```
FrankLeeeee, line 27: File name is V3 but the class name is V2.

PR #317, opened 2022-03-04
[zero] fix sharded optim with offload and add unit test

ver217:


colossalai/zero/sharded_model/sharded_grad.py
```diff
@@ -6,13 +6,11 @@
 
 
 class ShardedGradient:
-    def __init__(self,
-                 param: Parameter,
-                 sharded_module: nn.Module,
-                 offload_config: Optional[dict] = None
-                 ) -> None:
+
```
feifeibear, line 9: The ShardedGradient is going to be deprecated.
replace it with ShardedTensor.

PR #321, opened 2022-03-07
[zero] zero init context

feifeibear:
A context for model initialization.
1. shard the model during initialization on-the-fly. It can reduce peak GPU memory usage.
Now init a 8B model, requires 32 GB GPU memory (data fp16 + grad fp32). Use this context. It requires 32 GB/#GPU memory. Also we can offload the param to CPU on-the-fly.
2. convert param data/grad to fp16
3. adapt the param to ShardedParamV2.

### DIFF
--- a/colossalai/zero/init_ctx/init_context.py
+++ b/colossalai/zero/init_ctx/init_context.py
@@ -1,0 +1,115 @@
+import functools
+import torch
+from colossalai.zero.shard_utils import BaseShardStrategy
+from colossalai.zero.sharded_param import ShardedParamV2
+
+
+# Inserts _post_init_method at the end of init method
+# for all sub classes of torch.nn.Module
+class InsertPostInitMethodToModuleSubClasses(object):
+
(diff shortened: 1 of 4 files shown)


colossalai/zero/init_ctx/init_context.py
```diff
@@ -0,0 +1,128 @@
...
+    3. Shard the param and grad according to flags.
+    """
+
+    def __init__(
+        self,
+        is_convert_fp16: bool,
+        is_convert_cuda: bool,
+        shard_strategy: BaseShardStrategy,
+        is_shard_param: bool = False,
+        is_shard_grad: bool = False,
```
FrankLeeeee, line 91: In general , parameters in function signature do not start with `is`, e.g. `is_shard_param` should be just `shard_param`.

feifeibear: Got it! DONE

FrankLeeeee (approved):

PR #323, opened 2022-03-07
[zero] Update sharded model v2 using sharded param v2

ver217:
1. Refactor `ShardedParamV2`, as `grad` is never gathered. I think we can just use a `tensor` to store the `grad` , which meets our requirements well.
2. Refactor `ShardedModelV2` using `ShardedParamV2`.
3. Fix bugs， like circular import in `zero_hook.py`.
4. Rename typos.

### DIFF
--- a/colossalai/engine/ophooks/zero_hook.py
+++ b/colossalai/engine/ophooks/zero_hook.py
@@ -1,0 +1,45 @@
+import torch
+from . import BaseOpHook
+from colossalai.registry import OPHOOKS
+from colossalai.zero.shard_utils import TensorShardStrategy
+
+
+@OPHOOKS.register_module
+class ZeroHook(BaseOpHook):
+    """
+    A hook to process sharded param for ZeRO method.
+    """
+
+    def __init__(self, process_group):
+        super().__init__()
+        self.shard_strategy = TensorShardStrategy(process_group=process_group)
+
+    def pre_fwd_exec(self, module: torch.nn.Module, *args):
+        for param in module.parameters():
+            assert hasattr(param, 'ca_attr')
+            self.shard_strategy.gather([param.ca_attr.data])
+            param.data = param.ca_attr.data
+
+    def post_fwd_exec(self, module: torch.nn.Module, *args):
+        for param in module.parameters():
+            assert hasattr(param, 'ca_attr')
+            self.shard_strategy.shard([param.ca_attr.data])
+            param.data = None
+
+    def pre_bwd_exec(self, module: torch.nn.Module, input, output):
+        for param in module.parameters():
+            assert hasattr(param, 'ca_attr')
+            self.shard_strategy.gather([param.ca_attr.data])
+            param.data = param.ca_attr.data
+
+    def post_bwd_exec(self, module: torch.nn.Module, input):
+        for param in module.parameters():
+            assert hasattr(param, 'ca_attr')
+            self.shard_strategy.shard([param.ca_attr.data])
+            param.data = None
+
+    def pre_iter(self):
+        pass
+
+    def post_iter(self):
+        pass
--- a/colossalai/engine/ophooks/zero_hook.py
+++ b/colossalai/engine/ophooks/zero_hook.py
@@ -34,14 +34,16 @@ class ZeroHook(BaseOpHook):
             param.data = param.col_attr.data.payload
             # Store local accumulated grad shard
             if param.grad is not None:
-                if param.col_attr.grad is None:
+                if param.col_attr.bwd_count == 0:
                     # We haven't stored local accumulated grad yet
+                    assert param.col_attr.grad is None
                     param.col_attr.grad = param.grad.data
                     param.grad = None
                 else:
                     # We have stored local accumulated grad
                     # The grad here must be locally computed full grad in this backward pass
                     assert param.grad.shape == param.col_attr.data.origin_shape
+            param.col_attr.bwd_count += 1
 
     def post_bwd_exec(self, module: torch.nn.Module, input):
         for param in module.parameters():
--- a/colossalai/engine/ophooks/zero_hook.py
+++ b/colossalai/engine/ophooks/zero_hook.py
@@ -38,6 +38,12 @@ class ZeroHook(BaseOpHook):
             self.shard_strategy.shard([param.ca_attr.data])
             param.data = None
 
+            # save param.grad to some place  in case gradient accumulation.
+            # at the moment, the ca_attr.grad payload is useless.
+            # gather the memory first
+            self.shard_strategy.gather([param.ca_attr.grad])
+            param.grad = None
+
     def pre_iter(self):
         pass
 
--- a/colossalai/engine/ophooks/zero_hook.py
+++ b/colossalai/engine/ophooks/zero_hook.py
@@ -1,7 +1,8 @@
 import torch
-from . import BaseOpHook
 from colossalai.registry import OPHOOKS
 from colossalai.zero.shard_utils import TensorShardStrategy
+
+from ._base_ophook import BaseOpHook
 
 
 @OPHOOKS.register_module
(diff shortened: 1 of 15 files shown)


colossalai/zero/sharded_param/sharded_param.py
```diff
@@ -1,44 +1,38 @@
...
         param.data = torch.empty([], dtype=param.dtype, device=param.device)
 
     @property
     def data(self):
-        return self._data_sharded_tensor.payload
-
-    @data.setter
-    def data(self, t: torch.Tensor):
-        self._data_sharded_tensor.payload = t
+        return self._data_sharded_tensor
```
feifeibear, line 23: zero context init should be updated accordingly.

colossalai/engine/ophooks/zero_hook.py
```diff
@@ -0,0 +1,51 @@
...
+
+@OPHOOKS.register_module
+class ZeroHook(BaseOpHook):
+    """
+    A hook to process sharded param for ZeRO method.
+    """
+
+    def __init__(self, process_group):
+        super().__init__()
+        self.shard_strategy = TensorShardStrategy(process_group=process_group)
```
feifeibear, line 16: pass the strategy via parameters.

colossalai/zero/sharded_param/sharded_param.py
ver217: Ok, I will fix this.

colossalai/engine/ophooks/zero_hook.py
ver217: OK

colossalai/engine/ophooks/zero_hook.py
```diff
@@ -0,0 +1,51 @@
...
+    def pre_bwd_exec(self, module: torch.nn.Module, input, output):
+        for param in module.parameters():
+            assert hasattr(param, 'col_attr')
+            self.shard_strategy.gather([param.col_attr.data])
+            param.data = param.col_attr.data.payload
+
+            # Store local accumulated grad shard
+            if param.grad is not None:
+                param.col_attr.grad = param.grad.data
+                param.grad = None
```
feifeibear, line 39: I have a bad feeling for these code. We will catch you offline.

tests/test_zero_data_parallel/test_shard_model_v2.py
```diff
@@ -30,12 +30,7 @@ def run_fwd_bwd(model, x, enable_autocast=False):
...
 def run_dist(rank, world_size, port):
-    colossalai.launch(config=CONFIG,
-                      rank=rank,
-                      world_size=world_size,
-                      host='localhost',
-                      port=port,
-                      backend='nccl')
+    colossalai.launch(config=CONFIG, rank=rank, world_size=world_size, host='localhost', port=port, backend='nccl')
 
     model = Net(checkpoint=True).cuda()
```
feifeibear, line 35: test model with sharded parameters for different layers.

ver217: Can you give me an example? I don't exactly know what you mean.

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -45,21 +44,21 @@ def __init__(
         self.rank = dist.get_rank(self.process_group)
 
         # The module has to be placed on GPU
-        self.module = module.cuda()
+        self.module = module.half().cuda()
 
+        self.shard_strategy = TensorShardStrategy(process_group)
```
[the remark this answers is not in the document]
ver217: Done

colossalai/engine/ophooks/zero_hook.py
ver217: I fix this issue without gathering. Actually, if we make sure this code snippet is executed only when the `pre_fwd_hook` is first fired, I think there is no problem.

ver217:
I skip `test_sharded_optim_v2` now, as the `ShardedOptimV2` is incomplete. I will enable this unit test in next MR.

tests/test_zero_data_parallel/test_shard_model_v2.py
feifeibear: OK, it's fine. Can you add another unitest which init model with ZeroInitContext

feifeibear: Also, I just merged an MR polishing unittests. Do test with the following code.
```
from tests.components_to_test.registry import non_distributed_component_funcs

for get_components_func in non_distributed_component_funcs:
    model, train_dataloader, test_dataloader, optimizer, criterion = get_components_func()
    do_your_test()
```

ver217: OK,  I will refactor the unit test with this code snippet.

colossalai/engine/ophooks/zero_hook.py
```diff
@@ -0,0 +1,56 @@
...
+class ZeroHook(BaseOpHook):
+    """
+    A hook to process sharded param for ZeRO method.
+    """
+
+    def __init__(self, shard_strategy: BaseShardStrategy):
+        super().__init__()
+        self.shard_strategy = shard_strategy
+
+    def pre_fwd_exec(self, module: torch.nn.Module, *args):
```
feifeibear, line 18: add logic to deal with NestedNet.
see if elem in args is nn.Paramater.
gather the shared param?
and shard it afterwards?

feifeibear: You just skip the nested model.

ver217: OK

tests/test_zero_data_parallel/common.py
```diff
@@ -71,11 +71,11 @@ def check_grads_padding(model, zero_model, loose=False):
         chunks = torch.flatten(p.grad).chunk(dist.get_world_size())
         if rank >= len(chunks):
             continue
-        grad = chunks[rank]
+        grad = chunks[rank].float()
```
feifeibear, line 74: why convert to float?
When I see the name check_grads_padding, I won't expect you cast my model to fp16 during the check.

ver217: The original model is fp16, so the grad is also fp16.

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -1,37 +1,38 @@
...
 from colossalai.logging import get_dist_logger
+from colossalai.zero.shard_utils import BaseShardStrategy
 from colossalai.zero.sharded_model.reduce_scatter import ReduceScatterBucketer
-from colossalai.zero.sharded_model.sharded_grad import ShardedGradient
-from colossalai.zero.sharded_param import ShardedParam
 from torch.distributed import ProcessGroup
 from torch.nn.parameter import Parameter
 
-from ._zero3_utils import chunk_and_pad, get_gradient_predivide_factor
+from ..sharded_param import ShardedParamV2
```
feifeibear, line 19: use absolute path.

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -161,13 +169,30 @@ def _reduce_scatter_callback(self, param: Parameter, reduced_grad: torch.Tensor)
...
-        if self.mixed_precision:
-            orig_param_grad_data = reduced_grad.data
-            reduced_grad.data = reduced_grad.data.to(dtype=param.ca_attr.origin_dtype)
-            # Don't let this memory get reused until after the transfer.
-            orig_param_grad_data.record_stream(torch.cuda.current_stream())
-
-        param._sharded_grad.reduce_scatter_callback(reduced_grad)
+
+        # Make sure we store fp32 grad
+        reduced_grad.data = cast_tensor_to_fp32(reduced_grad.data)
```
feifeibear, line 174: Where does tensor reduced_grad come from?

ver217: When dp size = 1, `reduced_grad` is locally computed final gradient. When dp size > 1, `reduced_grad` is the gradient shard after `reduce-scatter`.

tests/test_zero_data_parallel/common.py
feifeibear: what is original model?
grad = chunks[rank].float() if zero_grad.dtype == torch.float else chunks[rank]

colossalai/zero/sharded_model/sharded_model_v2.py
feifeibear: I see

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -161,13 +169,30 @@ def _reduce_scatter_callback(self, param: Parameter, reduced_grad: torch.Tensor)
...
+
+        # Maybe offload
+        if self._cpu_offload:
+            reduced_grad.data = reduced_grad.data.cpu()
+
+        if param.col_attr.grad is None:
+            param.col_attr.grad = reduced_grad.data
+        else:
+            param.col_attr.grad.add_(reduced_grad.data)
```
feifeibear, line 183: if cpu_offload, you may add a cuda tensor with an CPU tensor?

tests/test_zero_data_parallel/common.py
ver217: In `test_shard_model_v2.py`, we call `model.half()`. So the parameter and gradient of the model which our `ShardedModel` compares are always fp16. And as we store fp32 grad shard for `ShardedModel`, `zero_grad` is always fp32 here.

colossalai/zero/sharded_model/sharded_model_v2.py
ver217: If  `param.col_attr.grad` is not `None` here, it must come from previous local accumulated gradient. And if we enable `CPU offload`, the previous local accumulated gradient must be on CPU. So I think there is no problem.

feifeibear (approved):

PR #324, opened 2022-03-07
refactored testing components

FrankLeeeee:
- collates the model, dataloader, optimizer and criterion into `tests/components_to_test`. In unit testing, if you want to iterate over the different sets of training components for higher test case coverage, use the following sample code.

```python
from tests.components_to_test.registry import non_distributed_component_funcs

for get_components_func in non_distributed_component_funcs:
    model, train_dataloader, test_dataloader, optimizer, criterion = get_components_func()
    do_your_test()
```

- refactored test engine to save `colossalai.launch` time and test over different non-distributed models

- A separate PR will be used to refactor other test cases.


feifeibear (approved):
LGTM

PR #329, opened 2022-03-08
added base profiler

kurisusnowdeng:


colossalai/utils/profiler/common.py
```diff
@@ -0,0 +1,75 @@
...
+from .comm_profiler import CommProfiler
+from typing import List
+from torch.utils.tensorboard import SummaryWriter
+
+
+class ProfilerEvent(Enum):
+    '''
+    Event types to profile
+    '''
+    COMM = CommProfiler
```
FrankLeeeee, line 11: why is profiler an event?

colossalai/utils/profiler/common.py
```diff
@@ -0,0 +1,75 @@
+from enum import Enum
+from .comm_profiler import CommProfiler
+from typing import List
+from torch.utils.tensorboard import SummaryWriter
```
FrankLeeeee, line 4: should catch import error here as this requires tensorboard installation

colossalai/utils/profiler/_base_profiler.py
```diff
@@ -0,0 +1,28 @@
+from torch.utils.tensorboard import SummaryWriter
+
+
+class BaseProfiler(object):
+
+    def __init__(self):
+        ...
```
FrankLeeeee, line 7: this `...` will raise error

colossalai/utils/profiler/_base_profiler.py
```diff
@@ -0,0 +1,28 @@
+from torch.utils.tensorboard import SummaryWriter
+
+
+class BaseProfiler(object):
```
FrankLeeeee, line 4: make it an abstract class with abc

colossalai/utils/profiler/common.py
```diff
@@ -0,0 +1,75 @@
...
+        for e in self.events:
+            e.disable()
+
+    def to_tensorboard(self, writer: SummaryWriter):
+        for e in self.events:
+            e.to_tensorboard(writer)
+
+    def to_file(self, filename: str):
+        for e in self.events:
+            e.to_file(filename)
```
FrankLeeeee, line 35: multiple profilers write into the same file?

colossalai/utils/profiler/common.py
```diff
@@ -0,0 +1,75 @@
...
+    from torch.utils.tensorboard import SummaryWriter
+
+    with colossalai_profiler(events=[ProfilerEvent.COMM]) as prof:
+        train()
+    writer = SummaryWriter('tb/path')
+    prof.to_tensorboard(writer)
+    prof.to_file('./prof.log')
+    prof.show()
+    '''
+    return ProfilerContext(*args, **kwargs)
```
FrankLeeeee, line 75: This wrapper does not do anything, can just let user import ProfilerContext

kurisusnowdeng: Just mean to use lowercase api for context

colossalai/utils/profiler/common.py
kurisusnowdeng: How about provide a path and add 'profilerevent_name.log' as a suffix

colossalai/utils/profiler/common.py
FrankLeeeee: It is alright to use `ProfilerContext`. In fact, the `*args` and `**kwargs` can hurt usability.

colossalai/utils/profiler/_base_profiler.py
kurisusnowdeng: Which error? I tried a simple test, and it was fine.

colossalai/utils/profiler/common.py
kurisusnowdeng: Pytorch already helps raise the import error

colossalai/utils/profiler/_base_profiler.py
FrankLeeeee: Not an error, but this function should be removed if no initialization is specified.

colossalai/utils/profiler/_base_profiler.py
feifeibear: derive from ABC if it's an abstract class

colossalai/utils/profiler/common.py
kurisusnowdeng: It is to distinguish events (or activities? as named by torch) to profile (e.g. communication, memory), and each event relates to a profiler

colossalai/utils/profiler/common.py
FrankLeeeee: This line makes TensorBoard a must.

kurisusnowdeng: Do you mean even there is no tensorboard installed we still make profiler work?

FrankLeeeee: Yes

FrankLeeeee: In fact, a better workaround of this is to import tensorboard only in `to_tensorboard`. Type hinting can be given as a string e.g. 

```python
writer: "torch.utils.tensorboard.SummaryWriter"
```

colossalai/utils/profiler/common.py
FrankLeeeee: It is kind of confusing as in communication profiler, it has defined its own event.

kurisusnowdeng: So change it to 'ProfilerActivity'?

colossalai/utils/profiler/common.py
FrankLeeeee: One way to handle this is to set the format as {output_path}/{prefix}_{profiler.__class__.__name__}_{timestamp}.log. This method requires changing the parameter filename to otuput_path

colossalai/utils/profiler/common.py
FrankLeeeee, line 11: Do not use abbreviation for API exposed to the user.

FrankLeeeee: ok

FrankLeeeee: It is not a common practice to put a class as the enum value.  It seems not standard to me even though it works. What do you think @feifeibear .

colossalai/utils/profiler/common.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class ProfilerActivity(Enum):
+    '''
+    Event types to profile
+    '''
+    COMM = CommProfiler
+
+
+class colossalai_profiler(object):
```
FrankLeeeee, line 17: Follow python class naming convention

FrankLeeeee, line 17: Follow python class naming convention

kurisusnowdeng:
@FrankLeeeee All fixed.

colossalai/utils/profiler/__init__.py
```diff
@@ -1 +1,2 @@
-from .comm_profiler import enable_communication_prof, communication_prof_show
+# from .comm_profiler import enable_communication_prof, communication_prof_show
+from .common import ProfilerActivity, colossalai_profiler
```
FrankLeeeee, line 2: Will this change cause the test_profiler unit testing to fail?

kurisusnowdeng: Definitely. Need to update unitests after the update of specific profilers.

colossalai/utils/profiler/common.py
feifeibear: Also weird to me

colossalai/utils/profiler/__init__.py
kurisusnowdeng: Sure. Skip it.

colossalai/utils/profiler/common.py
FrankLeeeee: we can directly do this

```python
with Profiler(...) as prof:
    do_something()
```

I don't think we need to have the repository name in the class name.

kurisusnowdeng: Sure, but conventionally context classes are lowercase

FrankLeeeee: I see, I can accept the current code then.

kurisusnowdeng:
@FrankLeeeee Fixed.

colossalai/utils/profiler/common.py
FrankLeeeee: @kurisusnowdeng Any idea about this part?

kurisusnowdeng: We can also use some keys to redirect to the profiler classes in a dict. But, I think this is a redundant step.

xcetos: Actually, using Enum as an option is quite common in C++ and Rust, but I have no idea whether this is appropriate for Python.

FrankLeeeee:
Unit testing failed.

kurisusnowdeng:
Updated. Should be fine now. Also changed the way to initialize the profiler context into passing profiler instances in, according to the suggestion from @1SAA and @xcetos 。

colossalai/utils/profiler/common.py
```diff
@@ -0,0 +1,83 @@
...
+class ProfilerContext(object):
+    '''
+    Profiler context manager
+
+    Usage:
+    from colossalai.utils.profiler import ProfilerActivity, profiler
+    from torch.utils.tensorboard import SummaryWriter
+
+    with profiler(activities=[ProfilerActivity.COMMUNICATION]) as prof:
+        train()
```
FrankLeeeee, line 18: This should be updated.

colossalai/utils/profiler/common.py
```diff
@@ -0,0 +1,83 @@
...
+    writer = SummaryWriter('tb/path')
+    prof.to_tensorboard(writer)
+    prof.to_file('./prof_logs/')
+    prof.show()
+    '''
+
+    def __init__(self, activities: List[BaseProfiler] = None):
```
FrankLeeeee, line 25: shall we change the parameter name as the type is a profiler

kurisusnowdeng: I think 'activities' also makes sense, and maybe easier for pytorch users to distinguish the specific profilers from the profiler context (just too many 'profiler's).

kurisusnowdeng: @FrankLeeeee BTW, ready to merge if there is no other problems.

FrankLeeeee:
According to @1SAA , this PR is replaced by #340 so I will close this PR now.

PR #331, opened 2022-03-08
Added activation offload

Gy-Lu:
Have passed the unit test.

### DIFF
--- a/colossalai/utils/activation_checkpoint.py
+++ b/colossalai/utils/activation_checkpoint.py
@@ -5,14 +5,16 @@ import torch
 from torch.utils.checkpoint import check_backward_validity, detach_variable
 
 from colossalai.context.random import get_states, get_current_mode, set_seed_states, set_mode, sync_states
-
+from .cuda import get_current_device
 
 class CheckpointFunction(torch.autograd.Function):
 
     @staticmethod
-    def forward(ctx, run_function, *args):
+    def forward(ctx, run_function, activation_offload=False, *args):
         check_backward_validity(args)
         ctx.run_function = run_function
+        ctx.activation_offload = activation_offload
+        ctx.device = get_current_device()
 
         # preserve rng states
         ctx.fwd_cpu_rng_state = torch.get_rng_state()
@@ -32,7 +34,12 @@ class CheckpointFunction(torch.autograd.Function):
         tensor_inputs = []
         for i, arg in enumerate(args):
             if torch.is_tensor(arg):
-                tensor_inputs.append(arg)
+                if ctx.activation_offload:
+                    tmp = arg.detach().cpu()
+                    tmp.requires_grad = arg.requires_grad
+                    tensor_inputs.append(tmp)
+                else:
+                    tensor_inputs.append(arg)
                 ctx.tensor_indices.append(i)
                 ctx.inputs.append(None)
             else:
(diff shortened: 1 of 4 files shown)


tests/test_utils/test_activation_checkpointing.py
```diff
@@ -19,7 +19,7 @@ def forward(x, weight):
 
 
 @pytest.mark.gpu
```
feifeibear, line 21: @pytest.mark.parametrize(cpu_offload [True, False])
def test_activation_checkpointing(cpu_offload):

colossalai/utils/activation_checkpoint.py
```diff
@@ -70,8 +77,9 @@ def backward(ctx, *args):
 
         # Fill in inputs with appropriate saved tensors.
         for i, idx in enumerate(tensor_indices):
-            inputs[idx] = tensors[i]
-
+            tmp = tensors[i].detach().to(ctx.device)
+            tmp.requires_grad = tensors[i].requires_grad
```
feifeibear, line 81: Both detach() and requires_grad setting are unnecessary.
Because inputs will be detached later in detach_variable.

colossalai/utils/activation_checkpoint.py
```diff
@@ -32,7 +34,12 @@ def forward(ctx, run_function, *args):
         tensor_inputs = []
```
feifeibear, line 34: You missed an import procedure.
The forward can be executed during BWD of the model (grad checkpointing).
args should be guaranteed on cuda device before passing to run_function(*args).
I believe in some cases the args may resident on CPU.
https://github.com/microsoft/DeepSpeed/blob/013df6a9e8aef5d20ae0cf059c5fbd81cf90b694/deepspeed/runtime/activation_checkpointing/checkpointing.py#L569

tests/test_utils/test_activation_checkpointing.py
Gy-Lu: For we use  `add_seed` in the function. It can not be re-run to test.
Have fixed it and added the pytest.

colossalai/utils/activation_checkpoint.py
```diff
@@ -5,14 +5,16 @@
 from torch.utils.checkpoint import check_backward_validity, detach_variable
 
 from colossalai.context.random import get_states, get_current_mode, set_seed_states, set_mode, sync_states
-
+from .cuda import get_current_device
```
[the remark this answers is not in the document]
Gy-Lu: It will cause an ImportError.
For the `activation_checkpoint.py` and `cuda.py` are both under `utils`, and used by the same `__init__`.

colossalai/utils/activation_checkpoint.py
Gy-Lu: The `detach()` seems to be unnecessary, for `to()` would create a new tensor.
But the `requires_grad` is necessary, I have tested that the to() method would set `requires_grad=False` as well.

Gy-Lu: The `detach()` seems to be unnecessary, for `to()` would create a new tensor.
But the `requires_grad` is necessary, I have tested that the to() method would set `requires_grad=False` as well.

colossalai/utils/activation_checkpoint.py
Gy-Lu: ok, I will try to add the strategy for this case

feifeibear:
https://github.com/Oldpan/Pytorch-Memory-Utils
Can you use this tool to check current tensors in CPU and GPU.

Gy-Lu:
> https://github.com/Oldpan/Pytorch-Memory-Utils Can you use this tool to check current tensors in CPU and GPU.

Hi, sorry for the late reply. I have used this tool and found that all tensors were in their positions.

feifeibear (approved):

PR #336, opened 2022-03-09
[fix] set criterion as optional in colossalai initialize

FrankLeeeee:
- criterion is no longer required in `colossalai.initialize`
- updated the type hinting


feifeibear (approved):
LGTM

PR #337, opened 2022-03-09
[test] add bert unittest

feifeibear:
After adding a huggingface bert to unitest, I found some issues.
1. the inputs to sharded model can be integer tensors. We shall not cast it fp16.
2. the FWD+BWD can not obtain correct grads in the bert cast.

```
python test_shard_model_v2.py
```
I have skipped the underdevelopment test cases of sharded model.

### DIFF
--- a/colossalai/zero/sharded_model/sharded_model_v2.py
+++ b/colossalai/zero/sharded_model/sharded_model_v2.py
@@ -17,8 +17,9 @@ from colossalai.zero.sharded_param import ShardedParamV2
 from torch.distributed import ProcessGroup
 from torch.nn.parameter import Parameter
 
-from ._zero3_utils import (cast_float_arguments, cast_tensor_to_fp16, cast_tensor_to_fp32, chunk_and_pad,
-                           get_gradient_predivide_factor)
+from ._zero3_utils import (cast_tensor_to_fp32, chunk_and_pad, get_gradient_predivide_factor)
+
+# from ._zero3_utils import cast_float_arguments, cast_tensor_to_fp16
 
 
 class ShardedModelV2(nn.Module):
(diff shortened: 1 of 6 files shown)


colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -79,7 +80,8 @@ def __init__(self,
         self._require_backward_grad_sync: bool = True
 
     def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
-        args, kwargs = cast_float_arguments(cast_tensor_to_fp16, *args, **kwargs)
+        # TODO args can be Long!
```
ver217, line 83: `cast_tensor_to_fp16` will check whether the tensor if float, if not it won't be cast to fp16.

feifeibear: DONE, revert the origin code.

tests/test_zero_data_parallel/test_shard_model_v2.py
```diff
@@ -31,30 +31,52 @@ def run_fwd_bwd(model, data, label, criterion, enable_autocast=False):
...
-            run_fwd_bwd(model, data, label, criterion, False)
-            run_fwd_bwd(zero_model, data, label, criterion, False)
+
+            if criterion is None:
+                data, label = data.cuda(), label.cuda()
+                run_fwd_bwd_no_criterion(model, data, label, False)
+                run_fwd_bwd_no_criterion(zero_model, data, label, False)
+            else:
+                # FIXME() data can be interger!
+                data, label = data.half().cuda(), label.cuda()
```
feifeibear, line 69: @ver217  you should check the type of data before convert it to half()...

ver217: I will do this in next MR.

tests/components_to_test/bert.py
```diff
@@ -0,0 +1,81 @@
...
+    num_layer = 2
+
+    def bert_model_builder(checkpoint):
+        config = BertConfig(
+            gradient_checkpointing=checkpoint,
+            hidden_size=hidden_dim,
+            intermediate_size=hidden_dim * 4,
+            num_attention_heads=num_head,
+            max_position_embeddings=sequence_length,
+            num_hidden_layers=num_layer,
```
ver217, line 48: I will set dropout to 0 in next MR.

ver217 (approved):

tests/components_to_test/bert.py
feifeibear: OK...

PR #338, opened 2022-03-09
[fp16] refactored grad scaler

FrankLeeeee:
- re-organized the grad scaler used in naive amp
- will refactor the naive fp16 optimizer in a separate PR


feifeibear (approved):
LGTM

PR #347, opened 2022-03-09
[zero] add test sharded optim with cpu adam

ver217:
1. Test `ShardedOptimV2` with `CPUAdam`. Grad and fp32 master param are offloaded to CPU.

### DIFF
--- a/tests/test_zero_data_parallel/test_sharded_optim_v2_with_cpu_adam.py
+++ b/tests/test_zero_data_parallel/test_sharded_optim_v2_with_cpu_adam.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+
+import copy
+from functools import partial
+
+import colossalai
+import pytest
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+from colossalai.nn.optimizer import CPUAdam
+from colossalai.utils import free_port
+from colossalai.zero.shard_utils import TensorShardStrategy
+from colossalai.zero.sharded_model import ShardedModelV2
+from colossalai.zero.sharded_optim import ShardedOptimizerV2
+from tests.components_to_test.registry import non_distributed_component_funcs
+from torch.nn.parallel import DistributedDataParallel as DDP
+from torch.optim import Adam
+
+from common import CONFIG, check_sharded_params_padding
+
+
+def run_step(model, optimizer, data, label, criterion, enable_autocast=False):
+    model.train()
+    optimizer.zero_grad()
+    with torch.cuda.amp.autocast(enabled=enable_autocast):
+        y = model(data)
+        loss = criterion(y, label)
+    loss = loss.float()
+    if isinstance(model, ShardedModelV2):
+        optimizer.backward(loss)
+    else:
+        loss.backward()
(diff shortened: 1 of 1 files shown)


tests/test_zero_data_parallel/test_sharded_optim_v2_with_cpu_adam.py
```diff
@@ -0,0 +1,88 @@
...
+        model, train_dataloader, test_dataloader, optimizer, criterion = get_components_func()
+        model = model(checkpoint=True).cuda()
+        zero_model = ShardedModelV2(copy.deepcopy(model), shard_strategy, offload_config={'device': 'cpu'})
+        if dist.get_world_size() > 1:
+            model = DDP(model)
+        optim = Adam(model.parameters(), lr=1e-3)
+        sharded_optim = ShardedOptimizerV2(CPUAdam(zero_model.parameters(), lr=1e-3),
+                                           zero_model,
+                                           shard_strategy,
+                                           initial_scale=2**5,
```
feifeibear, line 65: initial_scale setting is consist with the `Adam(model.parameters(), lr=1e-3)`?

FrankLeeeee (approved):

tests/test_zero_data_parallel/test_sharded_optim_v2_with_cpu_adam.py
```diff
@@ -0,0 +1,88 @@
...
+def run_dist(rank, world_size, port):
+    colossalai.launch(config=CONFIG, rank=rank, world_size=world_size, host='localhost', port=port, backend='nccl')
+    test_models = ['repeated_computed_layers', 'resnet18', 'bert']
+    for model_name in test_models:
+        get_components_func = non_distributed_component_funcs.get_callable(model_name)
+        shard_strategy = TensorShardStrategy()
+        model, train_dataloader, test_dataloader, optimizer, criterion = get_components_func()
+        model = model(checkpoint=True).cuda()
+        zero_model = ShardedModelV2(copy.deepcopy(model), shard_strategy, offload_config={'device': 'cpu'})
+        if dist.get_world_size() > 1:
```
feifeibear, line 59: can we still use DDP when world size = 1

tests/test_zero_data_parallel/test_sharded_optim_v2_with_cpu_adam.py
ver217: Adam doesn't have grad scaler.

FrankLeeeee: it should be set to initial_scale = 1

FrankLeeeee: but since grad is not checked, I think this is fine as well?

feifeibear: I see

tests/test_zero_data_parallel/test_sharded_optim_v2_with_cpu_adam.py
ver217: Yes, we can. But I think it's meaningless.

PR #351, opened 2022-03-09
[test] polish zero related unitest

feifeibear:


ver217 (approved):

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -91,6 +93,20 @@ def backward_by_grad(self, tensor, grad):
         torch.autograd.backward(tensors=tensor, grad_tensors=grad)
         self._final_backward_hook()
 
+    def deepcopy_to(self, other_model: nn.Module):
```
FrankLeeeee, line 96: I feel this method is more like a util method in the unit test module instead of an api of sharded model which can be called by the user.

tests/test_zero_data_parallel/test_shard_model_v2.py
```diff
@@ -3,81 +3,69 @@
...
 
             check_grads_padding(model, zero_model, loose=True)
 
 
 @pytest.mark.dist
 @pytest.mark.parametrize("world_size", [1, 2, 4])
-def test_shard_model_v2(world_size):
-    run_func = partial(run_dist, world_size=world_size, port=free_port())
+@pytest.mark.parametrize("enable_autocast", [True])
+@pytest.mark.parametrize("use_zero_init_ctx", [True, False])
```
FrankLeeeee, line 60: This parameter combination can increase the CI time. I feel that the bottleneck is the setup of process groups in most test cases. I shall create our own `parameterize` decorator so that we don't have to spawn multiple processes for each test case. I will accept the current implementation for now.

colossalai/zero/sharded_model/sharded_model_v2.py
feifeibear: good idea. DONE

tests/test_zero_data_parallel/test_shard_model_v2.py
feifeibear: Thanks. I also removed all the existing with_ctx tests. So the time will not change compared to the current develop branch.

PR #353, opened 2022-03-09
[zero] fix grad shape error for ShardededModelv2

ver217:
Current code can handle ZeRO-3. However, if we don't shard param (ZeRO-2), current code will throw errors, because of wrong grad shape. I fix grad shape, so that we can handle both ZeRO-3 and ZeRO-2.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     container:
       image: nvcr.io/nvidia/pytorch:21.07-py3
       options: --gpus all --rm --ipc=host -v /data/scratch/cifar-10:/data/scratch/cifar-10
-    timeout-minutes: 20
+    timeout-minutes: 25
     steps:
       - name: Setup Environment
         run: |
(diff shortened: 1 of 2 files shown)


ver217:
I will update unit test when #351 is merged.

ver217:
We have to write new hooks for ZeRO-2. This MR is compatitable with ZeRO-3. I think we can merge it now, and support ZeRO-2 later.

feifeibear (approved):
LGTM

feifeibear:
Merge the latest develop. You unittest is too time-consuming. I have remove some less important unittests.

FrankLeeeee:
@ver217 You can change the timeout limit in CI temporarily. I will refactor the tests after zero benchmarking today.

ver217:
> @ver217 You can change the timeout limit in CI temporarily. I will refactor the tests after zero benchmarking today.

I increased the timeout limit in CI temporarily. Please decrease it later.

PR #356, opened 2022-03-09
[profiler] add MemProfiler

raejaf:
TODOs:

- [x] offline implementation
    - [x] save to a specific file
    - [x] show results in CLI
    - [x] show on TensorBoard
- [ ] realtime implementation
    - [x] trainer hook
    - [x] save data


colossalai/engine/_base_engine.py
```diff
@@ -79,6 +84,20 @@ def criterion(self):
         """Criterion attached to the engine"""
         return self._criterion
 
+    def add_hook(self, ophook: Type[BaseOpHook]) -> None:
+        """add necessary hook"""
+        # whether this hook exist
+        for h in self._ophook_list:
+            if type(h) == type(ophook):
+                # TODO: what error should be raised
+                raise NotImplementedError
```
feifeibear, line 93: why did the hook exist you let the program fail? why not just return and give a warning?

raejaf: You are right. Giving users a warning would be much better. I will fix it after I test profiler usage.

> why did the hook exist you let the program fail? why not just return and give a warning?

You are right. I will fix it after I finish the overall profiler interface test.

raejaf: This error has been altered to warning. You could review 9fe8018d02242b9b919755df90f64d0f41a1b916 for more info.

raejaf:
The display of TB is shown as follow, image

feifeibear (approved):

PR #360, opened 2022-03-10
[zero] global model data memory tracer

feifeibear:
The model data memory tracer can trace model data memory usage changes in CPU and GPU during training.
It can be used to design a better offload strategy.


colossalai/utils/memory_tracer/allocator.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def col_tensor_mem_usage(t: Union[torch.Tensor, ShardedTensor]) -> int:
+    if isinstance(t, ShardedTensor):
+        target = t.payload
+    else:
+        target = t
+    unit_in_byte = 0
+    if target.dtype == torch.half:
+        unit_in_byte = 2
```
ver217, line 15: I think we can use `target.element_size()` here.

feifeibear: Good idea, DONE

ver217 (approved):

PR #371, opened 2022-03-10
[zero] used bucketize tensor copy in ShardedOptimv2

feifeibear:


ver217:
Could you resolve the conflicts? And then I will benchmark this branch.

PR #392, opened 2022-03-11
refactored fp16 optimizer

FrankLeeeee:


FrankLeeeee:
I have verified convergence on ResNet50 and CIFAR10 dataset with bs=128 and epochs = 100. The accuracy is shown below.

| Torch DDP (fp32) | ColossalAI FP16 | 
| -                             | -                           |
| 85.4                       | 85.5                     |

FrankLeeeee:
@kurisusnowdeng , this PR has been approved in #392 , but I dropped it out from the commit log to resolve the conflict occured last Friday.

PR #393, opened 2022-03-11
Added tensor detector

Gy-Lu:
Added tensor detector
The unittest and readme are attached as well, the result of unittest is in readme.


colossalai/utils/tensor_detector/tensor_detector.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+LINE_WIDTH = 108
+
+class TensorDetector():
+    def __init__(self,
+                 include_cpu: bool = False,
+                 module: Optional[nn.Module] = None
+                 ):
+        """This class is an dectector to detect tensor on different devices.
```
feifeibear, line 16: is a detector

feifeibear, line 16: is a detector

colossalai/utils/tensor_detector/tensor_detector.py
```diff
@@ -0,0 +1,132 @@
...
+        
+        :param include_cpu: whether to detect tensor on cpu, default False
+        :type include_cpu: bool
+        :param module: when sending an `nn.Module` it, the detector can name the tensors detected better
+        :type module: Optional[nn.Module]
+
+        """
+        self.include_cpu = include_cpu 
+        self.tensors = defaultdict(list)
```
feifeibear, line 25: self.tensor_info will be a better name

feifeibear, line 25: self.tensor_info will be a better name

colossalai/utils/tensor_detector/tensor_detector.py
```diff
@@ -0,0 +1,132 @@
...
+            print(f"Totle GPU Memery Allocated on {device} is {gpu_mem_alloc}")
+        print('-' * LINE_WIDTH)
+        print('\n\n')
+    
+    def detect(self):
+        self.collect_tensors_state()
+        self.print_tensors_state()
+        self.tensors.clear()
+        self.pre_order = self.order.copy()
+        self.order = []
```
feifeibear, line 132: did you clear the self.devices ?

feifeibear, line 132: did you clear the self.devices ?

colossalai/utils/tensor_detector/readme.md
```diff
@@ -0,0 +1,100 @@
...
+detector = TensorDetector(include_cpu=False, module=model)
+detector.detect()
+
+out = model(data)
+
+detector.detect()
+
+loss = out.sum()
+loss.backward()
+
```
feifeibear, line 43: what if I delete or change the model here.
```del model``` or ```model = `MLP().cuda()``

colossalai/utils/tensor_detector/tensor_detector.py
Gy-Lu: Oh...Thanks

colossalai/utils/tensor_detector/tensor_detector.py
Gy-Lu: ok, I'll rename it.

colossalai/utils/tensor_detector/tensor_detector.py
Gy-Lu: no, in fact, I don't think it is necessary. Seems the total devices would not change during running.

colossalai/utils/tensor_detector/readme.md
Gy-Lu: I'll try.

colossalai/utils/tensor_detector/tensor_detector.py
feifeibear: OK

colossalai/utils/tensor_detector/readme.md
Gy-Lu: I found that this tool refers to the module, which leads to an unreleased state. I have added a `close` method to solve the problem.

feifeibear (approved):

PR #403, opened 2022-03-14
[zero] Add bucket tensor shard strategy

ver217:
Add a bucket tensor shard strategy, which will gather a group of sharded tensor together. For naive tensor shard strategy, `bias` will be gathered seperately, which doesn't utilize bandwidth well. For bucket tensor shard strategy, we can pack `weight` and `bias` and gather them together.


FrankLeeeee (approved):

FrankLeeeee:
LGTM.

PR #409, opened 2022-03-14
[hotfix] fixed error when no collective communication in CommProfiler

1SAA:


colossalai/utils/profiler/comm_profiler.py
```diff
@@ -114,6 +114,9 @@ def append(s: str = None):
             append("Warnning: there exists multiple communication operations in the same time. As a result, "
                    "the profiling result is not accurate.")
 
+        if self.total_cuda_time == 0:
+            return "No collective communication has been called yet!"
+
```
FrankLeeeee, line 119: Why return a string?

FrankLeeeee (approved):

PR #420, opened 2022-03-15
[zero] hybrid adam and place part of OS on cuda

feifeibear:


colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -111,23 +111,46 @@ def __init__(self,
...
+        self._first = True
 
     def step(self, *args, **kwargs):
+
+        # When the first time enter the step(), we calculate the margin space for OS
+        # If margin space is enough, part of param fp32 are initialized on GPU.
+        if self._first:
+            self._first = False
+            cuda_margin_space = self.model.cuda_margin_space
+            # div 3 because the space is shared with param fp32, M and V
```
ver217, line 123: What about grad fp32?

feifeibear: OK, as we discussed. If we alway keep a grad fp32, it should be div 4 here.

colossalai/nn/optimizer/cpu_adam.py
```diff
@@ -51,47 +71,64 @@ def step(self, closure=None):
...
+                        state['exp_avg'] = torch.zeros_like(p.data, dtype=torch.float, device=cuda_device)
+                        # gradient variances
+                        state['exp_avg_sq'] = torch.zeros_like(p.data, dtype=torch.float, device=cuda_device)
+                        # memory_format=torch.preserve_format)
+
+                    state['step'] += 1
+                    beta1, beta2 = group['betas']
+
+                    # FIXME() prepare grad on cuda
+                    if p.grad.device.type == 'cpu':
```
Gy-Lu, line 119: Would `p` and `p.grad` have different devices?

colossalai/nn/optimizer/cpu_adam.py
```diff
@@ -51,47 +71,64 @@ def step(self, closure=None):
...
+
+                if p.device.type == 'cpu':
+                    # State initialization
+                    # intended device for step
+                    cpu_device = torch.device('cpu')
+                    if len(state) == 0:
+                        state['step'] = 0
+
+                        # gradient momentums
+                        state['exp_avg'] = torch.zeros_like(p.data, dtype=torch.float, device=cpu_device)
```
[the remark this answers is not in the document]
feifeibear: https://pytorch.org/docs/stable/generated/torch.zeros_like.html
you can change the device.

colossalai/nn/optimizer/cpu_adam.py
feifeibear: Yes, now p.grad is always on cpu.

colossalai/zero/sharded_optim/sharded_optim_v2.py
Gy-Lu: ok, I get it.

colossalai/nn/optimizer/cpu_adam.py
Gy-Lu: ok, I get it.

colossalai/nn/optimizer/cpu_adam.py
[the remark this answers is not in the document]
Gy-Lu: I mean, it is unnecessary to specify the device, and you can reduce the redundant code in `if...else` by using the default value.

PR #421, opened 2022-03-15
added model synchronization for MoE model

1SAA:


colossalai/utils/common.py
```diff
@@ -68,6 +66,25 @@ def sync_model_param(model, parallel_mode):
...
+        for param in model.parameters():
+            if hasattr(param, 'moe_param'):
+                continue
+            ranks = gpc.get_ranks_in_group(ParallelMode.DATA)
+            dist.broadcast(param, src=ranks[0], group=gpc.get_group(ParallelMode.DATA))
+        if moe_env.data_parallel_size > 1:
+            for param in model.parameters():
+                if hasattr(param, 'moe_param'):
+                    ranks = gpc.get_ranks_in_group(ParallelMode.MOE_DATA)
+                    dist.broadcast(param, src=ranks[0], group=gpc.get_group(ParallelMode.MOE_DATA))
```
feifeibear, line 85: Whether to consider compatibility with MOE and ZeRO.
For ZeRO param.data dose not store real payload

feifeibear: if hasattr(param, 'col_attr')

PR #422, opened 2022-03-15
[zero] use double buffers to handle grad

ver217:
In order to support partial offload, we use double buffers to handle grad (`fp16_grad` and `fp32_grad`). `fp16_grad` is always on GPU, and `fp32_grad` may be partially on GPU and CPU.

### DIFF
--- a/colossalai/zero/sharded_model/sharded_model_v2.py
+++ b/colossalai/zero/sharded_model/sharded_model_v2.py
@@ -1,5 +1,5 @@
 import functools
-from ast import Try
+from asyncio.log import logger
 from collections import OrderedDict
 from typing import Any, Optional
 
@@ -21,7 +21,7 @@ from colossalai.zero.sharded_param import ShardedParamV2
 from torch.distributed import ProcessGroup
 from torch.nn.parameter import Parameter
 
-from ._zero3_utils import (cast_float_arguments, cast_tensor_to_fp16, cast_tensor_to_fp32, chunk_and_pad,
+from ._zero3_utils import (cast_float_arguments, cast_tensor_to_fp16, cast_tensor_to_fp32, chunk_and_pad, free_storage,
                            get_gradient_predivide_factor)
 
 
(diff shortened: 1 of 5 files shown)


colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -216,23 +224,7 @@ def _reduce_scatter_callback(self, param: Parameter, reduced_grad: torch.Tensor)
...
-
-        if param.col_attr.grad is None:
-            param.col_attr.grad = reduced_grad.data
-        else:
-            # When dp size = 1
-            # param.col_attr.grad is local accumulated grad shard (full but flatten)
-            # But reduced_grad here is full grad
-            # We should call `view_as`
-            param.col_attr.grad.add_(reduced_grad.data.view_as(param.col_attr.grad))
+        param.col_attr.fp16_grad = reduced_grad.data
```
feifeibear, line 227: do you need to remove param.grad?

feifeibear:
Did you correctly manage the lifecycle of grad fp16? Because it can be released after it is copied to grad fp32.

colossalai/zero/sharded_model/sharded_model_v2.py
ver217: Yes. But we cannot set `param.grad` to `None` in parameters' hook. I tried to allocate an empty tensor and then free its storage before returning it, but unit tests failed. I'm debugging this problem.

ver217:
> Did you correctly manage the lifecycle of grad fp16? Because it can be released after it is copied to grad fp32.

Yes. It's set to `None` after being copied.

feifeibear:
Now the adam use grad fp32 to update? Did you implement it?

ver217:
> Now the adam use grad fp32 to update? Did you implement it?

Yes. It always uses fp32 grad.

colossalai/zero/sharded_model/sharded_model_v2.py
feifeibear: set to None is enough

feifeibear (approved):

PR #423, opened 2022-03-15
Added ViLT-MLM model

ExtremeViscent:


model_zoo/vilt/vilt.py
```diff
@@ -0,0 +1,227 @@
...
+            intermediate_size=hidden_size * mlp_ratio,
+            max_position_embeddings=max_sequence_length,
+            hidden_dropout_prob=dropout,
+            attention_probs_dropout_prob=attention_dropout,
+        )
+
+        self.pooler = Pooler(hidden_size)
+        self.token_type_embeddings = nn.Embedding(2, hidden_size)
+        self.token_type_embeddings.apply(init_weights)
+        self.text_embedding = BertEmbeddings(bert_config)
```
feifeibear, line 63: If the model zoo is designed to run with all kinds of colossal model parallelism.
Should you implement a BertEmbedding using the `PatchEmbedding` from `from colossalai import nn as col_nn`?
https://github.com/hpcaitech/ColossalAI/blob/main/model_zoo/vit/vit.py#L65
@kurisusnowdeng @FrankLeeeee  any comments?

FrankLeeeee: Yes, I think so

Issue #425, opened 2022-03-16
some bugs when running colossalai benckmark after colossalai version update


number1roy:
### 🐛 Describe the bug

1. for gpt benckmark, using gpt2-pp1d config:
![benchmark_pp1d](https://user-images.githubusercontent.com/82053485/158496799-5cd5e9b2-30ca-4dd2-840f-1f2b48b511b7.png)

2. for gpt benckmark, using gpt2-3d config:
![benchmark_3d](https://user-images.githubusercontent.com/82053485/158496826-69291fde-da96-41a5-a1e9-5e340e686956.png)

3. some codes need to be adapted after the version update
https://github.com/hpcaitech/ColossalAI/blob/main/colossalai/engine/schedule/_pipeline_schedule.py#L382
https://github.com/hpcaitech/ColossalAI-Examples/blob/5345187ad55e8c80c111e0c5f7ad9b9241e8f913/language/gpt/train_gpt.py#L47

4. codes in example and benchmark need to be adapted to new ZeRO after the version update or give a tutorial about using new ZeRO.

feifeibear:
Thanks for your comments. We will fix these issues ASAP!

FrankLeeeee:
@ver217 The first 2 bugs were fixed. Do refactor the 3-4 points mentioned by @number1roy when zero is ready.

ver217:
> @ver217 The first 2 bugs were fixed. Do refactor the 3-4 points mentioned by @number1roy when zero is ready.

I'm updating benchmark. And we can fix the 3rd issue in #431

FrankLeeeee:
I feel that we should extract the model_zoo to an independent repository ASAP. I see that benchmark  and examples use different GPT implementation.

ver217:
> I feel that we should extract the model_zoo to an independent repository ASAP. I see that benchmark and examples use different GPT implementation.

Agree.

feifeibear:
> I feel that we should extract the model_zoo to an independent repository ASAP. I see that benchmark and examples use different GPT implementation.

Can not agree more!

PR #430, opened 2022-03-16
fixed gpt attention mask in pipeline

FrankLeeeee:
This PR fixes the first problem mentioned in #425 . The bug occurs because the attention mask is not processed on pipeline stages whichever is not the first stage.


feifeibear (approved):
LGTM

PR #431, opened 2022-03-16
[refactory] refactory the initialize method for new zero design

feifeibear:
A very ugly refactoring to make ```initialize``` supports zero init. Change the type format of existing APIs.


tests/test_zero_data_parallel/test_zero_init_v2.py
```diff
@@ -0,0 +1,74 @@
...
+        engine.train()
+        i = 0
+        for data, label in train_dataloader:
+            if i > 5:
+                break
+
+            data, label = data.cuda(), label.cuda()
+
+            engine.zero_grad()
+            if criterion:
```
feifeibear, line 58: @FrankLeeeee can we use engine in this way?

FrankLeeeee: Engine can be used in this way even with current implementation as criterion can be None now.

colossalai/initialize.py
```diff
@@ -392,7 +396,7 @@ def initialize(model: nn.Module,
         gradient_handlers = [build_gradient_handler(cfg, model, optimizer) for cfg in gradient_handler_cfg]
 
     # check if optimizer is ColossalaiOptimizer
-    if not isinstance(optimizer, (ColossalaiOptimizer, ShardedOptimizer)):
+    if not isinstance(optimizer, (ColossalaiOptimizer, ShardedOptimizerV2)):
         optimizer = ColossalaiOptimizer(optim=optimizer)
 
     # gradient accumulation
```
feifeibear, line 402: @ver217 @FrankLeeeee 
can ```accumulate_gradient``` work together with zero?

FrankLeeeee: I think it is compatible. accumulate_gradient only skips the optimizer.step() call until accumulate_size is reached.

feifeibear: cool

colossalai/initialize.py
```diff
@@ -267,10 +270,27 @@ def initialize(model: nn.Module,
...
-    use_zero3 = hasattr(gpc.config, 'zero') and gpc.config.zero.level == 3
-    if not moe_env.is_initialized() and not use_zero3:
+    use_zero = hasattr(gpc.config, 'zero')
+    if use_zero:
+        zero_cfg = gpc.config.get('zero', None)
+        if zero_cfg is not None:
+            cfg_ = zero_cfg.copy()
+        else:
+            cfg_ = {}
+        model, optimizer = convert_to_zero_v2(model_builder=model, optimizer_class=torch.optim.Adam, **cfg_)
```
ver217, line 280: Why fix `optimizer_class` to `torch.optim.Adam`?

colossalai/zero/__init__.py
```diff
@@ -7,6 +10,43 @@
...
+    :type model: :class:`torch.nn.Module`
+    :param optimizer: Your optimizer object
+    :type optimizer: :class:`torch.optim.Optimizer`
+    :param zero_config: Configuration for zero
+    :type zero_config: dict
+
+    :return: (model, optimizer)
+    :rtype: Tuple
+    """
+    # FIXME() can we use gpc here?
```
ver217, line 33: Yes, we can.

colossalai/zero/__init__.py
```diff
@@ -7,6 +10,43 @@
...
+        with ZeroInitContext(convert_fp16='fp16' in gpc.config,
+                             target_device=torch.cuda.current_device(),
+                             shard_strategy=stard_strategy,
+                             shard_param=True):
+            model = model_builder()
+    else:
+        raise TypeError(f"convert_to_zero_v2 dose not support model_builder of type {type(convert_to_zero_v2)}")
+
+    zero_model = ShardedModelV2(model, shard_strategy=stard_strategy)
+    zero_optimizer = ShardedOptimizerV2(zero_model, optimizer_class, lr=1e-3)
```
ver217, line 48: Why fix `lr` to `1e-3`?

colossalai/initialize.py
```diff
@@ -324,7 +338,7 @@ def initialize(model: nn.Module,
         # 1. if optimizer is ZERO, then use zero grad handler
         # 2. if dp size is larger than 1 and pipeline is not used, use pytorch ddp
         # 3. if using pipeline and dp size larger than 1, use data parallel grad handler
-        if isinstance(optimizer, ShardedOptimizer):
+        if isinstance(optimizer, ShardedOptimizerV2):
             gradient_handler_cfg = [dict(type='ZeROGradientHandler')]
```
ver217, line 342: Do we still need `ZeROGradientHandler` now?

colossalai/zero/__init__.py
feifeibear: wait, I have to add a ADAM config

colossalai/zero/__init__.py
feifeibear: thx

colossalai/initialize.py
feifeibear: pass it via config

colossalai/initialize.py
FrankLeeeee: I don't think so.

colossalai/zero/__init__.py
feifeibear: DONE

colossalai/initialize.py
feifeibear: Keep it now, remove if necessary.

FrankLeeeee: Keeping it will lead to exception as this gradient handler requires the optimizer to have a `sync_grad` method.

feifeibear: OK, I have already added such a method in ShardedOptimV2

ver217 (approved):

PR #438, opened 2022-03-17
[unitest] polish zero config in unittest

feifeibear:


tests/test_zero_data_parallel/common.py
```diff
@@ -10,6 +10,18 @@
 
 LOGGER = get_dist_logger()
 
+_ZERO_OPTIMIZER_CONFIG = dict(optimizer_type=torch.optim.Adam, optimizer_config=dict(lr=1e-3))
```
FrankLeeeee, line 13: Do we need to follow the previous schema? The conventional schema we used previous is like:

```python
dict(
    type="Adam",
    lr=1e-3
)
```

The optimizer is registered in a registry and the build function will build the object via this schema

feifeibear: OK, I will improve it in another MR. I think a reactor implementation is needed to map the name to optimzer_cls.

feifeibear: I see. you mean I can use the global registry OPTIMIZERS. I will fix it in another MR ASAP.

FrankLeeeee: OK.

FrankLeeeee: Some registries are deprecated, I will remove them in another MR later.

FrankLeeeee (approved):

PR #445, opened 2022-03-17
[zero] hybrid cpu adam

feifeibear:
Make our customized CPUADAM more general. It did not run it on CUDA. Now it is possible.

### DIFF
--- a/colossalai/nn/optimizer/cpu_adam.py
+++ b/colossalai/nn/optimizer/cpu_adam.py
@@ -51,7 +51,6 @@ class CPUAdam(torch.optim.Optimizer):
                           loss_scale,
                           use_adamw=False):
         if loss_scale is not None:
-            # grad.div_(loss_scaler.loss_scale)
             grad.div_(loss_scale)
 
         if weight_decay != 0:
--- a/colossalai/nn/optimizer/cpu_adam.py
+++ b/colossalai/nn/optimizer/cpu_adam.py
@@ -1,4 +1,5 @@
 import torch
+import math
 
 
 class CPUAdam(torch.optim.Optimizer):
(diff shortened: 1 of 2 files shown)


colossalai/nn/optimizer/cpu_adam.py
```diff
@@ -31,18 +31,46 @@ def __init__(self,
...
+                          beta1,
+                          beta2,
+                          eps,
+                          weight_decay,
+                          bias_correction1,
+                          bias_correction2,
+                          loss_scale,
+                          use_adamw=False):
+        if loss_scale is not None:
+            # grad.div_(loss_scaler.loss_scale)
```
FrankLeeeee, line 54: remove this line

colossalai/nn/optimizer/cpu_adam.py
```diff
@@ -31,18 +31,46 @@ def __init__(self,
...
-                                     eps,
-                                     weight_decay,
-                                     adamw_mode,
-                                     simd_log)
+        self.cpu_adam_op.create_adam(self.opt_id, lr, betas[0], betas[1], eps, weight_decay, adamw_mode, simd_log)
 
     def __del__(self):
         self.cpu_adam_op.destroy_adam(self.opt_id)
 
+    def torch_adam_update(self,
```
ver217, line 39: Can we just use original `torch.optim.adam.adam` ?

feifeibear: I try to import the function, it failed
```
from torch.optim._functional.adam
```
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'torch.optim._functional.adam'; 'torch.optim._functional' is not a package
https://github.com/pytorch/pytorch/blob/0fc6bd2e47/torch/optim/_functional.py#L54

ver217: OK. I see.

FrankLeeeee (approved):

Issue #443, opened 2022-03-17
[DevOps] CI takes too long


FrankLeeeee:
### Describe the problem

Now we face major bottlenecks in our CI time. A single build CI can take 15 to 20 min. There is an urgent need to reduce the total CI time. Some changes need to be applied.

1. Change git clone into ssh mode. The git checkout action can take more than 5 mins sometimes. This is absurdly long and perhaps it is due to https connection to github.

2. Shorten unit testing time. This can be done by reducing the number of distributed network initialization and lowering the test problem size.

Some profiling results of the unit tests are shown below. The most time-consuming test cases will be optimised with priority.

<img width="557" alt="Screenshot 2022-03-17 at 11 22 45 AM" src="https://user-images.githubusercontent.com/31818963/158731356-a8117410-2717-4f74-90a7-281fd4ae77f0.png">

FrankLeeeee:
This has been fixed by #446 , #448 , #449 , and #452 .

PR #449, opened 2022-03-17
[DevOps] remove unneeded dependency in build workflow

FrankLeeeee:
This is part of the efforts to tackle #443  
- Remove `apex` and `deepspeed` from the dependency list in the build workflow so that they are no longer installed on the fly in the CI. `deepspeed` is no longer our dependency and `apex` is built-in in the Nvidia NGC docker.
- configure the checkout action to use ssh to shorten checkout time


feifeibear (approved):
LGTM

PR #475, opened 2022-03-21
[test] added rerun on exception for testing

FrankLeeeee:
- Added `rerun_on_exception` function to re-run a method upon detection of an exception. This is useful for the `Address already in use` error in our CI.
- Fixed the docstring of `parameterize` function.

### DIFF
--- a/colossalai/testing/utils.py
+++ b/colossalai/testing/utils.py
@@ -117,13 +117,10 @@ def rerun_on_exception(exception_type: Exception = Exception, pattern: str = Non
                     try_count += 1
                     func(*args, **kwargs)
                 except exception_type as e:
-                    if isinstance(e, exception_type):
-                        if pattern is None or re.match(pattern, str(e)):
-                            # when pattern is not specified, we always skip the exception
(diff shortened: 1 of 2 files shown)


colossalai/testing/utils.py
```diff
@@ -11,42 +12,121 @@ def parameterize(argument: str, values: List[Any]):
...
+
+        def _run_until_success(*args, **kwargs):
+            try_count = 0
+            assert max_try is None or isinstance(max_try, int), \
+                f'Expected max_try to be None or int, but got {type(max_try)}'
+
+            while max_try is None or try_count < max_try:
+                try:
+                    try_count += 1
+                    func(*args, **kwargs)
```
feifeibear, line 118: What if the func has already handled the exception. Then no exception will be caught here.
Does the wrapper only work for test func without exception handling?

FrankLeeeee: If the exception is handled by internal code, then we should not catch it. We only need to handle those which make the program stop.

feifeibear (approved):

PR #480, opened 2022-03-21
add LinearGate for MOE in NaiveAMP context

1SAA:


colossalai/nn/layer/moe/layers.py
```diff
@@ -216,6 +215,31 @@ def forward(self, inputs: torch.Tensor, use_kernel: bool = False, ep_group: Opti
             return cb_weight, sec_mask
 
 
+class LinearGate(nn.Module):
```
feifeibear, line 218: why not derive from torch.nn.Linear. 
set bias == False.

feifeibear (approved):

colossalai/nn/layer/moe/layers.py
```diff
@@ -216,6 +216,23 @@ def forward(self, inputs: torch.Tensor, use_kernel: bool = False, ep_group: Opti
             return cb_weight, sec_mask
 
 
+class LinearGate(nn.Linear):
```
feifeibear, line 219: @1SAA change the name to FP32LinearGater

1SAA: done

PR #481, opened 2022-03-21
[format] polish name format for MOE

feifeibear:


colossalai/nn/layer/moe/experts.py
```diff
@@ -34,7 +34,7 @@ class Experts(MoeExperts):
     :type num_experts: int
     """
 
-    def __init__(self, expert, num_experts, **expert_args):
+    def __init__(self, expert: nn.Module, num_experts: int, **expert_args):
```
1SAA, line 37: Here parameter `expert` is not a nn.Module instance. It is just a class name, the expert instance will be constructed later.

feifeibear: OK, its a TYPE[nn.Module]

see https://docs.python.org/3/library/typing.html

1SAA (approved):

PR #486, opened 2022-03-22
[zero] sharded optim support hybrid cpu adam

ver217:


feifeibear (approved):

colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -122,6 +110,20 @@ def __init__(self,
                     self.shard_strategy.gather([p.col_attr.data], self.dp_process_group)
 
     def step(self, *args, **kwargs):
+        if self._should_move_fp32_shards_h2d:
+            self._should_move_fp32_shards_h2d = False
```
feifeibear, line 114: Can we move these logic to an self.func...
You can polish it in another MR.

ver217: OK

PR #487, opened 2022-03-22
[test] fixed rerun_on_exception and adapted test cases

FrankLeeeee:
- fixed two bugs in `rerun_on_exception` function. The first is that it will run many times even though the function works fine. The second is that it is not compatible with `pytest.mark.parameterize` due to signature conflict.
- adapted the test cases to use `rerun_on_exception` decorator so that `address already in user` error can be avoided.

### DIFF
--- a/colossalai/testing/utils.py
+++ b/colossalai/testing/utils.py
@@ -1,6 +1,7 @@
 import re
 from typing import Callable, List, Any
 from functools import partial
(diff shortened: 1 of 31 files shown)


feifeibear (approved):

ver217:
Some unit tests are updated in other pull requests. Could you rebase main branch?

PR #491, opened 2022-03-22
add cpu adamW

Gy-Lu:
Added cpu AdamW.
Separate the class `CPUAdam` and `CPUAdamW`, users can not defined an adamw by passing a parameter `adamw_mode` when init.

### DIFF
--- a/colossalai/nn/optimizer/cpu_adamw.py
+++ b/colossalai/nn/optimizer/cpu_adamw.py
@@ -1,13 +1,9 @@
+import torch
 import math
 
-import torch
 
-
-class CPUAdam(torch.optim.Optimizer):
+class CPUAdamW(torch.optim.Optimizer):
     optimizer_id = 0
-    # Number of fp32 shards for per parameter
-    # Param weight, grad, momentum and variance
-    num_fp32_shards_per_param = 4
 
     def __init__(self,
                  model_params,
@@ -25,10 +21,10 @@ class CPUAdam(torch.optim.Optimizer):
         """
 
         default_args = dict(lr=lr, betas=betas, eps=eps, weight_decay=weight_decay, bias_correction=bias_correction)
-        super(CPUAdam, self).__init__(model_params, default_args)
-        self.opt_id = CPUAdam.optimizer_id
-        CPUAdam.optimizer_id = CPUAdam.optimizer_id + 1
-        adamw_mode = False
+        super(CPUAdamW, self).__init__(model_params, default_args)
+        self.opt_id = CPUAdamW.optimizer_id
+        CPUAdamW.optimizer_id = CPUAdamW.optimizer_id + 1
+        adamw_mode = True
         self.loss_scale = loss_scale
         try:
             import cpu_adam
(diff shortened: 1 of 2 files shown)


colossalai/nn/optimizer/cpu_adamw.py
```diff
@@ -0,0 +1,124 @@
+import torch
+import math
+
+
+class CPUAdamW(torch.optim.Optimizer):
+    optimizer_id = 0
+
```
ver217, line 7: Can we set `num_fp32_shards_per_param` to support hybrid optimization?

colossalai/nn/optimizer/cpu_adamw.py
```diff
@@ -0,0 +1,124 @@
...
+                if target_device.type == 'cpu':
+                    assert state['exp_avg'].device.type == 'cpu', "exp_avg should stay on cpu"
+                    assert state['exp_avg_sq'].device.type == 'cpu', "exp_avg should stay on cpu"
+                    self.cpu_adam_op.adam_update(self.opt_id, state['step'], group['lr'], beta1, beta2, group['eps'],
+                                                 group['weight_decay'], group['bias_correction'], p.data, p.grad.data,
+                                                 state['exp_avg'], state['exp_avg_sq'], self.loss_scale)
+                elif target_device.type == 'cuda':
+                    # FIXME() prepare grad on cuda
+                    if p.grad.device.type == 'cpu':
```
ver217, line 109: You can remove this.

colossalai/nn/optimizer/cpu_adamw.py
Gy-Lu: OK, I forget to update the code.

colossalai/nn/optimizer/cpu_adamw.py
```diff
@@ -0,0 +1,124 @@
...
+        except ImportError:
+            raise ImportError('Please install colossalai from source code to use CPUAdam')
+        self.cpu_adam_op = cpu_adam
+        self.cpu_adam_op.create_adam(self.opt_id, lr, betas[0], betas[1], eps, weight_decay, adamw_mode, simd_log)
+
+    def __del__(self):
+        if self.cpu_adam_op:
+            self.cpu_adam_op.destroy_adam(self.opt_id)
+
+    def torch_adamw_update(self,
```
ver217, line 40: Is it possible for us to use fused `Adam` here? @feifeibear @Gy-Lu

Gy-Lu: ok, I'll try :)

colossalai/nn/optimizer/cpu_adamw.py
Gy-Lu: Done

colossalai/nn/optimizer/cpu_adamw.py
```diff
@@ -0,0 +1,124 @@
+import torch
+import math
+
+
+class CPUAdamW(torch.optim.Optimizer):
```
feifeibear, line 5: why not derive from CPUAdam?
Or CPUAdam and CPUAdamW both derive from CPUAdamBase...
A lot of duplicated code between CPUAdam and CPUAdamW

Gy-Lu: Thanks for the advice, I'll consider it.

PR #492, opened 2022-03-22
[zero] sharded model manages ophooks individually

ver217:
`ShardedModel` manages ophooks individually. We can utilize full features of ophooks even without `Engine`.

### DIFF
--- a/colossalai/zero/sharded_model/sharded_model_v2.py
+++ b/colossalai/zero/sharded_model/sharded_model_v2.py
@@ -89,8 +89,8 @@ class ShardedModelV2(nn.Module):
         self._iter_cnter = 0
 
         # Register hooks
-        register_ophooks_recursively(self.module,
-                                     [ZeroHook(self.shard_strategy, self._memstats_collector, self.process_group)])
+        self._ophook_list = [ZeroHook(self.shard_strategy, self._memstats_collector, self.process_group)]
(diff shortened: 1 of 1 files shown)


colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -134,10 +134,14 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
...
         loss.backward()
         self._post_backward_operations()
+        for ophook in self._ophook_list:
+            ophook.post_iter()
 
     def backward_by_grad(self, tensor, grad):
         torch.autograd.backward(tensors=tensor, grad_tensors=grad)
         self._post_backward_operations()
+        for ophook in self._ophook_list:
+            ophook.post_iter()
```
feifeibear, line 145: why not move then to _post_backward_operations?

ver217: I think `_post_backward_operations()` just contains `ZeRO`'s own logic, so I just leave this code snippet here.

ver217: This code snippet may be more general than `_post_backward_operations()`, I think.

feifeibear: OK

feifeibear (approved):

PR #497, opened 2022-03-23
[cuda] modify the fused adam, support hybrid of fp16 and fp32

Gy-Lu:
now our fused adam supports the hybrid of p and d dtype.

### DIFF
--- a/tests/test_optimizer/unittest_cpu_adam.py
+++ b/tests/test_optimizer/unittest_cpu_adam.py
@@ -1,38 +1,7 @@
-# BSD 3-Clause License
-#
-# Copyright (C) 2021 THL A29 Limited, a Tencent company.  All rights reserved.
-#
-# Redistribution and use in source and binary forms, with or without modification,
-# are permitted provided that the following conditions are met:
-#
-#  * Redistributions of source code must retain the above copyright notice, this
-#    list of conditions and the following disclaimer.
-#
-#  * Redistributions in binary form must reproduce the above copyright notice,
-#    this list of conditions and the following disclaimer in the documentation
-#    and/or other materials provided with the distribution.
-#
-#  * Neither the name of the psutil authors nor the names of its contributors
-#    may be used to endorse or promote products derived from this software without
-#    specific prior written permission.
-#
-# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
-# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
-# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
-# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
-# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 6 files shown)


tests/test_optimizer/unittest_fused_adam.py
```diff
@@ -0,0 +1,68 @@
...
+        y = model_copy(data_copy[i])
+        loss = ((label_copy[i] - y) ** 2).sum()
+        torch_optim.zero_grad()
+        loss.backward()
+        torch_optim.step()
+
+    assert len(optim.param_groups[0]['params']) == len(torch_optim.param_groups[0]['params'])
+    max_diff = 0
+    for i in range(len(optim.param_groups[0]['params'])):
+        diff = torch.max(torch.abs(optim.param_groups[0]['params'][i]-(torch_optim.param_groups[0]['params'][i])))
```
ver217, line 60: Is it possible to use `torch.allclose()`?

Gy-Lu: yes, thanks for your advice.

tests/test_optimizer/unittest_fused_adam.py
```diff
@@ -0,0 +1,68 @@
...
+        y = model(data[i])
+        loss = ((label[i] - y) ** 2).sum()
+        optim.zero_grad()
+        loss.backward()
+        if p_dtype != g_dtype:
+            for i in range(len(optim.param_groups[0]['params'])):
+                optim.param_groups[0]['params'][i].grad.data = optim.param_groups[0]['params'][i].grad.data.to(g_dtype)
+        optim.step()
+
+    for i in range(len(data_copy)):
```
ver217, line 50: Can we extract this part and write a function which both optimizers can use?

Gy-Lu: Yes, but I am not sure if it is necessary. If we write a function to reuse, there would be some code to judge the states and a lot of parameters to pass.

tests/test_optimizer/unittest_cpu_adam.py
```diff
@@ -121,7 +88,7 @@ def check_res(
             p_grad.view(-1),    # fp32 grad
             exp_avg.view(-1),
             exp_avg_sq.view(-1),
-            loss_scale,
+            -1,
```
feifeibear, line 91: why we set loss_scale as -1?

tests/test_optimizer/unittest_fused_adam.py
```diff
@@ -0,0 +1,66 @@
...
+        loss.backward()
+        torch_optim.step()
+
+    assert len(optim.param_groups[0]['params']) == len(torch_optim.param_groups[0]['params'])
+    # we set the torch adam's dtype always being fp32
+    # when p and g both are fp16, there is a great loss
+    # so we make a loose threshold
+    
+    #threshold = 1e-3
+    print(adamw, p_dtype, g_dtype)
```
feifeibear, line 61: remove this line

tests/test_optimizer/unittest_fused_adam_kernel.py
```diff
@@ -0,0 +1,98 @@
...
+                m_copy,
+                v_copy,
+                -1,
+                adamw,
+            )
+        
+        if torch.isnan(p).any() or torch.isnan(p_copy).any():
+            count += 1
+            continue
+        assert count < 200, "too many nans"
```
feifeibear, line 97: did we allow nan in this check?

Gy-Lu: yes, for fp16, nan seems unavoidable, especially when there are convertions.

tests/test_optimizer/unittest_cpu_adam.py
Gy-Lu: oh, I just forget to change it...
it should be -1 or 2**5

Gy-Lu: Have fixed it.

tests/test_optimizer/unittest_fused_adam.py
Gy-Lu: Done

feifeibear (approved):

PR #498, opened 2022-03-23
[WIP] demonstrate writing parallel op with customized tensors structure

feifeibear:
With the help of [```__torch_function__```](https://pytorch.org/docs/stable/notes/extending.html), we can follow a new paradigm to develop parallel operators.
In this demo, I showed a way to define a customized Linear Op on ShardedTensor. In this way, users do not need to change its model definition while using our customized Op.
We can add Tensor Parallel/ZeRO logic in the customized Linear Op. Therefore the parameter tensor data structure of TP and ZeRO can be unified.

### DIFF
--- a/colossalai/nn/layer/parallel_ops/__init__.py
+++ b/colossalai/nn/layer/parallel_ops/__init__.py
@@ -1,0 +1,1 @@
+from .linear import *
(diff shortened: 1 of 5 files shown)


feifeibear:
@FrankLeeeee @ver217 @ver217 any comments?

ver217:
> @FrankLeeeee @ver217 @ver217 any comments?

What about the compatibility of `__torch_function__`?

PR #500, opened 2022-03-23
[polish] polish singleton and global context

feifeibear:
A standard method to define singleton.
Avoid circle reference of package.


colossalai/context/moe_context.py
```diff
@@ -1,6 +1,9 @@
 import torch
 import torch.distributed as dist
-from .parallel_mode import ParallelMode
+
+from colossalai.context.parallel_mode import ParallelMode
```
1SAA, line 4: I think the old way is better, since it has less probability to cause reference error.
When you import colossalai.context.parallel_mode, it will call __init__.py in colossalai, 
this may cause reference problem.

colossalai/context/parallel_context.py
```diff
@@ -15,30 +15,16 @@
 
 from .parallel_mode import ParallelMode
 from .random import add_seed, get_seeds, set_mode
+from colossalai.context.singleton_meta import SingletonMeta
```
1SAA, line 18: Is "from .singleton_meta import SingletonMeta" better?

colossalai/context/moe_context.py
feifeibear: No... It won't.

Unfortunately, relative imports can be messy, particularly for shared projects where directory structure is likely to change. Relative imports are also not as readable as absolute ones, and it’s not easy to tell the location of the imported resources.

colossalai/context/parallel_context.py
feifeibear: Remember that you should generally opt for absolute imports over relative ones, unless the path is complex and would make the statement too long.

https://realpython.com/absolute-vs-relative-python-imports/

You can read Pytorch or tensorflow code. They soldom use relative path...

1SAA: ok

1SAA (approved):

PR #503, opened 2022-03-24
[memory] add model data tensor moving api

feifeibear:
The api not only moves tensors but also collects the model data volume change on CUDA.


colossalai/utils/memory_utils/utils.py
```diff
@@ -0,0 +1,59 @@
...
+from colossalai.utils.memory_tracer.model_data_memtracer import GLOBAL_MODEL_DATA_TRACER
+
+from typing import Union
+
+
+def colo_cuda_memory_capacity():
+    """
+    Get cuda memory capacity of the current cuda.
+    """
+    return torch.cuda.get_device_properties(get_current_device()).total_memory
```
ver217, line 13: Is it possible to get allowed memory?

feifeibear: we can fix it in another MR...

ver217 (approved):

PR #506, opened 2022-03-24
[memory] set cuda mem frac

feifeibear:


colossalai/utils/memory_utils/utils.py
```diff
@@ -0,0 +1,88 @@
...
+from colossalai.utils import get_current_device
+from colossalai.zero.sharded_param.sharded_tensor import ShardedTensor
+from colossalai.utils.memory_tracer.model_data_memtracer import GLOBAL_MODEL_DATA_TRACER
+
+from typing import Union
+
+_GLOBAL_CUDA_MEM_FRACTION = 1.0
+
+
+def colo_set_process_memory_fraction(ratio: float):
```
feifeibear, line 11: @ver217  is this what you want?

ver217 (approved):

colossalai/utils/memory_utils/utils.py
ver217: Yes

ver217 (approved):

PR #510, opened 2022-03-24
[MOE] add MOEGPT model

1SAA:


model_zoo/moe/gpt.py
```diff
@@ -0,0 +1,217 @@
...
+        self.attn = GPTSelfAttention(dim=dim,
+                                     num_heads=num_heads,
+                                     attention_dropout=attention_dropout,
+                                     dropout=dropout,
+                                     bias=bias,
+                                     fuse_scale_mask_softmax=fuse_scale_mask_softmax,
+                                     dtype=dtype)
+        self.norm2 = col_nn.LayerNorm(normalized_shape=dim, eps=layernorm_epsilon, dtype=dtype)
+
+        mlp_factory = dict(dim=dim, mlp_ratio=mlp_ratio, activation=activation, dropout=dropout, dtype=dtype, bias=bias)
```
feifeibear, line 43: rename it as mpl_factory_dict

model_zoo/moe/gpt.py
```diff
@@ -0,0 +1,217 @@
...
+                        depth=24,
+                        num_heads=16,
+                        **kwargs)
+    return _create_moegpt_model(**model_kwargs)
+
+
+@MODELS.register_module
+def prmoe_51b(**kwargs):
+    if 'use_residual' in kwargs:
+        kwargs.pop('use_residual')
```
feifeibear, line 210: use_residual = kwargs.pop('use_residual', False) ?

model_zoo/moe/gpt.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def _create_moegpt_model(**model_kwargs):
+    model = MOEGPT(**model_kwargs)
+    return model
+
+
+@MODELS.register_module
+def prmoe_4b(**kwargs):
+    if 'use_residual' in kwargs:
```
feifeibear, line 183: why remove use_residual?

model_zoo/moe/gpt.py
1SAA: In PR-MOE, 'use_residual' keeps True, but I want key word 'use_residual' removed in dict in case that
it has two same key word in below constructor function.

model_zoo/moe/gpt.py
1SAA: done

model_zoo/moe/gpt.py
1SAA: I add warning when there is no 'use_residual' or it's False.

feifeibear (approved):

PR #517, opened 2022-03-25
[refactor] remove old zero code

feifeibear:
Old zero code is confusing for developers. We remove them here.


ver217:
Can we also clean `utils`?

ver217 (approved):

PR #519, opened 2022-03-25
[log] polish disable_existing_loggers

ver217:
1. Polish `disable_existing_loggers()`. Make its API clearer.
2. Rename `colossalai.logging.logging` to `colossalai.logging.logger`. In `colossalai/logging/__init__.py`, `logging` is imported. We actually want to import standard library here, but there is `colossalai/logging/logging.py`, which makes me confused.


kurisusnowdeng (approved):

feifeibear (approved):

FrankLeeeee:
@ver217 , I think @feifeibear  has written a singleton meta class, can you refactored the current singleton logger with his implementation?

ver217:
> @ver217 , I think @feifeibear has written a singleton meta class, can you refactored the current singleton logger with his implementation?

This is a little different from the standard singleton. For loggers, every "name" corresponds to a "singleton". Maybe we can optimize  it in next MR.

FrankLeeeee:
LGTM.

PR #527, opened 2022-03-25
[zero]added hybrid adam, removed loss scale in adam

Gy-Lu:
Added hybrid adam. Passed the test.
removed loss scale in adam which may be against using habit.

### DIFF
--- a/tests/test_optimizer/unittest_hybrid_adam.py
+++ b/tests/test_optimizer/unittest_hybrid_adam.py
@@ -39,6 +39,3 @@ def test_adam(adamw, device, p_dtype, g_dtype):
             continue
         assert torch.allclose(p.data, p_copy.data, 1e-4, 1e-2), \
                               f"adaw mode {adamw}, device {device}, p_dtype {p_dtype}, g_dtype {g_dtype}"
-
-test_adam()
-                                
--- a/tests/test_optimizer/unittest_hybrid_adam.py
+++ b/tests/test_optimizer/unittest_hybrid_adam.py
@@ -1,0 +1,44 @@
+import torch
+import torch.nn as nn
+from torch.optim.adam import Adam
+from torch.optim import AdamW
+
+from colossalai.nn.optimizer.hybrid_adam import HybridAdam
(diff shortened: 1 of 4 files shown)


feifeibear (approved):

tests/test_optimizer/unittest_hybrid_adam.py
```diff
@@ -0,0 +1,44 @@
...
+
+        optim.step()
+        torch_optim.step()
+
+        if torch.isnan(p.data).any() or torch.isnan(p_copy.data).any():
+            continue
+        assert torch.allclose(p.data, p_copy.data, 1e-4, 1e-2), \
+                              f"adaw mode {adamw}, device {device}, p_dtype {p_dtype}, g_dtype {g_dtype}"
+
+test_adam()
```
feifeibear, line 43: if __name__ == '__main__':
    test_adam()

colossalai/nn/optimizer/cpu_adam.py
```diff
@@ -110,7 +105,7 @@ def step(self, closure=None):
                     assert state['exp_avg_sq'].device.type == 'cpu', "exp_avg should stay on cpu"
                     self.cpu_adam_op.adam_update(self.opt_id, state['step'], group['lr'], beta1, beta2, group['eps'],
                                                  group['weight_decay'], group['bias_correction'], p.data, p.grad.data,
-                                                 state['exp_avg'], state['exp_avg_sq'], self.loss_scale)
+                                                 state['exp_avg'], state['exp_avg_sq'], -1)
```
feifeibear, line 108: loss scale why -1

Gy-Lu: We think setting loss scale in adam is against the using habit.

tests/test_optimizer/unittest_hybrid_adam.py
Gy-Lu: Thanks. for pytest, I would remove it.

feifeibear (approved):

PR #528, opened 2022-03-25
[zero] fix grad offload

feifeibear:


colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -198,16 +198,16 @@ def _post_backward_operations(self) -> None:
             # the shape `grad` is the same as unsharded param
             # So we can just use `view(-1)` to ensure grad is a flat tensor shard
             if self.reuse_fp16_shard:
-                grad = p.col_attr.sharded_data_tensor.payload
+                grad_payload = p.col_attr.sharded_data_tensor.payload
             else:
-                grad = cast_tensor_to_fp32(p.col_attr.fp16_grad)
+                grad_payload = cast_tensor_to_fp32(p.col_attr.fp16_grad)
             if p.col_attr.offload_grad:
-                colo_model_data_move_to_cpu(grad)
```
ver217, line 205: I think we don't need to use `tensor.detach().clone()` here.  `tensor.to()` is enough.

feifeibear: OK

ver217 (approved):

PR #535, opened 2022-03-28
[Zero] process no-leaf-module in Zero

1SAA:
At present, we only add Zero hooks for leaf nn.modules.
When a nn.module has both subordinate nn.Module and nn.Parameter, those parameters belonging to itself won't get 
processed.


feifeibear (approved):
LGTM

PR #536, opened 2022-03-28
[zero] get memory usage for sharded param

feifeibear:
A PR towards tracing non-model data memory usage inside an iteration.

### DIFF
--- a/colossalai/zero/sharded_param/sharded_param.py
+++ b/colossalai/zero/sharded_param/sharded_param.py
@@ -1,7 +1,7 @@
 import torch
 import torch.distributed as dist
 from colossalai.zero.sharded_param import ShardedTensor
-from typing import Optional
+from typing import Optional, Tuple
 
 
 class ShardedParamV2(object):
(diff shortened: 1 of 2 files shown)


colossalai/zero/sharded_param/sharded_param.py
```diff
@@ -40,3 +40,28 @@ def sharded_data_tensor(self):
...
+        cuda_mem_use, cpu_mem_use = 0, 0
+
+        def _update_mem_use(t: Optional[torch.Tensor]):
+            if t is None:
+                return
+            assert isinstance(t, torch.Tensor)
+            nonlocal cuda_mem_use
+            nonlocal cpu_mem_use
+            if t.device.type == 'cpu':
+                cpu_mem_use += t.numel() * t.element_size()
```
Gy-Lu, line 59: Seems it would return the number of Byte, do you consider formatting it to KB/MB/GB?

feifeibear: Yeah, we can convert the unit at the time it is used. Currently, we can just make the API as simple as possible.

Gy-Lu: ok

Gy-Lu (approved):

colossalai/zero/sharded_param/sharded_param.py
```diff
@@ -40,3 +40,28 @@ def sharded_data_tensor(self):
...
+            elif t.device.type == 'cuda':
+                cuda_mem_use += t.numel() * t.element_size()
+
+        _update_mem_use(self.sharded_data_tensor.payload)
```
1SAA, line 63: If the tensor is shared but it payload hasn't been removed from CUDA,
I think we should caculate the memory of param.data in this case.

PR #537, opened 2022-03-28
[zero] refactor model data tracing

feifeibear:
I want to depict the cuda memory usage curve of the non-model data in a warmup iteration.
non-model = system cuda usage - model-data.
The old implementation, which has been removed, incrementally traces the model tensor moving, allocating and releasing.
So I can know the volume changing, and derive current model data memory by adding and deleting the changed tensor.
The idea is stupid.
Now, I just scan the model and calculate the model param usage.


ver217 (approved):
During `_post_backward_operations()`, `p.col_attr.fp16_grad` and `p.col_attr.fp32_grad` will be set to None, and `p.grad` contains corrsponding grad. You can watch `p.grad` in `get_memory_usage()` in next MR.

PR #539, opened 2022-03-28
[zero] optimize grad offload

ver217:
When `reuse_fp16_shard` is `True`, grad is stored in `p.col_attr.sharded_data_tensor.payload` and we move all payload to CPU. This reduces max GPU memory usage during backward.

### DIFF
--- a/colossalai/zero/sharded_model/sharded_model_v2.py
+++ b/colossalai/zero/sharded_model/sharded_model_v2.py
@@ -14,16 +14,13 @@ from colossalai.logging import get_dist_logger
 from colossalai.utils.memory_tracer.memstats_collector import MemStatsCollector
 from colossalai.utils.memory_tracer.model_data_memtracer import \
     GLOBAL_MODEL_DATA_TRACER
-from colossalai.utils.memory_utils.utils import (colo_cuda_memory_capacity,
-                                                 colo_model_data_move_to_cpu,
-                                                 colo_model_tensor_clone)
+from colossalai.utils.memory_utils.utils import colo_cuda_memory_capacity
 from colossalai.zero.shard_utils import BaseShardStrategy
 from colossalai.zero.sharded_model.reduce_scatter import ReduceScatterBucketer
 from torch.distributed import ProcessGroup
 from torch.nn.parameter import Parameter
 
-from ._utils import (cast_float_arguments, cast_tensor_to_fp16,
-                     cast_tensor_to_fp32, chunk_and_pad, free_storage,
+from ._utils import (cast_float_arguments, cast_tensor_to_fp16, cast_tensor_to_fp32, chunk_and_pad, free_storage,
                      get_gradient_predivide_factor)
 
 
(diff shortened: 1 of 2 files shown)


colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -161,7 +160,7 @@ def step(self, *args, **kwargs):
                 # Since p.data is fp32 and p.col_attr.sharded_data_tensor is fp16
 
                 # TODO() optimize this line CPU (fp32) -> GPU (fp16)
-                colo_model_data_tensor_move(p, p.col_attr.sharded_data_tensor)
+                p.col_attr.sharded_data_tensor.reset_payload(p.half().to(torch.cuda.current_device()))
```
feifeibear, line 163: why not use colo_model_data_tensor_move_inline API.

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -206,7 +207,7 @@ def _post_backward_operations(self) -> None:
             else:
                 grad_payload = cast_tensor_to_fp32(p.col_attr.fp16_grad)
             if p.col_attr.offload_grad:
-                grad_payload = colo_model_tensor_clone(grad_payload, torch.device('cpu'))
+                grad_payload.data = grad_payload.data.cpu()
```
feifeibear, line 210: colo_model_data_tensor_move_inline?

colossalai/zero/sharded_optim/sharded_optim_v2.py
ver217: DONE

colossalai/zero/sharded_model/sharded_model_v2.py
ver217: DONE

colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -161,7 +161,8 @@ def step(self, *args, **kwargs):
                 # Since p.data is fp32 and p.col_attr.sharded_data_tensor is fp16
 
                 # TODO() optimize this line CPU (fp32) -> GPU (fp16)
-                colo_model_data_tensor_move(p, p.col_attr.sharded_data_tensor)
+                p.col_attr.sharded_data_tensor.reset_payload(
+                    colo_model_tensor_clone(p.half(), torch.cuda.current_device()))
```
feifeibear, line 165: why not use colo_model_data_tensor_move?
```
colo_model_data_tensor_move(p.half().cuda(current_device(), p.col_attr.sharded_data_tensor)
```

ver217: 1. When reuse fp16 shard and offload grad, `sharded_data_tensor` is on CPU. `colo_model_data_tensor_move` calls `tensor.copy_()` which won't change the device of `sharded_data_tensor` to CUDA.
2. According to the profile result of @1SAA , `tensor.copy_()` is slower than `tensor.half().to(torch.cuda.current_device())`. Besides, the latter one is easy to profile using our `PCIEProfiler`.

feifeibear: OK

feifeibear (approved):

PR #540, opened 2022-03-28
[zero] polish ZeroInitContext

ver217:
1. Remove `convert_to_fp16` argument. We don't support fp32 training, parameters must be converted to fp16.
2. Remove `shard_grad` argument. We only support ZeRO-2 and ZeRO-3. Gradients are always sharded.
3. Set the default dtype of `model_numel_tensor` to `torch.long`, in case of overflow.
4. Polish corresponding unit tests.

### DIFF
--- a/colossalai/zero/init_ctx/init_context.py
+++ b/colossalai/zero/init_ctx/init_context.py
@@ -4,12 +4,12 @@ from typing import Optional
 import torch
 from colossalai.context.parallel_mode import ParallelMode
 from colossalai.core import global_context as gpc
+from colossalai.logging import disable_existing_loggers, get_dist_logger
 from colossalai.utils.memory_utils.memory_monitor import colo_cuda_memory_used
 from colossalai.zero.shard_utils import BaseShardStrategy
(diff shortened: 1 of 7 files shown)


colossalai/zero/init_ctx/init_context.py
```diff
@@ -4,12 +4,12 @@
 import torch
 from colossalai.context.parallel_mode import ParallelMode
 from colossalai.core import global_context as gpc
+from colossalai.logging import disable_existing_loggers, get_dist_logger
```
FrankLeeeee, line 7: `disable_existing_loggers` is not used in this file and can be removed.

FrankLeeeee:
LGTM.

colossalai/zero/init_ctx/init_context.py
ver217: DONE

feifeibear (commented):
Did you update examples correspondingly?

ver217:
> Did you update examples correspondingly?

I will create a PR ASAP.

PR #542, opened 2022-03-28
[zero] get memory usage of sharded optim v2.

feifeibear:


colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -119,6 +126,39 @@ def __init__(self,
...
+            t_cuda_use, t_cpu_use = col_tensor_mem_usage(t)
+            cuda_use += t_cuda_use
+            cpu_use += t_cpu_use
+
+        for _, p_fp32 in self.master_params.items():
+            update_mem_use(p_fp32)
+        for group in self.optim.param_groups:
+            for p in group['params']:
+                state = self.optim.state[p]
+                if 'exp_avg' in state:
```
ver217, line 155: Maybe we can traverse all tensor attribute of `self.state[p]` instead of fixed attributes.

colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -142,8 +182,13 @@ def step(self, *args, **kwargs):
                 # Now p.data is sharded
                 # So optimizer states are sharded naturally
 
+        self._logger.info(f"Before step ShardedOptimizerV2 consumes {self.get_memory_usage()[0]/1e6} MB CUDA Memory!",
```
ver217, line 185: Can we use `logger.debug` in this class?

colossalai/zero/sharded_optim/sharded_optim_v2.py
feifeibear: You are right.

colossalai/zero/sharded_optim/sharded_optim_v2.py
feifeibear: Sure

ver217 (approved):

PR #546, opened 2022-03-29
[zero] add zero config to neutralize zero context init

1SAA:


colossalai/zero/init_ctx/init_context.py
```diff
@@ -82,6 +83,31 @@ def _post_context_exec(self):
...
+        self.rm_torch_payload_on_the_fly: bool = rm_torch_payload_on_the_fly
+
+
+ZERO_CONTEXT = ZeroContextConfig()
+
+
+@contextlib.contextmanager
+def modify_zero_context_config(**kwargs):
+    global ZERO_CONTEXT
+    ZERO_CONTEXT = ZeroContextConfig(ZERO_CONTEXT, **kwargs)
```
feifeibear, line 106: https://www.codefactor.io/repository/github/hpcaitech/colossalai/pull/546
can you avoid to use global variable?

colossalai/zero/init_ctx/init_context.py
```diff
@@ -82,6 +83,31 @@ def _post_context_exec(self):
         pass
 
 
+class ZeroContextConfig(object):
+
+    def __init__(self,
+                 previous_config=None,
+                 convert_fp16: bool = False,
```
feifeibear, line 90: is convert_fp16 useful?
@ver217 has removed it from ZeroInitCtx

1SAA: We should set convert_fp16. Otherwise, the weight of gate will be converted to fp16.
@ver217 assumes all parameters should become fp16. Here, we need to have fp32 parameter.

feifeibear: Add comments for this class?
some args make developer confusing...

ver217: Can we just handle gate by ourself? E.g. we can overwrite the `half()` method. Or we can use a flag to indicate some parameters must be fp32.

ver217: I don't think letting users to control the precision is a good decision.

ver217 (approved):

colossalai/zero/init_ctx/init_context.py

PR #548, opened 2022-03-29
[zero] dump memory stats for sharded model

feifeibear:


colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -190,13 +193,15 @@ def step(self, *args, **kwargs):
                 # Now p.data is sharded
                 # So optimizer states are sharded naturally
 
-        self._logger.debug(f"Before step ShardedOptimizerV2 consumes {self.get_memory_usage()[0]/1e6} MB CUDA Memory!",
-                           ranks=[0])
+        self._logger.info(
```
ver217, line 196: Why use `info()`?

Gy-Lu (approved):

colossalai/zero/sharded_optim/sharded_optim_v2.py
feifeibear: I will turn it back to debug!

raejaf (approved):

PR #551, opened 2022-03-30
[zero] label state for param fp16 and grad

feifeibear:
The state of grads are not set correctly.

### DIFF
--- a/colossalai/zero/sharded_param/tensorful_state.py
+++ b/colossalai/zero/sharded_param/tensorful_state.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from logging import NullHandler
+from typing import Optional
 import torch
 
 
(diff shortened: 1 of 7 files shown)


colossalai/zero/sharded_param/tensorful_state.py
```diff
@@ -8,22 +8,22 @@ class TensorState(Enum):
...
-
+    """A Structure stores a Torch Tensor and labeled states. 
+    Inspired from the paper:
     PatrickStar: Parallel Training of Pre-trained Models via Chunk-based Memory Management
 
     https://arxiv.org/abs/2108.05818
     """
 
-    def __init__(self, tensor: torch.Tensor, state: TensorState = TensorState.HOLD) -> None:
+    def __init__(self, tensor: torch.Tensor, state: Optional[TensorState] = TensorState.HOLD) -> None:
```
ver217, line 22: When is the `state` `None`?

ver217: But `state: Optional[TensorState]` means `state` can be `TensorState` or `None`. When can it be `None`?

feifeibear: OK, I misuse the Optional. remove it

ver217 (approved):

Issue #562, opened 2022-03-31
[RFC] heterogenous training to extreme model scale


feifeibear:
## Background of ZeRO implementation (3.30)
Currently, if the param fp16 + grad fp16/fp32 can not host in CUDA. The system will fail.
If the OS can not be host in CPU. The ZeRO offload will fail.

But, we can break the limitations. On 8x80 GB A100 and 500 GB CPU. We can offload some OS to GPU.
On 1x80GB A100 and 500 GB CPU, we can offload some param/grad to CPU.

The tensor data structure is different for different types of model data. The param fp16 is wrapped as ShardedTensor, which is a class derived from StatefulTensor. Grad fp16/fp32 are wrap as StatefulTensor. While, M, V are still PyTorch tensors.

Model data: param fp16, grad fp16/fp32, OS(param fp32, M, V).
Non-model data: mainly activations.

## Runtime Memory tracer
1. Monitor cuda/cpu memory usage of model data and non-model data during training
2. Using Sampling. Here we define sampling moment as the moment trigger sampling.
  1. Sampling is triggered at the moments of {pre, after} x {fwd, bwd} of an operator via zero hook, collecting two kinds of memory statistics
  2. System mem allocation: The maximum memory usage of the system between the current sampling moment and the previous sampling moment.
  3. Model data mem allocation: When Sharded Model and Sharded Optim are initialized, they register themselves in the global variable GLOBAL_MODEL_DATA_TRACER. At the sampling moment, we scan all of parameters, model, and optimizer to get the cpu and cuda memory usage of model data.
  4. Non-model data allocation：The diff of aboves. sys mem. alloc. - model data mem. alloc. The is not precious, but enough for use.
 
![20220331-140334](https://user-images.githubusercontent.com/5706969/160989444-1e5d52c9-9965-4212-970a-9eaecb11fe9c.jpeg)

## Stateful Tensor Manager
1. Label state for all tensors during training.
  1. Its memory layout is managed by a manager.
2. A tensor is labeled as COMPUTE at the exact moment before computing, and labeled as HOLD at the exact moment after computing.
3. Before executing a operator (FWD/BWD/ADAM)，make COMPUTE tensor resident on the computing device，if the device memory is full，evict some HOLD tensors out.
  1. How can I know the device memory is full？Take computing device = GPU as an example.
    1. cuda_demand = size of tensors as COMPUTE but not on GPU now.
[...]

PR #571, opened 2022-03-31
[zero] trace states of fp16/32 grad and fp32 param

ver217:
Make fp16/32 grad and fp32 param `StatefulTensor` and trace their states.


feifeibear (approved):

feifeibear:
related #562

PR #579, opened 2022-03-31
fix parallel_input flag for Linear1D_Col gather_output

Wesley-Jzy:


colossalai/nn/layer/parallel_1d/layers.py
```diff
@@ -302,7 +302,10 @@ def __init__(self,
         with seed(ParallelMode.TENSOR):
             self.reset_parameters(weight_initializer, bias_initializer)
         self._set_tensor_parallel_attributes()
-        set_parallel_input(True)
+        if self.gather_output:
```
kurisusnowdeng, line 305: How about `set_parallel_input(not self.gather_output)`

kurisusnowdeng (changes requested):

colossalai/nn/layer/parallel_1d/layers.py
Wesley-Jzy: LGTM.
Also, what about
```
is_parallel_output = not self.gather_output
set_parallel_input(is_parallel_output)
```? Make it easier to understand and scale in the future as well as to keep simplicity.

Wesley-Jzy: @kurisusnowdeng Done.

feifeibear:
#600

Issue #256, opened 2022-02-24
Any checkpoint saving/loading tutorial provided?


kurisusnowdeng:
### 📚 The doc issue

So far is there any tutorial that shows how to save/load a model in checkpoint?

FrankLeeeee:
Good point, was considering adding this tutorial as someone mentioned in the discussion post as well. I will write a tutorial to cover this next week.

FrankLeeeee:
I have assigned this issue to myself, will close this issue next week upon completion.

feifeibear:
We are planning to provide this feature this week (28th May).

binmakeswell:
https://github.com/hpcaitech/ColossalAI/tree/main/applications/Chat#how-to-saveload-checkpoint
This issue was closed due to inactivity. Thanks.

PR #590, opened 2022-03-31
[model checkpoint] CPU communication ops

kurisusnowdeng:
- Updated communication ops for cpu tensors
- Added scatter_object_list from torch

### DIFF
--- a/colossalai/communication/collective.py
+++ b/colossalai/communication/collective.py
@@ -8,10 +8,13 @@ from torch import Tensor
 
 from colossalai.context import ParallelMode
 from colossalai.core import global_context as gpc
-from colossalai.utils import get_current_device
 
 
-def all_gather(tensor: Tensor, dim: int, parallel_mode: ParallelMode, async_op: bool = False) -> Tensor:
+def all_gather(tensor: Tensor,
+               dim: int,
+               parallel_mode: ParallelMode,
+               on_cpu: bool = False,
+               async_op: bool = False) -> Tensor:
     r"""Gathers all tensors from the parallel group and concatenates them in a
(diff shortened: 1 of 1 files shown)


colossalai/communication/collective.py
```diff
@@ -8,10 +8,13 @@
...
 from colossalai.context import ParallelMode
 from colossalai.core import global_context as gpc
-from colossalai.utils import get_current_device
 
 
-def all_gather(tensor: Tensor, dim: int, parallel_mode: ParallelMode, async_op: bool = False) -> Tensor:
+def all_gather(tensor: Tensor,
+               dim: int,
+               parallel_mode: ParallelMode,
+               on_cpu: bool = False,
```
feifeibear, line 16: on_cpu may have a conflict with tensor.device.
we have to ensure tensor.device is CPU?

colossalai/communication/collective.py
```diff
@@ -54,6 +59,7 @@ def reduce_scatter(tensor: Tensor,
                    dim: int,
                    parallel_mode: ParallelMode,
                    op: ReduceOp = ReduceOp.SUM,
+                   on_cpu: bool = False,
```
feifeibear, line 62: on_cpu may have a conflict with tensor.device.
we have to ensure tensor.device is CPU?

feifeibear:
#256

colossalai/communication/collective.py
kurisusnowdeng: Let me fix. I think maybe we do not need this option, but just check if the tensor is on cpu.

feifeibear (approved):

PR #592, opened 2022-03-31
[model checkpoint] checkpoint utils

kurisusnowdeng:
- Updated checkpoint save/load utils

### DIFF
--- a/colossalai/utils/checkpointing.py
+++ b/colossalai/utils/checkpointing.py
@@ -1,212 +1,253 @@
-import os
-import os.path as osp
-import re
-from typing import Tuple
-from pathlib import Path
+from collections import OrderedDict
+from itertools import chain
 
 import torch
-
-from colossalai.context import Config
+import torch.distributed as dist
+from colossalai.communication.collective import scatter_object_list
 from colossalai.context.parallel_mode import ParallelMode
 from colossalai.core import global_context as gpc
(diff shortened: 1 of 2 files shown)


colossalai/utils/checkpointing.py
```diff
@@ -1,212 +1,253 @@
...
-    tp_local_rank = 0
-    if gpc.is_initialized(ParallelMode.TENSOR):
-        tp_local_rank = gpc.get_local_rank(ParallelMode.TENSOR)
 
-    # pipeline parallel
-    pp_local_rank = 0
-    if gpc.is_initialized(ParallelMode.PIPELINE):
-        pp_local_rank = gpc.get_local_rank(ParallelMode.PIPELINE)
+def broadcast_state_dict(state_dict, parallel_mode):
+    state_dict = [state_dict.copy() if isinstance(state_dict, dict) else state_dict]
```
feifeibear, line 17: what type could state_dict be?
isinstance(state_dict, dict) may not be enough.
If it is a List or Set you also need to assign a copy?

feifeibear:
#256

colossalai/utils/checkpointing.py
kurisusnowdeng: @feifeibear Only the source rank has the dict (an `OrderedDict` loaded from file) to broadcast. Otherwise, we don't care what type it is, since it is going to be replaced by broadcast anyway.

feifeibear (approved):

PR #593, opened 2022-03-31
[model checkpoint][hotfix] unified layers for save&load

kurisusnowdeng:
Reworked unified layers for ease of save/load states
- Added a base unified module class
- Updated initialization of unified layers


feifeibear:
image
#256

kurisusnowdeng:
@feifeibear Merge #594 first, so this check may become correct.

feifeibear (approved):

PR #594, opened 2022-03-31
Feature/checkpoint 1D

kurisusnowdeng:
Updated saving/loading for 1d layers

### DIFF
--- a/colossalai/nn/layer/parallel_1d/layers.py
+++ b/colossalai/nn/layer/parallel_1d/layers.py
@@ -2,6 +2,7 @@
 # -*- encoding: utf-8 -*-
 
 import math
+from collections import OrderedDict
 from typing import Callable, Tuple
 
 import torch
@@ -10,20 +11,25 @@ from colossalai.communication import broadcast
 from colossalai.context import ParallelMode, seed
 from colossalai.core import global_context as gpc
 from colossalai.global_variables import tensor_parallel_env as env
+from colossalai.kernel import LayerNorm
 from colossalai.nn import init as init
 from colossalai.registry import LAYERS
+from colossalai.utils.checkpointing import (broadcast_state_dict, gather_tensor_parallel_state_dict,
+                                            partition_tensor_parallel_state_dict)
 from colossalai.utils.cuda import get_current_device
 from torch import Tensor
 from torch.nn.parameter import Parameter
+from ..vanilla import VanillaPatchEmbedding
 
 from ..base_layer import ParallelLayer
+from ..colossalai_layer._utils import ColossalaiModule
 from ..utils import divide, set_tensor_parallel_attribute_by_partition
 from ._utils import (gather_forward_split_backward, get_parallel_input, reduce_grad, reduce_input, set_parallel_input,
                      split_forward_gather_backward)
 
 
 @LAYERS.register_module
-class Linear1D(torch.nn.Module):
+class Linear1D(ColossalaiModule):
     r"""Linear layer for 1D parallelism.
 
     Args:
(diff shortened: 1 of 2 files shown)


colossalai/nn/layer/parallel_1d/layers.py
```diff
@@ -52,37 +58,69 @@ def __init__(self,
...
 
-    @property
-    def weight(self):
-        return self.layer.weight
 
-    @property
-    def bias(self):
-        return self.layer.bias
+@LAYERS.register_module
+class LayerNorm1D(ColossalaiModule):
```
feifeibear, line 84: The weight and bias are duplicated on every process in TP process group?

colossalai/nn/layer/parallel_1d/layers.py
```diff
@@ -615,3 +893,66 @@ def forward(self, input_: Tensor) -> Tensor:
...
+            for key in param_keys:
+                param = state_dict.pop(key, None)
+                if param is not None:
+                    local_state[key] = param
+
+        local_state = broadcast_state_dict(local_state, ParallelMode.PARALLEL_1D)
+        super()._load_from_state_dict(local_state, prefix, *args)
+
+    def _save_to_state_dict(self, destination, prefix, keep_vars):
+        if gpc.get_local_rank(ParallelMode.TENSOR) == 0:
```
feifeibear, line 957: What if we are using hybrid parallelism.
For example DP+TP. The each TP group will save a ckp?

feifeibear:
Did you provide unit tests for the functionality?

colossalai/nn/layer/parallel_1d/layers.py
kurisusnowdeng: @feifeibear They are supposed to be. 1D does not use tensor parallel layernorm since activations are not parallelized.

colossalai/nn/layer/parallel_1d/layers.py
```diff
@@ -404,6 +586,52 @@ def _set_tensor_parallel_attributes(self):
...
+    def _save_to_state_dict(self, destination, prefix, keep_vars):
+        weight_key = prefix + 'weight'
+        bias_key = prefix + 'bias'
+        local_state = OrderedDict({weight_key: self.weight})
+        if self.bias is not None:
+            local_state[bias_key] = self.bias
+        local_state = gather_tensor_parallel_state_dict(local_state,
+                                                        ParallelMode.PARALLEL_1D,
+                                                        dims={
+                                                            weight_key: -1,
```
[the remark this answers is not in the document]
kurisusnowdeng: Correct. `dims` tells what dimension to partition, and `partition_states` tells whether to partition

colossalai/nn/layer/parallel_1d/layers.py
kurisusnowdeng: @feifeibear Each tp rank 0 will have a state_dict. Then pp ranks gather them to pp rank 0. Then dp ranks gather them to dp rank 0. Finally, we will have a complete state_dict on global rank 0.

colossalai/nn/layer/parallel_1d/layers.py
feifeibear: That is good.

colossalai/nn/layer/parallel_1d/layers.py
feifeibear: OK, sounds good. A cascade way

feifeibear (approved):

feifeibear:
#256

PR #595, opened 2022-03-31
Feature/checkpoint 2D

kurisusnowdeng:
Updated saving/loading for 2d layers

### DIFF
--- a/colossalai/nn/layer/parallel_2d/layers.py
+++ b/colossalai/nn/layer/parallel_2d/layers.py
@@ -1,4 +1,5 @@
 import math
+from collections import OrderedDict
 from typing import Callable
 
 import torch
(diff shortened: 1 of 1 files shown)


colossalai/nn/layer/parallel_2d/layers.py
```diff
@@ -174,10 +345,10 @@ def forward(self, x: Tensor) -> Tensor:
 
         output = layernorm_2d(x, E_x, Var_x, self.normalized_shape, ParallelMode.PARALLEL_2D_ROW,
                               ParallelMode.PARALLEL_2D_COL)
-        bias = add_bias_2d(None, self.beta, self.partitioned_partition, self.row_rank, self.col_rank,
+        bias = add_bias_2d(None, self.bias, self.partitioned_partition, self.row_rank, self.col_rank,
```
feifeibear, line 348: why change beta. gemma -> bias and weight?
The pytorch version improved?

kurisusnowdeng: @feifeibear Yes. In pytorch, they are named as `weight` and `bias`.

feifeibear: I suggest you check the version and be compatible with the older PyTorch.

kurisusnowdeng: @feifeibear Since pytorch 0.4.0, the most early version `LayerNorm` is provided, their names are `weight` and `bias`, and never `gamma` and `beta`.

feifeibear (approved):

PR #598, opened 2022-03-31
[model checkpoint] checkpoint hook

kurisusnowdeng:
Updated checkpoint hook

### DIFF
--- a/colossalai/trainer/hooks/_checkpoint_hook.py
+++ b/colossalai/trainer/hooks/_checkpoint_hook.py
@@ -1,14 +1,11 @@
 #!/usr/bin/env python
 # -*- encoding: utf-8 -*-
 
-import os.path as osp
 from colossalai.logging import get_dist_logger
 
 from colossalai.registry import HOOKS
 from colossalai.trainer.hooks import BaseHook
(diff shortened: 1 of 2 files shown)


colossalai/trainer/hooks/_checkpoint_hook.py
```diff
@@ -17,9 +14,8 @@ class SaveCheckpointHook(BaseHook):
     """Saves the model by interval in training process.
 
     Args:
-       interval (int, optional): Saving interval, defaults to 1.
-       checkpoint_dir (str, optional): Directory of saving checkpoint, defaults to None.
-       suffix (str, optional): Saving suffix of the file, defaults to ''.
+       interval (int, optional): Number of epochs between saving the checkpoint, defaults to 1.
```
feifeibear, line 17: why in saving in the interval of epoch?
What about using iterations.

kurisusnowdeng: We are absolutely able to do so. Just add an `after_train_step` function for this hook.

feifeibear (approved):

PR #601, opened 2022-04-01
[zero] adapt zero for unsharded parameters (Optimizer part)

1SAA:


colossalai/zero/init_ctx/init_context.py
```diff
@@ -88,6 +88,7 @@ class ZeroContextConfig(object):
     """The configuration used to control zero context initialization.
 
     Args:
+        target_device (torch.device): The device where param data after exiting the context.
```
feifeibear, line 91: The device where param data are after exiting the context.

colossalai/zero/init_ctx/init_context.py
```diff
@@ -236,7 +247,8 @@ def hijack_context_config(self, **kwargs):
 
 
 def no_shard_zero_context(is_replicated: bool = True):
```
feifeibear, line 249: def no_shard_zero_context(is_replicated: bool = True) -> ZeroContextConfig
add return type hinting.

feifeibear: get_no_shard_zero_context is more appropriate

colossalai/zero/init_ctx/init_context.py

colossalai/zero/init_ctx/init_context.py

colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -257,6 +249,22 @@ def zero_grad(self, *args, **kwargs):
     def sync_grad(self):
         pass
 
+    def _register_master_weight(self):
```
feifeibear, line 252: If it is an independent function, you'd better Deal with master_params has already exist?

Maybe delete attr and set it again?
delattr, setattr

colossalai/zero/init_ctx/init_context.py
1SAA: The return type of it is Context Manager. 
The utility is "with no_shard_zero_context():"
I think it's appropriate.

colossalai/zero/sharded_optim/sharded_optim_v2.py
1SAA: This function is used to avoid CodeFactor's issue.
It should not be used in other places.

colossalai/zero/init_ctx/init_context.py
feifeibear: OK

colossalai/zero/sharded_optim/sharded_optim_v2.py
feifeibear: Fine

feifeibear (approved):

PR #632, opened 2022-04-01
Feature/Loki-Promtail monitoring configurations

SMesForoush:


feifeibear:
What kind of logs are you supposed to analyze with loki-promtail. 
During the training of DNN, we know just output some debug info.
The metrics like loss and accuracy are logged to file and display with tensorboard.

SMesForoush:
> 

Its main purpose is administrative stuff. Helping to gather all debug info in one place to help check clusters help. I thought clustering of the logs based on their labels and also the ability to filter them could be useful for finding problems. 
It can also be completed with a Prometheus which monitors process CPU usage etc.

feifeibear:
Sorry, the PR currently is not required by the core functionality of ColossalAI.
If we need a monitor, I will reopen it.

PR #633, opened 2022-04-01
fixed bugs in CPU adam

1SAA:
* add cpu adam counter for all cpu adam

* fixed updating error in adam kernel

### DIFF
--- a/colossalai/kernel/cuda_native/csrc/cpu_adam.cpp
+++ b/colossalai/kernel/cuda_native/csrc/cpu_adam.cpp
@@ -493,7 +493,7 @@ int adam_step(int optimizer_id,
                 grads_ptr,
                 exp_avg_ptr,
                 exp_avg_sq_ptr,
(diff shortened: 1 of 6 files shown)


feifeibear (approved):

colossalai/kernel/cuda_native/csrc/cpu_adam.cpp
```diff
@@ -493,7 +493,7 @@ int adam_step(int optimizer_id,
                 grads_ptr,
                 exp_avg_ptr,
                 exp_avg_sq_ptr,
-                params_c.size(0),
+                params_c.numel(),
```
ver217, line 496: @Gy-Lu How did your code pass unit tests if this is a bug?

Gy-Lu: oh...
we don't use suah a huge data size in unit tests to make this parameter in use

PR #660, opened 2022-04-02
[activation checkpoint] removed offload test in ckpt unit test

Gy-Lu:
Offload should be tested by model, not by a forward function.

This test file is for ckpt, not for offload.

I have tested offload with gpt2-xl, and it worked well.


feifeibear (approved):

feifeibear (approved):

FrankLeeeee:
Can you fix the bug in the unit testing?

PR #664, opened 2022-04-03
fix bugs for unsharded parameters when restore data

1SAA:


colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -268,8 +268,14 @@ def _zero_grad(self, recover_data: bool = False):
                 reuse_fp16_shard = p.colo_attr.saved_grad.data_ptr() == p.colo_attr.sharded_data_tensor.data_ptr()
                 p.colo_attr.saved_grad.set_null()
                 if recover_data and reuse_fp16_shard:
+                    # We should write like this to trigger ForceFP32Paramter's half method
+                    p.data = self.master_params[p].payload
```
feifeibear, line 272: @ver217 see if this valid?

feifeibear:
Can you add some explanation about your modification?

colossalai/zero/sharded_optim/sharded_optim_v2.py
feifeibear: The payload of master param may be located on CPU.
This will damage the efficiency.

1SAA: Don't understand why it damages efficiency.
This equals to what @ver217 wrote.

1SAA: You can see [sharded_optim/sharded_optim_v2.py] line 350.
We use this way to transmit param weight from cpu to gpu.

feifeibear: p.data has already been on GPU.
But master param may be on CPU.
You PR always use the master param. Compared to the original one, it inncreases CPU-GPU transmission.

feifeibear: I see, if you updates only work for recover_data, it dose not matter.

feifeibear (approved):

colossalai/zero/sharded_optim/sharded_optim_v2.py
1SAA: In this situation, we can not reuse p.data, since we use it to restore p.grad.
This only works for reuse_fp16 = True.

1SAA: p.data may on GPU. But we can't find it.

PR #679, opened 2022-04-06
[pipeline] refactor pipeline

YuliangLiu0306:
infer pipeline schedule parameters, such as tensor_shape or scatter_gather_tensors, from user config file


colossalai/engine/schedule/_pipeline_schedule.py
```diff
@@ -16,6 +16,29 @@
...
+            seq_size = gpc.get_world_size(ParallelMode.SEQUENCE)
+        else:
+            seq_size = 1
+
+        tensor_shape = (gpc.config.SEQ_LENGTH // seq_size,
+                        gpc.config.GLOBAL_BATCH_SIZE // dp_size // gpc.config.NUM_MICRO_BATCHES,
+                        gpc.config.HIDDEN_SIZE)
+        return tensor_shape
+    else:
+        return None
```
feifeibear, line 41: raise error?

YuliangLiu0306: Some cases in our examples do not set either TENSOR_SHAPE or SEQ_LENGTH in config file. Therefore, I think it should be okay to return None.

ver217:
Is this a good practice? In this way, users rely on config file more heavier, which is diffenret from Pytorch's pattern.

YuliangLiu0306:
> Is this a good practice? In this way, users rely on config file heavier, which is diffenret from Pytorch's pattern.

Actually, this PR do not add any addtitional requirement to the config file except set NUM_MICRO_BATCHES in pipeline parallel mode. We could infer those argument for schedule before, but we don't. This PR just fill it.

kurisusnowdeng:
I think a good way to define tensor shape for pipeline is to set static batch size, seq length and hidden size for the model. The schedule reads the sizes from the model. If they are None, we transfer the shape in the first step.

YuliangLiu0306:
> I think a good way to define tensor shape for pipeline is to set static batch size, seq length and hidden size for the model. The schedule reads the sizes from the model. If they are None, we transfer the shape in the first step.

I agree with you, but I don't find a standard to define the config file, so the rule in this PR is induced from the examples config. If they are None, we will get the shape in the warmup stage.

kurisusnowdeng:
@FrankLeeeee In what case we need to transfer multiple tensors across the pipeline?

FrankLeeeee:
> @FrankLeeeee In what case we need to transfer multiple tensors across the pipeline?

In cases such as dual path network, one example is SlowFast, one of the most popular video networks.

kurisusnowdeng:
@YuliangLiu0306 `scatter_gather_tensors` is only used for 1D parallelism. We suggest to enable it when 1D is used. So, how about hide this option from users?

YuliangLiu0306:
> @YuliangLiu0306 `scatter_gather_tensors` is only used for 1D parallelism. We suggest to enable it when 1D is used. So, how about hide this option from users?

Yes. As we put schedule into engine, not only that option, the whole initialization process of schedule has been hidden. From users' view, they don't have to care about how to initialize schedule.

FrankLeeeee (approved):

PR #681, opened 2022-04-06
[refactor] make os stateful tensor

ver217:
related to #562 
Stateful OS tensors making them be managed dynamically in CPU-GPU space.

### DIFF
--- a/colossalai/zero/sharded_optim/sharded_optim_v2.py
+++ b/colossalai/zero/sharded_optim/sharded_optim_v2.py
@@ -139,6 +139,8 @@ class ShardedOptimizerV2(ColossalaiOptimizer):
         if self._use_memory_tracer:
             GLOBAL_MODEL_DATA_TRACER.register_optimizer(self)
 
+        self._state: Dict[Parameter, Dict[str, StatefulTensor]] = {}
+
     def get_memory_usage(self) -> Tuple[int, int]:
         """ Get the memory usage of the optimizer. Including master_params (param fp32),
         momentum (``self.state[p]['exp_avg']``) variance (``self.state[p]['exp_avg_sq']``)
@@ -183,6 +185,7 @@ class ShardedOptimizerV2(ColossalaiOptimizer):
             return
 
         self._point_param_fp16_to_master_param()
+        self._set_optim_states_before_step()
 
         self._logger.debug(
             f"Before step ShardedOptimizerV2 consumes {self.get_memory_usage()[0] / 1e6} MB CUDA Memory, {self.get_memory_usage()[1] / 1e6} MB CUDA Memory!",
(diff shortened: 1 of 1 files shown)


colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -139,6 +139,8 @@ def __init__(self,
         if self._use_memory_tracer:
             GLOBAL_MODEL_DATA_TRACER.register_optimizer(self)
 
+        self._state: Dict[Parameter, Dict[str, StatefulTensor]] = {}
```
feifeibear, line 142: will step() uses self._state?

colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -358,3 +362,28 @@ def _copy_master_param_to_param_fp16(self):
...
+    def _set_optim_states_before_step(self):
+        for group in self.optim.param_groups:
+            for p in group['params']:
+                # We don't need to set
+                # self.optim.state[p][k] = self._state[p][k].payload
+                # since payload is moved over devices inline
+                # self.optim.state[p][k]'s tensor id won't change
+                # the only change is its data pointer
+                for optim_state in self._state[p].values():
+                    optim_state.trans_state(TensorState.COMPUTE)
```
feifeibear, line 375: In this way, you set all OS tensor to compute?
You should set a tensor to COMPUTE the exact time it is about to be computed.

colossalai/zero/sharded_optim/sharded_optim_v2.py
ver217: It will use its payload.

colossalai/zero/sharded_optim/sharded_optim_v2.py
ver217: We are unable to do this unless we use a custom optimizer.

ver217: That is to say, users cannot use torch.optim.SGD, torch.optim.Adam anymore.

feifeibear: I am afraid so. If you use the stateful tensor, then you have to use our stateful optimizer.

ver217: I discuss with @1SAA . We both think that OS is not suitable for dynamic memory layout. Static layout like now maybe more  efficient.

ver217: Move OS dynamically meybe not reduce computation time, however, it increases data movement time. Besides, it won't save momery usage.

ver217: And we have to make sure OS, fp32 weight and gradient are on the same device for each parameter. That is to say, our eviction algorithm has to identify those stateful tensors are "linked" to which parameter. Although they are all "tensors", but they are not "independent" tensors.

ver217: I think dynamic memory layout is suitable for fp16 params. OS, fp32 weight and gradient can be moved once and stay on the device.

feifeibear: OK, we make param fp16 stateful tensor first.

PR #686, opened 2022-04-07
[zero] fix init bugs in zero context

1SAA:
- adapt model weight initialization for methods in Pytorch nn.init

### DIFF
--- a/colossalai/zero/init_ctx/init_context.py
+++ b/colossalai/zero/init_ctx/init_context.py
@@ -3,6 +3,8 @@ import functools
 from typing import Optional
 
 import torch
+import torch.nn as nn
+import torch.distributed as dist
 from colossalai.context.parallel_mode import ParallelMode
 from colossalai.core import global_context as gpc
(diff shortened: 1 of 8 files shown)


colossalai/zero/init_ctx/init_context.py
```diff
@@ -167,25 +157,76 @@ def is_replicated(self):
...
+
+        return fan_in, fan_out
 
     def _pre_context_exec(self):
         """ 
         The Callback function when entering the context
         """
         self.logger = get_dist_logger("ZeroInitContext")
 
+        # substitute fan-in and fan-out calculation
```
feifeibear, line 196: Can you exhaust all possible functions？
_calculate_fan_in_and_fan_out and calc_fanin_fanout just two of them.

1SAA: It works for all init functions in nn.init and colossalai.nn.init.

feifeibear (approved):

Issue #698, opened 2022-04-08
[BUG]: Classifier Failed for 2D and 2.5D in PyTorch 1.9


FrankLeeeee:
### 🐛 Describe the bug

When running unit testing with PyTorch 1.9, the following error occurs when testing the classifier. 

<img width="1077" alt="Screenshot 2022-04-08 at 1 57 08 PM" src="https://user-images.githubusercontent.com/31818963/162373610-05197a31-0d03-41d0-91cf-c3cb11870e72.png">

PR #700, opened 2022-04-08
[compatibility] fixed tensor parallel compatibility with torch 1.9

FrankLeeeee:
This fixed the issue #698 . As the CI uses PyTorch 1.10, I have pasted the screenshot for the unit testing `test_layers`.

Screenshot 2022-04-08 at 2 00 49 PM

### DIFF
--- a/colossalai/nn/layer/parallel_2d/_operation.py
+++ b/colossalai/nn/layer/parallel_2d/_operation.py
@@ -58,6 +58,7 @@ def matmul_2d(
 
 
 class _Classifier2D(torch.autograd.Function):
+
     @staticmethod
     @custom_fwd(cast_inputs=torch.float16)
     def forward(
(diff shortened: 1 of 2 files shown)


colossalai/nn/layer/parallel_2d/_operation.py
```diff
@@ -76,7 +77,7 @@ def forward(
         pipeline_parallel_size: int,
         tensor_parallel_size: int,
     ) -> Tensor:
-
+        A = A.clone().detach()
```
ver217, line 80: Do you try `A = A.detach()`?

FrankLeeeee: I didn't try it. However, I think `A.detach()` will lead to incorrect computation graph because `A` is the input tensor of the linear layer. If `A` is detached from the graph, the graph will be disconnected.

ver217 (approved):

colossalai/nn/layer/parallel_2d/_operation.py
FrankLeeeee: According to the PyTorch documentation, `detach` will create a tensor which shares the storage with the original tensor. If there happens to be some in-place operations in the forward and backward functions, there will be incorrect arithmetic results.

Screenshot 2022-04-11 at 1 40 14 PM

ver217: OK

PR #708, opened 2022-04-09
[zero] improve adaptability for not-shard parameters

1SAA:
* adapt post grad hooks for not-shard parameters
* adapt optimizer for not-shard parameters
* offload gradients for not-replicated parameters


colossalai/zero/init_ctx/init_context.py
```diff
@@ -242,6 +242,10 @@ def half_fn(t: torch.Tensor):
             # mark whether the param is replicated
             param.is_replicated = self.is_replicated
```
ver217, line 243: As other attributes are set in `param.colo_attr`. Should we also set `is_replicated` and `keep_not_shard` in `param.colo_attr`?

1SAA: Since `colo_attr` is an instance of `ShardedParamV2`, which is assumed to be sharded naturally.
It is weird to add an attribute `keep_not_shard` for it.

ver217 (approved):

PR #710, opened 2022-04-11
[hotfix] fix stateful tensor manager's cuda model data size

ver217:


colossalai/zero/shard_utils/stateful_tensor_mgr.py
```diff
@@ -48,14 +49,13 @@ def adjust_layout(self) -> None:
         # find stateful tensor in state COMPUTE
         move_to_cuda_tensor_list = []
         cuda_demand = 0
-        used_cuda_model_data = 0
+        used_cuda_model_data = GLOBAL_MODEL_DATA_TRACER.cuda_usage
```
FrankLeeeee, line 52: The previous version is an int, but here it is a function, can you add some description about the impact of this change?

FrankLeeeee (approved):

feifeibear (approved):

PR #711, opened 2022-04-11
[hotfix]fixed bugs of assigning grad states to non leaf nodes

Gy-Lu:


kurisusnowdeng (approved):

colossalai/utils/activation_checkpoint.py
```diff
@@ -13,7 +13,8 @@ def copy_to_device(obj, device):
         # Notice:
         # When in no_grad context, requires_gard is False after movement
         ret = obj.to(device)
-        ret.requires_grad = obj.requires_grad
+        if ret.is_leaf or ret.retains_grad:
```
kurisusnowdeng, line 16: I think activations to offload should be all non-leaf?

Gy-Lu: yes... it is weird but we have met the problem.

Gy-Lu: yes. I found that to() or cpu() or gpu() would change a leaf tensor to a non leaf one.

Issue #688, opened 2022-04-07
[BUG]: Poor Compatibility with PyTorch 1.9


FrankLeeeee:
### 🐛 Describe the bug

I have conducted unit testing with PyTorch 1.9 and CUDA 11.1.1. Many unit tests failed and require our attention. This is important because that we can't assume that users are using latest PyTorch. 

I will try to fix these problems this week. Future efforts will be made for lower versions of PyTorch as well. The aim is to support as low as PyTorch 1.6 (at the very least the major features).

<img width="510" alt="Screenshot 2022-04-07 at 4 20 12 PM" src="https://user-images.githubusercontent.com/31818963/162153954-2b56b309-728a-414e-9d6b-92954ac01fe7.png">

### Environment

Python: 3.8
PyTorch: 1.9.1
CUDA: 11.1.1
Docker: 20.10

FrankLeeeee:
All bugs are fixed in #700 , #719 and #721
<img width="490" alt="Screenshot 2022-04-12 at 12 47 48 PM" src="https://user-images.githubusercontent.com/31818963/162883005-2f8ad90e-4469-4ab4-b40a-17c9dd671c62.png">

PR #719, opened 2022-04-11
[util] fixed activation checkpointing on torch 1.9

### LINKED ISSUE #718: [BUG]: Activation Checkpointing Failed Test with PyTorch 1.9
### 🐛 Describe the bug

When running unit testing with PyTorch 1.9, the following exception occurs for `tests/test_utils/test_activation_checkpointing.py`. 

## Screenshot

Lark20220411-162602

## Reasons

This bug occurs because of the invalid use of `ctx.save_for_backward`. In torch autograd function, we need to provide either inputs or intermediate values to `ctx.save_for_backward`. However, when `activation_offload=True`, the current implementation will copy the tensors to cpu, creating new tensors which are neither inputs or intermediate activations (i.e. does not have gradient accumulator objects associated with them). 

```python
for i, arg in enumerate(args):
      if torch.is_tensor(arg):
          if activation_offload:
              tensor_inputs.append(copy_to_device(arg, 'cpu'))
          else:
              tensor_inputs.append(arg)
          ctx.tensor_indices.append(i)
          ctx.inputs.append(None)
      else:
          ctx.inputs.append(arg)

  ctx.save_for_backward(*tensor_inputs)
```

This will trigger the [assertion statement](https://github.com/pytorch/pytorch/blob/ed7ece389daa980a03a2f09848a6439a90af9782/torch/csrc/autograd/saved_variable.cpp#L196) as shown below.

Screenshot 2022-04-11 at 4 27 50 PM

## Solution

Use `cxt.tensor_inputs = tensor_inputs` instead of `ctx.save_for_backward(*tensor_inputs)` if `activation_offload=True`.

### Environment

CUDA: 11.1
PyTorch: 1.9.1

FrankLeeeee:
This fixed #718  as part of the effort to tackle #688 

The screenshot of successful test with `torch 1.9` is shown below.

Screenshot 2022-04-11 at 4 34 19 PM

### DIFF
--- a/colossalai/utils/activation_checkpoint.py
+++ b/colossalai/utils/activation_checkpoint.py
@@ -68,7 +68,10 @@ class CheckpointFunction(torch.autograd.Function):
             else:
                 ctx.inputs.append(arg)
 
-        ctx.save_for_backward(*tensor_inputs)
+        if activation_offload:
+            ctx.tensor_inputs = tensor_inputs
+        else:
+            ctx.save_for_backward(*tensor_inputs)
         return outputs
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


ver217 (approved):

Gy-Lu:
so is there a bug in torch 1.9's `save_for_backward`?

FrankLeeeee:
> so is there a bug in torch 1.9's `save_for_backward`?

It is not a bug, it is a requirement of pytorch 1.9

Issue #722, opened 2022-04-11
[FEATURE]: Detect the Number of Processes on the Current Node


FrankLeeeee:
### Describe the feature

In memory optimization, it is important to know which processes are running on the same node so as to calculate the available CPU memory per process. We can support this feature and store the result in parallel context. Some solutions are discussed.

1. use `torch.cuda.device_count`

This built-in torch function can directly give us how many GPUs are available on the device. However, we must assume that all GPUs are used if using this function and this may not be true. Moreover, schedulers like SLURM may automatically bind one process to one GPU so this function will always return 1.

2. use environment variable

SLURM and OpenMPI can provide environment variables for local rank, we can all-reduce the local rank to get the max local rank value. For example, if max local rank on each node is 7, this means that there are 8 processes on this node. However, this may not be true for unusual process launching by hostfile. For example, we can launch 2 processes on 1 node and 4 processes on 1 node. This would give result of 4 for the 2-process node as well.

3. manual counting by hostname

We can count the number of processes by hostnames. As hostname is unique, we can gather the hostname from all process and do a simple counting. The result will be the count of the hostname detected by the current process.

In my opinion, method 3 is the most universal one for implementation.

PR #723, opened 2022-04-11
[util] support detection of number of processes on current node

FrankLeeeee:
Implemented method 3 mentioned in #722 .

### DIFF
--- a/colossalai/context/parallel_context.py
+++ b/colossalai/context/parallel_context.py
@@ -2,6 +2,8 @@
 # -*- encoding: utf-8 -*-
 
 import random
+import socket
+from collections import Counter
 from typing import Union
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


colossalai/context/parallel_context.py
```diff
@@ -81,6 +84,13 @@ def load_config(self, config: Union[dict, str]):
         else:
             raise TypeError("Invalid type for config, only dictionary or string is supported")
 
+    def detect_num_processes_on_current_node(self):
```
feifeibear, line 87: Can we just run this once and cache the results. 
Frequent all gather is not good for performance

FrankLeeeee: Yes, it only runs once in colossalai.launch

PR #726, opened 2022-04-12
[utils] correct cpu memory used and capacity in the context of multi-process

feifeibear:
make `colo_device_memory_capacity` return the correct value.
rename zero test directory.


tests/test_utils/test_memory.py
```diff
@@ -0,0 +1,32 @@
...
+from colossalai.utils.memory import colo_set_process_memory_fraction, colo_device_memory_capacity
+from colossalai.utils import free_port
+
+from functools import partial
+import torch.multiprocessing as mp
+
+
+def _run_colo_set_process_memory_fraction_and_colo_device_memory_capacity():
+    frac1 = colo_device_memory_capacity(get_current_device())
+    colo_set_process_memory_fraction(0.5)
```
FrankLeeeee, line 14: I checked that this API is only supported from PyTorch 1.8 onwards, will this have an impact on the users who use zero with lower torch versions?

FrankLeeeee (approved):

tests/test_utils/test_memory.py
feifeibear: DONE, the function will return if torch < 1.8 and print warning

FrankLeeeee: Great.

PR #750, opened 2022-04-13
[hotfix] layernorm

kurisusnowdeng:
Allow layernorm without bias

### DIFF
--- a/colossalai/nn/layer/vanilla/layers.py
+++ b/colossalai/nn/layer/vanilla/layers.py
@@ -254,3 +254,37 @@ class VanillaClassifier(nn.Module):
 
     def forward(self, input_: Tensor) -> Tensor:
         return F.linear(input_, self.weight, self.bias)
+
+
+@LAYERS.register_module
+class VanillaLayerNorm(nn.Module):
+    r"""
+    Layer Normalization for colossalai
+
+    Args:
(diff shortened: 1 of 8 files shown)


colossalai/nn/layer/vanilla/layers.py
```diff
@@ -254,3 +254,37 @@ def reset_parameters(self, weight_initializer, bias_initializer):
 
     def forward(self, input_: Tensor) -> Tensor:
         return F.linear(input_, self.weight, self.bias)
+
+
+@LAYERS.register_module
+class VanillaLayerNorm(nn.Module):
```
ver217, line 260: Is it different from `torch.nn.LayerNorm`?

kurisusnowdeng: @ver217 A little. We allow it to ignore bias.

ver217 (approved):

feifeibear:
Do we have a unit test for layernorm?

feifeibear:
Do we need a unitest for non-bias cases? But, if you have checked the updating, that's fine.

kurisusnowdeng:
> Do we need a unitest for non-bias cases? But, if you have checked the updating, that's fine.

 @feifeibear Sure. We can refine the coverage of tests, as well as for the linear layers.

PR #752, opened 2022-04-14
[zero] use factory pattern for tensor_placement_policy

feifeibear:
using global variable is not recommanded.


1SAA:
Why thre is an empty file called colossalai/zero/utils/policy_factory.py?

feifeibear:
I remove the useless code

ver217 (approved):

PR #753, opened 2022-04-14
[hotfix] fix auto tensor placement policy

ver217:


colossalai/zero/utils/tensor_placement_policy.py
```diff
@@ -45,7 +45,8 @@ class AutoTensorPlacementPolicy(TensorPlacementPolicy):
 
     def __init__(self, mem_stats_collector: Optional[MemStatsCollector] = None) -> None:
         super().__init__(None, mem_stats_collector=mem_stats_collector)
-        self._warmup_non_model_data_ratio: float = 0.2
+        # model data will use 1-self._warmup_non_model_data_ratio CUDA memory in warmup phase
+        self._warmup_non_model_data_ratio: float = 0.8
```
FrankLeeeee, line 49: should this parameter be configurable?

ver217: @feifeibear What do you think of this?

ver217: I think this parameter can be configurable, however, it's difficult for the most users to tune.

FrankLeeeee: We can give a default value for this argument, but I am not sure if it is good to fix it inside the class init like what it is now.

feifeibear: I think make it configurable is necessary.
We can have a Genini config later.

feifeibear (approved):

PR #762, opened 2022-04-14
[hotfix] fix test_stateful_tensor_mgr

### LINKED ISSUE #761: [BUG]: fix the stateful tensor mgr unitest bug.
### 🐛 Describe the bug

now we skip the testing in test_stateful_tensor_mgr.py in CI.
However, we should at least make it run smoothly in our development environment.

### Environment

_No response_

ver217:
Fixed #761

### DIFF
--- a/tests/test_zero/test_stateful_tensor_mgr.py
+++ b/tests/test_zero/test_stateful_tensor_mgr.py
@@ -5,7 +5,7 @@ import torch.multiprocessing as mp
 from colossalai.utils.cuda import get_current_device
 from colossalai.utils.memory_tracer import MemStatsCollector
 from colossalai.utils.memory_tracer.model_data_memtracer import GLOBAL_MODEL_DATA_TRACER
-from colossalai.utils.memory import colo_device_memory_capacity, colo_set_process_memory_fraction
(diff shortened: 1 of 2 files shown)


tests/test_zero/test_stateful_tensor_mgr.py
```diff
@@ -27,12 +27,16 @@ def __init__(self) -> None:
...
+def limit_cuda_memory(memory_in_g: float):
+    cuda_capacity = torch.cuda.get_device_properties(get_current_device()).total_memory
+    fraction = (memory_in_g * 1024**3) / cuda_capacity
     colo_set_process_memory_fraction(fraction)
+
+
+def run_stm():
+    # warmup phase use 20% CUDA memory to store params
+    # only 2 params can be on CUDA
+    limit_cuda_memory(5.1)
```
FrankLeeeee, line 39: why 5.1?

ver217: Because we set CUDA memory limit by a fraction which is a floating point, and this leads to precision error. So we set it to a larger value.

FrankLeeeee (approved):

feifeibear (approved):

Issue #669, opened 2022-04-05
[BUG]: Check in 2D dim % world_size should be put after checking the shape of dim.


Wesley-Jzy:
### 🐛 Describe the bug

[This line](https://github.com/hpcaitech/ColossalAI/blob/ce8a3eae5be301d1e3f536ede32a2b0c65f7c51f/colossalai/nn/layer/parallel_2d/_operation.py#L735) should be put after checking dim.

PR #772, opened 2022-04-15
change the check assert in split batch 2d

Wesley-Jzy:
Dim shape check should be put before we really prepare to chunk it.
https://github.com/hpcaitech/ColossalAI/issues/669

### DIFF
--- a/colossalai/nn/layer/parallel_2d/_operation.py
+++ b/colossalai/nn/layer/parallel_2d/_operation.py
@@ -738,12 +738,13 @@ def split_batch_2d(input_: Tensor, dim: int = 0) -> Tensor:
         :class:`torch.tensor`: The tensor has been split.
     """
     dim_size = input_.size(dim)
+    if dim_size <= 1:
+        return input_
+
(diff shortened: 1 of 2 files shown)


ver217 (approved):

colossalai/nn/layer/parallel_2d/_operation.py
```diff
@@ -738,12 +738,13 @@ def split_batch_2d(input_: Tensor, dim: int = 0) -> Tensor:
         :class:`torch.tensor`: The tensor has been split.
     """
     dim_size = input_.size(dim)
+    if dim_size <= 1:
```
kurisusnowdeng, line 741: Not really correct.

kurisusnowdeng (changes requested):

colossalai/nn/layer/parallel_2d/_operation.py
Wesley-Jzy: I see. You are right.

kurisusnowdeng (approved):

PR #777, opened 2022-04-15
modefied the pp build for ckpt adaptation

Gy-Lu:
Our former pipeline build is not suitable for ckpt.
I have fixed it.

### DIFF
--- a/colossalai/builder/pipeline.py
+++ b/colossalai/builder/pipeline.py
@@ -237,10 +237,9 @@ def build_pipeline_model_from_cfg(config,
     return nn.ModuleList(models) if len(models) > 1 else models[0]
 
 
-def build_pipeline_model(layers: nn.Sequential, num_chunks: int = 1, verbose: bool = False):
+def build_pipeline_model(layers: nn.Sequential, num_chunks: int = 1, verbose: bool = True):
     """An intializer to split the model into different stages for pipeline parallelism.
     Note that `layer` must be `torch.nn.Sequential`.
-
     Args:
         layers (`torch.nn.Sequential`): Layers of model
         num_chunks: The number of chunks you want to have on the current stage. This value should be 1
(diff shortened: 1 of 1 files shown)


colossalai/builder/pipeline.py
```diff
@@ -250,9 +250,15 @@ def build_pipeline_model(layers: nn.Sequential, num_chunks: int = 1, verbose: bo
     pipeline_parallel_size = gpc.get_world_size(ParallelMode.PIPELINE)
     pipeline_rank = gpc.get_local_rank(ParallelMode.PIPELINE)
     partitions = partition_uniform(len(layers), pipeline_parallel_size, num_chunks)
-    module_list = []
+    layer_marks = []
     for start, end in partitions[pipeline_rank]:
-        module_list.append(nn.Sequential(*layers[start:end]))
+        layer_marks += [i for i in range(start, end)]
+    module_list = []
+    for i, _ in enumerate(layers):
```
ver217, line 257: You can just use `for i in range(layers)`.

colossalai/builder/pipeline.py
```diff
@@ -262,4 +268,4 @@ def build_pipeline_model(layers: nn.Sequential, num_chunks: int = 1, verbose: bo
                 log_str += f'===== chunk={chunk}, layer=[{start}-{end}] =====\n'
                 log_str += '\n'.join([str(layer) for layer in layers[start:end]]) + '\n'
             logger.info(log_str, ranks=[0])
-    return nn.ModuleList(module_list) if len(module_list) > 1 else module_list[0]
+    return nn.Sequential(*tuple(module_list)) if len(module_list) > 1 else module_list[0]
```
ver217, line 271: `ModuleList` is used for interleaved pipeline. Each module in this `ModuleList` is a `Sequential` module. Why do you use `Sequential` here? Actually I think they shouldn't be sequential.

colossalai/builder/pipeline.py
Gy-Lu: I am not sure. I need the no. of the layer. `range(layers)` seems to provide the layer name?

colossalai/builder/pipeline.py
Gy-Lu: I forget to consider the interleaved pipeline. Seems `ModuleList` is fine, I'll test it.

Gy-Lu: I reconsider it. I think the interleaved pipeline is supported, for we use `torch.nn.Identity`. `Sequntial` is ok as well.

Gy-Lu: I am totally wrong. It needs to be redesigned.

PR #781, opened 2022-04-17
[hotfix] fix bugs in zero

1SAA:
* solve memory leak when using reuse_fp16_shard
* fix type hint in gradient scaler
* add verbose to control the logger in ShardedOptimizerV2

### DIFF
--- a/colossalai/zero/sharded_optim/sharded_optim_v2.py
+++ b/colossalai/zero/sharded_optim/sharded_optim_v2.py
@@ -83,11 +83,12 @@ class ShardedOptimizerV2(ColossalaiOptimizer):
                  min_scale: float = 1,
                  growth_factor: float = 2,
                  backoff_factor: float = 0.5,
-                 growth_interval: float = 1000,
-                 hysteresis: float = 2,
-                 max_scale: int = 2**32,
+                 growth_interval: int = 1000,
+                 hysteresis: int = 2,
+                 max_scale: float = 2**32,
                  dp_process_group: Optional[ProcessGroup] = None,
-                 mp_process_group: Optional[ProcessGroup] = None) -> None:
+                 mp_process_group: Optional[ProcessGroup] = None,
+                 verbose: bool = False) -> None:
         assert isinstance(sharded_model, ShardedModelV2), 'model must be wrapped with ShardedModel'
 
         super().__init__(optimizer)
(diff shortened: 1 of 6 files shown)


colossalai/zero/sharded_param/sharded_param.py
```diff
@@ -5,18 +5,13 @@
...
-FAKE_EMPTY_TENSOR = torch.BoolTensor([], device='cpu')
-
 EMPTY_TENSOR_DICT = {}
 
 
 def get_empty_tensor(device: torch.device, dtype: torch.dtype):
     key = (device, dtype)
     if key not in EMPTY_TENSOR_DICT:
-        EMPTY_TENSOR_DICT[key] = FAKE_EMPTY_TENSOR.to(device, dtype)
+        EMPTY_TENSOR_DICT[key] = torch.empty(0, dtype=dtype, device=device)
```
feifeibear, line 14: The original code. different keys share the same memory space if the device and type are the same.
Now, for each key has an individual memory space.
@ver217

colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -193,15 +197,20 @@ def step(self, *args, **kwargs):
 
         self._point_param_fp16_to_master_param()
 
-        self._logger.debug(
-            f"Before step ShardedOptimizerV2 consumes {self.get_memory_usage()[0] / 1e6} MB CUDA Memory, {self.get_memory_usage()[1] / 1e6} MB CUDA Memory!",
-            ranks=[0])
+        if self._verbose:
+            gpu_mem, cpu_mem = self.get_memory_usage()
+            self._logger.debug(
+                f"Before step ShardedOptimizerV2 consumes {gpu_mem / 1e6} MB CUDA Memory, {cpu_mem / 1e6} MB CUDA Memory!",
```
feifeibear, line 203: You imply that we shall not call function to build the str for logging?

1SAA: Yeah, since the functions are always called whether the string would be printed.

colossalai/zero/sharded_param/sharded_param.py
1SAA: This is used to acoid wasted memory in cpu.

feifeibear: OK

feifeibear (approved):

PR #789, opened 2022-04-18
[cli] added micro benchmarking for tp

YuliangLiu0306:


FrankLeeeee:
Hi @YuliangLiu0306 , CodeFactor has found some minor issues in your code, can you fix it?

YuliangLiu0306:
> Hi @YuliangLiu0306 , CodeFactor has found some minor issues in your code, can you fix it?



> Hi @YuliangLiu0306 , CodeFactor has found some minor issues in your code, can you fix it?

Done.

setup.py
```diff
@@ -213,6 +213,10 @@ def cuda_ext_helper(name, sources, extra_cuda_flags, extra_cxx_flags=[]):
     ext_modules=ext_modules,
     cmdclass={'build_ext': BuildExtension} if ext_modules else {},
     install_requires=fetch_requirements('requirements/requirements.txt'),
+    entry_points='''
+        [console_scripts]
+        colossal=colossalai.scripts.cli:cli
```
FrankLeeeee, line 218: Is there any purpose of creating two-level directories? I think the norm is to have a folder directly under the main repo, e.g. `/colossalai/cli`

colossalai/benchmark/run.py
```diff
@@ -0,0 +1,86 @@
+import torch
+import inspect
+import os
+import subprocess
+import sys
```
FrankLeeeee, line 5: Should we put the benchmark files into the cli folder as well? I think this module is not gonna be used by the user directly, so better to hide it from the user.

FrankLeeeee (approved):

PR #798, opened 2022-04-19
[refactor] remove circle imports

feifeibear:


colossalai/engine/schedule/_pipeline_schedule.py
```diff
@@ -115,7 +116,7 @@ def load_micro_batch(self):
     def pre_processing(self, engine):
         # TODO: remove this after testing new zero with pipeline parallelism
         model = engine.model
-        if isinstance(model, (NaiveAMPModel)) or hasattr(model, 'colo_attr'):
+        if isinstance(model, (NaiveAMPModel)) or isinstance(model, ShardedModelV2):
```
feifeibear, line 119: revert it back to yuliang's logic. I mistakingly modify the code in the previous PR.

colossalai/engine/gradient_accumulation/__init__.py
```diff
@@ -6,6 +6,11 @@
 from torch.optim.lr_scheduler import _LRScheduler
 from ._gradient_accumulation import GradAccumDataloader, GradAccumOptimizer, GradAccumLrSchedulerByStep, GradAccumGradientHandler
```
FrankLeeeee, line 7: There is a test for gradient accumulation in `test_utils`. I guess this movement will lead to test error.

feifeibear: OK, I will fix it

PR #804, opened 2022-04-19
[refactor] moving grad acc logic to engine

feifeibear:
Avoid circle import.


feifeibear:
The failed CI case is due to address in use. I tested it locally.
image

FrankLeeeee:
> 

Ok, this occurs as the `rerun_if_address_in_use` decorator is not added.

colossalai/engine/schedule/_pipeline_schedule.py
```diff
@@ -386,8 +387,7 @@ def __init__(self,
         self.num_model_chunks = num_model_chunks
 
     def pre_processing(self, engine):
-        # FIXME(jiaruifang) we shall not use ShardedModelV2 in pipeline mode, due to circular dependency.
-        if hasattr(engine.model, 'colo_attr'):
+        if isinstance(engine.model[0], ShardedModelV2):
```
ver217, line 390: Maybe we can use `isinstance(engine.model[0], (ShardedModelV2, NaiveAMPModel))`

feifeibear: This is a bug. It should be if isinstance(engine.model, ShardedModelV2):
I fixed it

ver217 (approved):

PR #808, opened 2022-04-19
[DO NOT MERGE] [zero] init fp16 params directly in ZeroInitContext

ver217:


colossalai/zero/init_ctx/init_context.py
```diff
@@ -24,13 +24,15 @@ def _substitute_init_recursively(cls, func):
...
-        pass
+        self._old_default_dtype = None
 
     def __enter__(self):
         r"""
         Enter the context scope.
         """
 
+        self._old_default_dtype = torch.get_default_dtype()
+        torch.set_default_dtype(torch.half)
```
FrankLeeeee, line 35: As the impact of fp16 initialization on model convergence is not known at this moment, I would recommend that we set it as an argument for `__init__` and the default value is None. We only set default type when we pass in a dtype  to `__init__`.

ver217: DONE

feifeibear (approved):

PR #809, opened 2022-04-19
[gemini] APIs to set cpu memory capacity

feifeibear:


ver217 (approved):

colossalai/utils/memory.py
```diff
@@ -152,3 +152,27 @@ def colo_set_process_memory_fraction(ratio: float) -> None:
...
     _GLOBAL_CUDA_MEM_FRACTION = ratio
     torch.cuda.set_per_process_memory_fraction(_GLOBAL_CUDA_MEM_FRACTION, get_current_device())
+
+
+def colo_set_cpu_memory_capacity(size: int) -> None:
+    global _GLOBAL_CPU_MEM_CAPACITY
+    mem_info = _get_cpu_memory_info()
+    total_size = mem_info.total
+    if size <= total_size:
+        _GLOBAL_CPU_MEM_CAPACITY = size
```
ver217, line 162: Do we realy limit cpu memory here? If activation offload is enabled and cpu memory exceeds this value, we won't get OOM, right?

feifeibear: It is only meaningful for auto stateful tensor manager. Otherwise, it is useless.

ver217: ok

feifeibear:
image

PR #811, opened 2022-04-20
[log] local throughput metrics

feifeibear:


feifeibear:
image

I will fix the global var code factor issue in the next PR

feifeibear:
use_local = False
image

use_local = True
image

FrankLeeeee (approved):

PR #813, opened 2022-04-20
[gemini] collect cpu-gpu moving volume in each iteration

feifeibear:


ver217 (approved):

feifeibear:
We can display cpu-gpu data moving volume like this

image

PR #823, opened 2022-04-21
[tensor] lazy init

feifeibear:


colossalai/tensor/colo_tensor.py
```diff
@@ -1,16 +1,44 @@
...
+        torch_tensor=None,
+    ):
+        self._size = size
+        self._dtype = dtype
+        self._requires_grad = requires_grad
+        self._pin_memory = pin_memory
+        self._torch_tensor = torch_tensor
+
+    @staticmethod
+    def _init_from_torch_tensor(tensor: torch.Tensor):
```
Wesley-Jzy, line 32: Why let this staticmethod start with '_'? It seems not a internal function.

feifeibear: OK

Wesley-Jzy (approved):

PR #828, opened 2022-04-22
[tensor] ZeRO use ColoTensor as the base class.

feifeibear:
ColoTensor is more capable. It can pass directly to PyTorch functions, if it has been wrapped by `colo_op_impl`.
Next, I will merge `StatefulTensor` to ColoTensor.
Also let TP use the ColoTensor.


Wesley-Jzy:
The interface name `torch_tensor()` is not intuitive compared to PyTorch's `local_tensor` or our previous `payload`. A torch tensor which is gathered for global op is also a `torch_tensor`. We may change it after implementing sharding.

Wesley-Jzy (approved):

PR #833, opened 2022-04-22
[dependency] removed torchvision

FrankLeeeee:
`torchvision` and some libraries are not necessary dependencies for `colossalai`. Therefore, they should be removed and only appear in tests. The registry is no longer needed except for a few, this will be cleaned in a separate PR.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,9 +1,7 @@
 torch>=1.8
-torchvision>=0.9
 numpy
 tqdm
 psutil
-tensorboard
 packaging
 pre-commit
 rich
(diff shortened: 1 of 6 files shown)


feifeibear (approved):

requirements/requirements.txt
```diff
@@ -1,9 +1,7 @@
 torch>=1.8
-torchvision>=0.9
 numpy
 tqdm
 psutil
-tensorboard
```
ver217, line 6: Our hook may use tensorboard. Do we have tips if users use the hook using tensorboar without installing it?

FrankLeeeee: We only import tensorboard within the `__init__` of the hook. This is implemented so that tensorboard is not a must. The change in the `requirement.txt` will not break colossalai.

ver217 (approved):

FrankLeeeee:
The test is passed.

Screenshot 2022-04-22 at 3 23 13 PM

PR #837, opened 2022-04-22
[utils] refactor profiler

ver217:
1. refactor profiler. Our profiler is a subclass of pytorch's profiler now, which can use other utils of pytorch's profiler (such as tensorboard_trace_handler and schedule).
2. add stateful tensor memory profiler extention.
3. move old profiler to colossalai.utils.profiler.legacy.

Example:
```python
from colossalai.utils.profiler import profile
from torch.profiler import ProfilerActivity, tensorboard_trace_handler, schedule

engine, *_ = colossalai.initialize(model, optimizer, criterion)

with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
             with_stack=False,
             record_shapes=True,
             profile_memory=True,
             on_trace_ready=tensorboard_trace_handler('events/st'),
             schedule=schedule(skip_first=0, wait=0, warmup=0, active=NUM_EPOCHS * NUM_STEPS),
             profile_stateful_tensor_memory=True,
             engine=engine) as prof:
    for epoch in range(NUM_EPOCHS):
        for i in range(NUM_STEPS):
            run_step()
            prof.step()
```


feifeibear (approved):

ver217:
All tests passed.

image

PR #852, opened 2022-04-24
[Tensor] add layer norm Op

feifeibear:


colossalai/tensor/_ops/layernorm.py
```diff
@@ -0,0 +1,37 @@
...
+        eps = kwargs['eps']
+
+    if isinstance(input, ColoTensor):
+        input = input.torch_tensor()
+    if isinstance(weight, ColoTensor):
+        weight = weight.torch_tensor()
+    if isinstance(bias, ColoTensor):
+        bias = bias.torch_tensor()
+
+    return ColoTensor.init_from_torch_tensor(torch.nn.functional.layer_norm(input, normalized_shape, weight, bias, eps))
```
Wesley-Jzy, line 37: It's 1D-specific layernorm.

feifeibear: You mean TP 1D row splitting?

feifeibear: You suggest I assign a Spec to the weight/bias tensor of LN? and add a if for weight.spec to make sure the logic is only work for 1D row?

Wesley-Jzy (approved):

PR #854, opened 2022-04-24
[Tensor ] Add 1Drow weight reshard by spec

Wesley-Jzy:
1. Add 1Drow weight reshard by spec
2. Update unit test


colossalai/tensor/_ops/linear.py
```diff
@@ -39,12 +40,15 @@ def colo_linear(types, args, kwargs, pg):
             # Input:S[1]
             input_per_partition = split_forward_gather_backward(input_tensor, ParallelMode.PARALLEL_1D, dim=-1)
             # Output:P
-            partial_output = torch.nn.functional.linear(input_per_partition, weight.torch_tensor())
+            device = get_current_device() # TODO where to put to(deivce)?
```
feifeibear, line 43: We can not support cpu linear if so.

feifeibear (approved):

PR #866, opened 2022-04-25
[Tensor] colo init context add device attr.

feifeibear:


colossalai/tensor/colo_tensor.py
```diff
@@ -57,6 +58,10 @@ def size(self):
     def shape(self):
         return torch.Size(self._size)
 
+    @property
+    def device(self):
+        return self._torch_tensor.device
```
Wesley-Jzy, line 63: Maybe we should return self._deivce and ensure the correctness of self._device.

Wesley-Jzy (approved):

PR #867, opened 2022-04-25
[gemini] add stateful tensor container

1SAA:


colossalai/gemini/stateful_tensor_container.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(self, compute_step_dict: Dict[StatefulTensor, List[int]], total_step: int):
+        super().__init__(compute_step_dict, total_step)
+        self.container = None
+
+    def empty(self) -> bool:
+        assert self.container is not None
+        return self.container == []
+
+    def create(self, stateful_tensor_list: List[StatefulTensor]) -> None:
```
feifeibear, line 97: Can you try to use a PQ?
from queue import PriorityQueue

ver217 (approved):

colossalai/gemini/stateful_tensor_container.py
1SAA: Heap is faster.
It has an linear time creation, while PQ needs nlogn time.
Furthermore, PQ is the same as heap, since we provide our API like a queue.

PR #869, opened 2022-04-25
[Tensor] Add function to spec and update linear 1Drow and unit tests

Wesley-Jzy:
1. Add function to TensorSpec
2. Update tensot unit tests


colossalai/tensor/spec.py
```diff
@@ -44,4 +38,28 @@ class TensorSpec(Enum):
...
+    def num_policy(self):
+        return len(self._parallel_action_list)
+
+    @property
+    def compute_patterns(self):
+        return [parallel_action.compute_pattern for parallel_action in self._parallel_action_list]
+
+    def sort(self):
+        if len(self._parallel_action_list) > 0:
+            self._parallel_action_list.sort(key=lambda parallel_action : parallel_action.priority, reverse=True)
```
feifeibear, line 59: from small to large? so the lower the priority value the earlier to be executed?

colossalai/tensor/spec.py
```diff
@@ -44,4 +38,28 @@ class TensorSpec(Enum):
...
+    def __init__(self, parallel_action_list: List[ParallelAction] = []):
+        self._parallel_action_list = parallel_action_list
+        self.sort()
+
+    @property
+    def parallel_action_list(self):
+        return self._parallel_action_list
+
+    @property
+    def num_policy(self):
```
feifeibear, line 50: num_action is more properiate!

colossalai/tensor/colo_tensor.py
```diff
@@ -109,27 +108,26 @@ def torch_tensor(self) -> torch.Tensor:
...
-            # leaf variable in the autograd graph.
-            self._torch_tensor = self._torch_tensor.narrow(dim, local_rank * chunk_size, chunk_size).detach(
-            ).contiguous()    # TODO Shall we clone() here since detach() will point to the old tensor?
-            self._torch_tensor.requires_grad = self._requires_grad
-            self._size = self._torch_tensor.size()
-            self._device = device    # TODO A `fake` device now because torch_tensor.device always = cpu
+        if self._shard_spec.num_policy == 1:
+            if ComputePattern.TP1DRow in self._shard_spec.compute_patterns:
+                num_partition = gpc.get_world_size(ParallelMode.TENSOR)
+                local_rank = gpc.get_local_rank(ParallelMode.TENSOR)
```
feifeibear, line 121: fetch the process group from spec

colossalai/tensor/_ops/linear.py
```diff
@@ -7,7 +7,7 @@
 from colossalai.core import global_context as gpc
 from packaging import version
 from colossalai.utils.cuda import get_current_device
-
+from colossalai.tensor.spec import TensorSpec, ComputePattern, ParallelAction
```
feifeibear, line 10: from colossalai.tensor import
Use the public API in shallow directory as much as possible

colossalai/tensor/_ops/linear.py
```diff
@@ -30,32 +30,33 @@ def colo_linear(types, args, kwargs, pg):
...
-
+        elif weight.shard_spec.num_policy == 1:
+            if ComputePattern.TP1DRow in weight.shard_spec.compute_patterns:
+                # Input:S[1] x Weight:S[0] = Output:P
+                # All-Reduce(Output) + bias = res
+                assert divide(input_tensor.shape[-1], gpc.tensor_parallel_size) == weight.size(-1), \
+                'Invalid shapes in 1Drow forward: input={}, weight={}. Expected last dim of input {}.'.format(
+                    input_tensor.shape, weight.size, weight.size[-1] * gpc.tensor_parallel_size)
+                # Input:S[1]
+                input_per_partition = split_forward_gather_backward(input_tensor, ParallelMode.PARALLEL_1D, dim=-1)
```
feifeibear, line 47: fecth the parallel mode from spec please.

colossalai/tensor/_ops/linear.py
```diff
@@ -30,32 +30,33 @@ def colo_linear(types, args, kwargs, pg):
...
+                # Output:P
+                weight_ = weight.torch_tensor()
+                partial_output = torch.nn.functional.linear(input_per_partition, weight_)
+                # Reduce(Output)
+                output = reduce_input(partial_output, ParallelMode.PARALLEL_1D)
+                # Bias
+                if bias is not None:
+                    bias_ = bias
+                    output = output + bias_
+                return output
```
feifeibear, line 57: return a ColoTensor?
do you think it should be a default procedure?

colossalai/tensor/colo_tensor.py
```diff
@@ -28,7 +27,7 @@ def __init__(
             pin_memory=False,
             device=None,
             torch_tensor=torch.empty(0),
-            shard_spec: str = None,
+            shard_spec: TensorSpec = None,
```
feifeibear, line 30: I think the default should be compute pattern as DP?

colossalai/tensor/colo_tensor.py
```diff
@@ -39,7 +38,7 @@ def __init__(
         self._shard_spec = shard_spec
 
     @property
-    def shard_spec(self) -> Optional[str]:
+    def shard_spec(self) -> Optional[TensorSpec]:
```
feifeibear, line 41: If the default is a DP compute pattern. the spec won't be None.

colossalai/tensor/colo_tensor.py
```diff
@@ -109,27 +108,26 @@ def torch_tensor(self) -> torch.Tensor:
...
+        if self._shard_spec.num_policy == 1:
+            if ComputePattern.TP1DRow in self._shard_spec.compute_patterns:
+                num_partition = gpc.get_world_size(ParallelMode.TENSOR)
+                local_rank = gpc.get_local_rank(ParallelMode.TENSOR)
+                dim = -1
+                chunk_size = divide(self._size[dim], num_partition)
+                # Reshape to get shard for this rank and we don't want autograd
+                # recording here for the narrow op and 'local_shard' should be a
+                # leaf variable in the autograd graph.
+                self._torch_tensor = self._torch_tensor.narrow(dim, local_rank * chunk_size, chunk_size).detach(
```
feifeibear, line 127: if weight is of size (m, k)
Did you split it as (m, k/N)?
Why such splitting is called 1D row? it is splitting on the column dim?

colossalai/tensor/spec.py
Wesley-Jzy: Done

colossalai/tensor/spec.py
Wesley-Jzy: Done

colossalai/tensor/colo_tensor.py
Wesley-Jzy: Done

colossalai/tensor/_ops/linear.py
Wesley-Jzy: Done

colossalai/tensor/_ops/linear.py
Wesley-Jzy: Done

colossalai/tensor/colo_tensor.py
Wesley-Jzy: It's because the calculation of torch.nn.functional.linear is Input x transpose(W)

colossalai/tensor/colo_tensor.py
Wesley-Jzy: Done

colossalai/tensor/colo_tensor.py
Wesley-Jzy: Done

colossalai/tensor/_ops/linear.py
Wesley-Jzy: I return a ColoTensor now. But I'm not sure whether we shall do that. We have hijacked all Parameters in context(i.e. All torch functions have redirected into ours.), so the type of Input should take no effects in my opition.

feifeibear: OK

feifeibear (approved):

feifeibear (approved):

PR #870, opened 2022-04-25
[model_zoo] change qkv processing

Gy-Lu:
I am trying to load pre_trained parameters of hugging face gpt2.
And I found our qkv processing is different from [theirs](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L312), maybe leading to bugs of loading ckpt.


Gy-Lu:
I have tested that the shapes of q, k, and v are still the original ones.

feifeibear (approved):

PR #874, opened 2022-04-26
[Tensor] get named parameters for model using ColoTensors

feifeibear:
The `named_parameters` won't work for ColoTensor model


feifeibear:
image

colossalai/tensor/utils.py
```diff
@@ -2,6 +2,57 @@
...
+        for name, val in vars(mod).items():
+            if isinstance(val, ColoTensor) and val not in memo:
+                memo.add(val)
+                name = mod_prefix + ('.' if mod_prefix else '') + name
+                yield name, val
+
+    # find all nn.Parameters
+    for name, val in module.named_parameters():
+        yield name, val
+
```
Wesley-Jzy, line 55: Need we add the same return of named_parameters() after the colotensor parameters? It means this function will return both types, ColoTensor and torch.Parameter.

feifeibear: Yes... we need all parameters for the Optim

Wesley-Jzy: But this function will return some ColoTensor and some Parameter. It may make it difficult to be handled by the downstream work. Can we return all values with ColoTensor or Parameter?

feifeibear: OK, we can let the caller to check the type of the yielded elements.

Wesley-Jzy: OK.

Wesley-Jzy (approved):

PR #875, opened 2022-04-26
[tensor] customized op returns ColoTensor

feifeibear:
We have to make the output of Op a ColoTensor.
Therefore we can assign the spec for them.
For example, the output of 1Drow Linear should be a shard of the global tensor. We can better represent it using ColoTensor than a Torch Tensor.


feifeibear:
image

tests/test_tensor/test_op.py
```diff
@@ -53,11 +53,11 @@ def test_linear():
 
     # torch.nn.functional.linear(torch.randn(1, in_dim), sharded_weight, sharded_bias)
     out = fc(input_tensor)
-    loss = out.sum()
```
Wesley-Jzy, line 56: How will we support torch tensor's functionality? Implement all of them in the ColoTensor?

feifeibear: Maybe, implement all of them in ColoTensor.

colossalai/tensor/_ops/linear.py
```diff
@@ -1,3 +1,4 @@
+from numpy import isin
```
Wesley-Jzy, line 1: Auto added by vscode?

feifeibear: rm it

Wesley-Jzy (approved):

PR #878, opened 2022-04-26
[gemini] accelerate adjust_layout()

ver217:
As our eviction strategies are deterministic. For each phase (warmup and steady phase) and each computation ID, ` move_to_cuda_tensor_list` and `hold_cuda_tensor_list` are the same. We can use a cache to store the results. Also, as OPT algorithm is deterministic for static computation graph, we can also we a cache to store the results.

Cache ` move_to_cuda_tensor_list` and `hold_cuda_tensor_list`:
image

Cache all:
image
image
image
image

I trained GPT2-18B with 90 transformer blocks here. `layout time` is the time to compute ` move_to_cuda_tensor_list` and `hold_cuda_tensor_list`. `evict time` is eviction time besides cpu-gpu data movement time.


ver217:
image

All tests passed.

ver217:
image

All tests passed.

1SAA (approved):

PR #879, opened 2022-04-26
[Tensor] make a simple net works with 1D row TP

feifeibear:
In the next PR, I will check the accuracy of our TP simple net


feifeibear:
image

colossalai/tensor/colo_tensor.py
```diff
@@ -157,5 +157,21 @@ def __torch_function__(cls, func, types, args=(), kwargs=None):
...
+        return ColoTensor.init_from_torch_tensor(self.torch_tensor().view(*args))
+
+    def permute(self, *args) -> "ColoTensor":
+        return ColoTensor.init_from_torch_tensor(self.torch_tensor().permute(*args))
+
+    def transpose(self, *args) -> "ColoTensor":
+        return ColoTensor.init_from_torch_tensor(self.torch_tensor().transpose(*args))
+
+    def contiguous(self):
+        return ColoTensor.init_from_torch_tensor(self.torch_tensor().contiguous())
```
Wesley-Jzy, line 177: What about overload __getattr__ to redirect all unknown Tensor functions to payload's functions?
```
class B:
    def __init__(self):
        self.aa = 1

    def foo(self, a):
        print("B.foo: ", a)

class A:
    def __init__(self):
        self.a = 1
        self.b = 2
        self.c = 3
        self.sub = B()

    def __getattr__(self, name):
        print(name)
        return getattr(self.sub, name)

    def f(self):
        print("f")

    def g(self):
        print("g")

a = A()
a.foo(2)
```

Wesley-Jzy (approved):

PR #890, opened 2022-04-27
[setup] add local version label

ver217:
If building colossalai with cuda extension, the version can be `0.1.3+torch110cu113`. If not, the version is just `0.1.3`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,12 @@ def fetch_readme():
 
 def get_version():
     with open('version.txt') as f:
-        return f.read().strip()
+        version = f.read().strip()
+        if build_cuda_ext:
+            torch_version = ''.join(torch.__version__.split('.')[:2])
+            cuda_version = ''.join(get_cuda_bare_metal_version(CUDA_HOME)[1:])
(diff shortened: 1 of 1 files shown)


FrankLeeeee (approved):

setup.py
```diff
@@ -94,7 +94,12 @@ def fetch_readme():
 
 def get_version():
     with open('version.txt') as f:
-        return f.read().strip()
+        version = f.read().strip()
+        if build_cuda_ext:
+            torch_version = ''.join(torch.__version__.split('.')[:2])
```
feifeibear, line 99: [:2] is hard coding...
from packaging import version
use this lib to detect version number.

feifeibear (approved):

setup.py
ver217: OK, I will test it and create a PR later.

ver217: `packaging` is not in standard library. If users want to install from source, they must install `pacaking` first. I think it's a trouble for users, so I think we should not use `packaging` now.

PR #895, opened 2022-04-28
[schema] added configuration schema for validation and consistency

FrankLeeeee:


colossalai/schema/zero.py
```diff
@@ -0,0 +1,42 @@
...
+    offload_config: dict = dict(device="cpu")
+    gradient_predivide_factor: StrictFloat = 1.0
+    use_memory_tracer: StrictBool = False
+    reuse_fp16_shard: StrictBool = False
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    @validator('offload_config')
+    def check_offload_device(cls, v):
```
feifeibear, line 21: We should leave the check as a member function in the ShardedModel or corresponding class.
Therefore if we change the interface, we do not have to remember to change the logic here.

FrankLeeeee: Got it.

PR #905, opened 2022-04-29
[ci] added wheel build scripts

FrankLeeeee:
This PR checks for all available combination of torch+cuda+python wheels and build ColossalAI individually as wheels for distribution. All torch wheel information is stored and updated in [public_assets](https://github.com/hpcaitech/public_assets/tree/main/colossalai/torch_wheel).

We can set the cuda version and the ColossalAI we want to build, then this PR will automatically build ColossalAI with various torch and python versions.

Screenshot 2022-04-29 at 2 41 43 PM

### DIFF
--- a/.github/workflows/release_bdist.yml
+++ b/.github/workflows/release_bdist.yml
@@ -3,18 +3,21 @@ name: Release bdist wheel
 on:
   workflow_dispatch:
     inputs:
-      version:
+      cuda_version:
         type: choice
-        description: version for testing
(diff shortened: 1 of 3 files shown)


ver217:
Why do we upload `all_dist/` and `torch.whl`?

FrankLeeeee:
> Why do we upload `all_dist/` and `torch.whl`?

deleted, they are left-over files after testing locally.

PR #911, opened 2022-05-05
[Tensor] add a basic bert.

Wesley-Jzy:
1. Add basic huggingface bert. Remain to be replaced with 1D ColoTensor like col_nn.


tests/test_tensor/test_bert.py
```diff
@@ -0,0 +1,95 @@
...
+    device = get_current_device()
+    
+    set_seed(1)
+    with ColoInitContext(device=device):
+        model = model_builder(checkpoint=True)
+    
+    parallel_action_list_row = [
+        ParallelAction(priority=1, compute_pattern=ComputePattern.TP1DRow_Linear, parallel_mode=ParallelMode.PARALLEL_1D)
+    ]
+    spec_row = TensorSpec(parallel_action_list_row)
```
FrankLeeeee, line 37: This is not used?

tests/test_tensor/test_bert.py
```diff
@@ -0,0 +1,95 @@
...
+def set_seed(seed):
+    random.seed(seed)
+    os.environ['PYTHONHASHSEED'] = str(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed(seed)
+    torch.backends.cudnn.deterministic = True
+
+def run_bert_1d():
+    get_components_func = non_distributed_component_funcs.get_callable('bert')
```
feifeibear, line 26: just reuse the test_model.py

tests/test_tensor/test_bert.py
```diff
@@ -0,0 +1,95 @@
...
+        print(loss.torch_tensor())
+
+def run_dist(rank, world_size, port):
+    config = dict(parallel=dict(tensor=dict(mode="1d", size=world_size),))
+    colossalai.launch(config=config, rank=rank, world_size=world_size, host='localhost', port=port, backend='nccl')
+    run_bert_1d()
+
+@pytest.mark.skip
+@pytest.mark.dist
+@parameterize('world_size', [1])
```
FrankLeeeee, line 88: For entry function starting with `test_`, we should use `@pytest.mark.paramterize` so that `pytest` can count the number of test cases when invoked with `pytest`.

tests/test_tensor/test_bert.py
```diff
@@ -0,0 +1,95 @@
...
+        ParallelAction(priority=1, compute_pattern=ComputePattern.TP1DCol_Linear, parallel_mode=ParallelMode.PARALLEL_1D)
+    ]
+    spec_col = TensorSpec(parallel_action_list_col)
+
+    parallel_action_list_embedding_col = [
+        ParallelAction(priority=1, compute_pattern=ComputePattern.TP1DCol_Embedding, parallel_mode=ParallelMode.PARALLEL_1D)
+    ]
+    spec_embedding_col = TensorSpec(parallel_action_list_embedding_col)
+
+    for name, p in named_params_with_colotensor(model):
```
feifeibear, line 49: use model.named_parameters()

tests/test_tensor/test_bert.py
Wesley-Jzy: OK.

Wesley-Jzy:
@feifeibear continues updating bert with ColoTensor. This MR is closed now.

tests/test_tensor/test_bert.py
Wesley-Jzy: Done.

tests/test_tensor/test_bert.py
Wesley-Jzy: Done

tests/test_tensor/test_bert.py
Wesley-Jzy: It seems no use..

feifeibear (approved):

tests/test_tensor/test_model.py
```diff
@@ -249,23 +307,95 @@ def run_1d_row_tp():
...
+
+    for name, p in model.named_parameters():
+        print(f"{name}: {p}")
+
+    for name, p in named_params_with_colotensor(model):
+        if not isinstance(p, ColoTensor):
+            continue
+        #print(name)
+        if 'classifier' in name and ('weight' in name or 'bias' in name):
+            p.set_spec(spec_col)
```
FrankLeeeee, line 342: Just a side note regarding the current design regarding customized operations, will such usage be an over-kill? I have two questions as listed below:
1. does the user need to know the name of the parameters for replacement?
2. this seems like one-size-fits-all and does not cater to the demand of customization? For example, I only want to keep some linear layers parallelized and the others not, is it possible to achieve this?

Wesley-Jzy: 1. You are right. Current usage is unsuitable to be exposed to users. It's just a verification that we have a method to provide parellel without changing models at all. Maybe specs should be inserted by analyzing op graph automatically in the final.
2. Yes, every op is splitted if we handle output carefully(just like insert a reshape op into the ops with different shape).

FrankLeeeee: Ok, understand

PR #914, opened 2022-05-06
[Tensor] init ColoParameter

feifeibear:
As discussed in our private report (redesigning tensor structure for ColossalAI), we have to distinguish the model data and non-model data tensors. They have different implementations.
For example, model data has to be applied with Gemini and arranged in chunks if necessary. 
To be consistent with PyTorch, I initialized a class as ColoParameter. Currently, the only difference is the self._type attribute.
but in the future, we will add more instinctive features in `ColoParameter`.


feifeibear:
image

Wesley-Jzy (approved):
LGTM.

PR #916, opened 2022-05-06
[Tensor] fix test_model

Wesley-Jzy:


tests/components_to_test/bert.py
```diff
@@ -37,9 +37,11 @@ def get_training_components():
     num_head = 4
     sequence_length = 12
     num_layer = 2
+    vocab_size = 30524
 
     def bert_model_builder(checkpoint):
-        config = BertConfig(gradient_checkpointing=checkpoint,
+        config = BertConfig(vocab_size=vocab_size,
+                            gradient_checkpointing=checkpoint,
```
feifeibear, line 44: The latest best model does not use this flag?
you can see the warning

feifeibear (approved):

tests/components_to_test/bert.py
Wesley-Jzy: Yes, line 65 is for this.

PR #917, opened 2022-05-07
[Graph] building computing graph with ColoTensor, Linear only

feifeibear:
Conduction a forward pass under the GraphContext() to build the computing graph.
Add graph-building logic in every customize Ops (now just for linear). Maybe we can reduce the code redundancy by some programing tricks. Any advice?


feifeibear:
Conduction a forward pass under the GraphContext() to build the computing graph.
Add graph-building logic in every customize Ops (now just for linear). Maybe we can reduce the code redundancy by some programing tricks. Any advice?

Wesley-Jzy:
> Conduction a forward pass under the GraphContext() to build the computing graph. Add graph-building logic in every customize Ops (now just for linear). Maybe we can reduce the code redundancy by some programing tricks. Any advice?

Use one function to pack all logics registing an op node, then invoke it before return.
Or move the logics back to the nn.linear handle function.

kurisusnowdeng:
- Do we really need a `GraphNode` class? Like tensor.is_leaf in pytorch, we can simply add each tensor's parents and children as ColoTensor's attributes, which may also help to deal with dynamic graphs.
- Similarly, do we really need `GraphOpNode`? What op information do we need in terms of building parallel models? From my perspective, the op information is mainly useful for autograd.
- Besides, I guess recording tensor instances in a graph may keep all of their data in memory, but in fact we only need their meta information.

feifeibear:
> Few shallow ideas:
> 
> * Do we really need a `GraphNode` class? Like tensor.is_leaf in pytorch, we can simply add each tensor's parents and children as ColoTensor's attributes, which may also help to deal with dynamic graphs.
> * Similarly, do we really need `GraphOpNode`? What op information do we need in terms of building parallel models? From my perspective, the op information is mainly useful for autograd.
> * Besides, I guess recording tensor instances in a graph may keep all of their data in memory, but in fact we only need their meta information.

1. what do you mean tensor's parents and children? I think the tensor.is_leaf is a completely concept different from my graphnode.
2. At least we need the params shapes. what do you mean op information?
3. what do you mean all of their data? They just keep some pointers and consume very little memory.
What do you mean meta information?

kurisusnowdeng:
> > Few shallow ideas:
> > 
> > * Do we really need a `GraphNode` class? Like tensor.is_leaf in pytorch, we can simply add each tensor's parents and children as ColoTensor's attributes, which may also help to deal with dynamic graphs.
> > * Similarly, do we really need `GraphOpNode`? What op information do we need in terms of building parallel models? From my perspective, the op information is mainly useful for autograd.
> > * Besides, I guess recording tensor instances in a graph may keep all of their data in memory, but in fact we only need their meta information.
> 
> 1. what do you mean tensor's parents and children? I think the tensor.is_leaf is a concept completely different from my graphnode.
> 2. At least we need the params shapes. what do you mean op information?
> 3. what do you mean all of their data? They just keep some pointers and consume very little memory.
>    What do you mean meta information?

I am talking about how to record the graph. As far as I have seen here, the necessary information only includes each tensor's parents and children. So, in the most concise format, I think making them be ColoTensor's attributes is enough.

feifeibear:
> > > Few shallow ideas:
> > > 
> > > * Do we really need a `GraphNode` class? Like tensor.is_leaf in pytorch, we can simply add each tensor's parents and children as ColoTensor's attributes, which may also help to deal with dynamic graphs.
> > > * Similarly, do we really need `GraphOpNode`? What op information do we need in terms of building parallel models? From my perspective, the op information is mainly useful for autograd.
> > > * Besides, I guess recording tensor instances in a graph may keep all of their data in memory, but in fact we only need their meta information.
> > 
> > 
> > 
> > 1. what do you mean tensor's parents and children? I think the tensor.is_leaf is a concept completely different from my graphnode.
> > 2. At least we need the params shapes. what do you mean op information?
> > 3. what do you mean all of their data? They just keep some pointers and consume very little memory.
> >    What do you mean meta information?
> 
> I am talking about how to record the graph. As far as I have seen here, the necessary information only includes each tensor's parents and children. So, in the most concise format, I think making them be ColoTensor's attributes is enough.

I see. We will add more information to the graph node later. So using an independent class for graphs seems better.

feifeibear:
> > > Few shallow ideas:
> > > 
> > > * Do we really need a `GraphNode` class? Like tensor.is_leaf in pytorch, we can simply add each tensor's parents and children as ColoTensor's attributes, which may also help to deal with dynamic graphs.
> > > * Similarly, do we really need `GraphOpNode`? What op information do we need in terms of building parallel models? From my perspective, the op information is mainly useful for autograd.
> > > * Besides, I guess recording tensor instances in a graph may keep all of their data in memory, but in fact we only need their meta information.
> > 
> > 
> > 
> > 1. what do you mean tensor's parents and children? I think the tensor.is_leaf is a concept completely different from my graphnode.
> > 2. At least we need the params shapes. what do you mean op information?
> > 3. what do you mean all of their data? They just keep some pointers and consume very little memory.
> >    What do you mean meta information?
> 
> I am talking about how to record the graph. As far as I have seen here, the necessary information only includes each tensor's parents and children. So, in the most concise format, I think making them be ColoTensor's attributes is enough.

I realize that your advice 3 is very **insightful**. Now I keep a graph while the graph node is an attribute of Colotensor. This will break the reference counting and stay more activation tensors in memory.

feifeibear:
After discussing with @kurisusnowdeng  and some coffee, my mind is more clear. 
The graph should be organized with Op as node and Action Tensor as edge. Then it won't add an extra reference to activation tensors.

Wesley-Jzy (approved):

feifeibear:
I updated the PR.
The graph's node is Op, and the edge is the activation tensor.
For a 4 Linear stacked model, the graph is looks like the follows.

image

colossalai/tensor/graph/graph_node.py
```diff
@@ -0,0 +1,92 @@
...
+    def __enter__(self):
+        GraphGlobalEnv().graph_building = True
+
+    def __exit__(self, *exc_info):
+        GraphGlobalEnv().graph_building = False
+        self.graph_nodes = GraphGlobalEnv().graph_node_list
+
+
+class GraphNode(object):
+    myid = 0
```
Wesley-Jzy, line 34: We need a func to reset this global var before running a model every time.

feifeibear: We just run the graph building once under GraphContext.

Wesley-Jzy: Also, what about using param name itself as node identification.

feifeibear: OK, the name can be an attribute.
I have updated the id resetting.

feifeibear: The name can be very long in some cases. Now just use a number.

PR #921, opened 2022-05-09
[Tensor] add from_pretrained support and bert pretrained test

Wesley-Jzy:


colossalai/tensor/colo_tensor.py
```diff
@@ -55,6 +55,15 @@ def shard_pattern(self):
     def data(self):
         return self._torch_tensor.data
 
+    @data.setter
+    def data(self, tensor: Union[torch.Tensor, "ColoTensor"]):
```
feifeibear, line 59: we should consider the lazy inited cases.

That is to say a colossal tensor is built from `init_from_torch_tensor(..., save_payload = False)`.

feifeibear (approved):

PR #934, opened 2022-05-12
[tensor] design DistSpec and DistSpecManager for ColoTensor

ver217:
1. design DistSpec and DistSpecManager for ColoTensor
2. update current ops, besides loss
3. skip `test_model.py`


colossalai/tensor/colo_tensor.py
```diff
@@ -127,85 +132,21 @@ def torch_tensor(self) -> torch.Tensor:
...
-        dim = self._get_gather_dim()
-        out_tensor = torch.cat(tensor_list, dim=dim).contiguous()
-
-        return out_tensor
-
-    def is_gathered(self) -> bool:
-        return self._shard_pattern == ShardPattern.NA
+    def set_spec(self, spec: TensorSpec, shard: bool = True, init_pg: bool = False) -> None:
+        spec = copy(spec)
+        if init_pg:
```
Wesley-Jzy, line 137: What is init_pg?

colossalai/tensor/colo_tensor.py
```diff
@@ -296,3 +228,9 @@ def __mul__(self, other) -> "ColoTensor":
             raise TypeError(f'{type(other)} is not supported in ColoTensor __mul__')
 
     __rmul__ = __mul__
+
+    def to_dist_spec(self, dist_spec: _DistSpec) -> None:
+        self._torch_tensor = DistSpecManager.handle_trans_spec(self.torch_tensor(), self.spec.dist_spec, dist_spec)
+        # self._torch_tensor.requires_grad = self._requires_grad
+        self._size = self._torch_tensor.size()
```
Wesley-Jzy, line 235: I think the attributes in current ColoTensor need to be defined clearly, like size,device.... Maybe they mean either local or global tensor info now.

colossalai/tensor/colo_tensor.py
ver217: Some tensor is initialized with wrong process group. We cannot handle process group change now, so we just copy new process group to the previous DistSpec. It's a temporary design and I may change this function in the future.

ver217: I already removed this arg.

colossalai/tensor/colo_tensor.py
ver217: Shall we add two methods like `is_global()` and `to_global()` ? In this way, we assume ColoTensor is local by default. If you have any good ideas, please let me know.

colossalai/tensor/colo_tensor.py
Wesley-Jzy: ok

colossalai/tensor/colo_tensor.py
Wesley-Jzy: Make sense. I added a global_torch_tensor() in ColoTensor before to get tensors in global view, but it only serves for activations. Maybe we can re-implement it with dictspec.

ver217: We can do this in the next PR.

ver217:
All tests passed.
image

colossalai/tensor/__init__.py
```diff
@@ -1,13 +1,14 @@
-from .spec import ComputePattern, ParallelAction, TensorSpec, ShardPattern
+from .spec import ComputePattern, ParallelAction, TensorSpec
 from .op_wrapper import (
     colo_op_impl,)
 from .colo_tensor import ColoTensor
 from .colo_parameter import ColoParameter
 from .utils import convert_parameter, named_params_with_colotensor
 from ._ops import *
 from .optim.colo_optimizer import ColoOptimizer
+from . import dist_spec
```
feifeibear, line 9: why not name it as DistSpec?

feifeibear: Or DistSignature? to be consistent with our design doc?

ver217: It's a package not class. I just use package to implement the "namespace" in C++.

colossalai/tensor/colo_tensor.py
```diff
@@ -127,85 +132,17 @@ def torch_tensor(self) -> torch.Tensor:
...
-    def is_gathered(self) -> bool:
-        return self._shard_pattern == ShardPattern.NA
+    def set_spec(self, spec: TensorSpec) -> None:
+        spec = copy(spec)
+        self.to_dist_spec(spec.dist_spec)
+        self._spec = spec
 
     def has_spec(self) -> bool:
-        return self._shard_spec is not None and self._shard_spec.num_action > 0
+        return self._spec is not None and self._spec.num_action > 0
```
feifeibear, line 141: we'd better set _spec as a dummy one by default.
I think making it an optional variable may bring some trouble. According to our experiments in ShardTensor design.

colossalai/tensor/__init__.py
feifeibear: OK,that's fine

ver217: I've updated this part in our design doc. I think current implementation is simpler and it's easy to comprehend for users.

colossalai/tensor/_ops/linear.py
```diff
@@ -4,72 +4,50 @@
...
-    elif input_tensor.shard_pattern == ShardPattern.Col:
-        # Splited by 1Dcol
-        assert input_tensor.shape[-1] == weight.size(-1), \
-        'Invalid shapes in 1Drow forward: input={}, weight={}. Expected last dim of input {}.'.format(
-        input_tensor.shape, weight.size, weight.size(-1))
-        input_per_partition = input_tensor.torch_tensor()
-    else:
-        raise NotImplementedError
+    input_tensor.to_dist_spec(
+        dist_spec.shard(weight.spec.dist_spec.process_group, [-1], [weight.spec.dist_spec.process_group.size()]))
```
feifeibear, line 17: The nesting naming is confusion....
a file.and sepc.py also a file containing dist_spec? But dist_spec.py is also a file.

feifeibear: why not just make a file named api.py?

colossalai/tensor/colo_tensor.py
ver217: Oh, yes. Actually our default value of _spec is not None. This code snippet is outdated and I forgot to update.

colossalai/tensor/_ops/linear.py
feifeibear: In short, I think the following usage looks better to developers.
weight.spec.get_process_group()

colossalai/tensor/colo_tensor.py
ver217: updated

colossalai/tensor/_ops/linear.py
Wesley-Jzy: I think the first spec is TensorSpec and the second one is DistSpec. Do we really neead TensorSpec abs?

ver217: Flattern the folder structure now.

ver217: TensorSpec includes a action list now.

ver217: If we move the action list to op rather than parameter, then TensorSpec and DistSpec can be merged into one.

Wesley-Jzy: It's actually introduced by ComputePattern bound to weight.

Wesley-Jzy: Use the weight.spec.get_process_group() mentioned by @feifeibear or add a TODO for ComputePattern code refining.

Wesley-Jzy (approved):

colossalai/tensor/_ops/linear.py
ver217: `spec.get_process_group()` added.

PR #947, opened 2022-05-13
[Tensor] add DistSpec for loss and test_model

Wesley-Jzy:


colossalai/tensor/spec.py
```diff
@@ -84,3 +84,16 @@ def get_action_by_compute_pattern(self, compute_pattern: ComputePattern):
...
+    def get_placement(self):
+        return self.dist_spec.placement
+
+    def is_gathered(self):
+        return self.dist_spec.placement == DistPlacementPattern.REPLICATE \
+            or (len(self.dist_spec.num_partitions) == 1 
+                and self.dist_spec.num_partitions[0] == 1) \
+            or (self.dist_spec.process_group.size() == 1)
+
+    def is_1Dcol(self):
```
ver217, line 97: I think we can design a general module which can derive computation pattern from DistSpec of params in the future. It's OK now.

ver217 (approved):

colossalai/tensor/spec.py
Wesley-Jzy: OK.

PR #971, opened 2022-05-16
[tensor] derive compute pattern from dist spec

ver217:
Derive compute pattern from dist spec, and remove unneeded compute patterns. Only `TP1D`, `ZeRO` and `DP` are remained.
We don't need determine whether a parameter is computed as 1D Row or 1D Column, since they can be derived from dist spec. So as for 1D tensor parallel, we don't need to distinguish 1D Row and 1D Column w.r.t. compute pattern.

### DIFF
--- a/colossalai/tensor/_ops/addmm.py
+++ b/colossalai/tensor/_ops/addmm.py
@@ -70,11 +70,12 @@ def colo_addmm(types, args, kwargs, pg):
 
     # Add communication logic before and after linear call.
     ret_tensor = None
-    if mat2.spec.is_gathered():    # No Model Parallel Applied
+    if not mat2.has_spec():    # No Model Parallel Applied
+        assert mat2.spec.is_gathered(), 'Invalid mat2 spec for native addmm op'
         assert input_tensor.spec.is_gathered(), 'Invalid input spec for native addmm op'
         ret_tensor = ColoTensor.init_from_torch_tensor(
             torch.addmm(input_tensor.torch_tensor(), mat1, mat2.torch_tensor(), beta=beta, alpha=alpha))
-    else:    # Single Model Parallel Applied
+    elif mat2.spec.has_compute_pattern(ComputePattern.TP1D):    # Single Model Parallel Applied
         spec = TensorSpec(dist_spec.replicate(mat2.spec.get_process_group()))
         mat1 = args[1] if isinstance(args[1], ColoTensor) else ColoTensor.init_from_torch_tensor(args[1], spec=spec)
         if mat2.spec.is_1D_row() and input_tensor.spec.is_gathered():
@@ -83,6 +84,8 @@ def colo_addmm(types, args, kwargs, pg):
             ret_tensor = colo_addmm_1Dcol(input_tensor, mat1, mat2, beta, alpha)
         else:
             raise NotImplementedError
+    else:
+        raise NotImplementedError
 
     # building the computing graph, op -> output
     # if GraphGlobalEnv().graph_building:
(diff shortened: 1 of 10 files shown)


colossalai/tensor/_ops/addmm.py
```diff
@@ -70,22 +70,19 @@ def colo_addmm(types, args, kwargs, pg):
 
     # Add communication logic before and after linear call.
     ret_tensor = None
-    if not mat2.has_spec():    # No Model Parallel Applied
-        assert not input_tensor.has_spec(), 'Invalid input spec for native addmm op'
+    if mat2.spec.is_gathered():    # No Model Parallel Applied
```
Wesley-Jzy, line 73: So spec must exist now?

ver217: `has_spec()` actually checks whether the colo tensor has parallel actions. And now, all colo tensors have `.spec`. If TP is used, `has_spec()` must be true. If not, no matter whether `has_spec()` is true is OK.

Wesley-Jzy:
I have no other problems for this mr. But I worry that this change will reduce the flexibility of ColoTensor. Since every ComputePattern is deduced from input type, how will 2D or other parallel be added into the auto-tp1D?

colossalai/tensor/_ops/addmm.py
Wesley-Jzy: I see

feifeibear:
I also doubt if we need to approve this MR right now. Decoupling the computation pattern of OP and the distribution pattern o TENSOR is the main purpose of ColoTensor. This MR merges them together again.

ver217:
> I also doubt if we need to approve this MR right now. Decoupling the computation pattern of OP and the distribution pattern o TENSOR is the main purpose of ColoTensor. This MR merges them together again.

We have to figure out whether shard pattern of parameters is static. If so, the computation pattern is deterministic. In this way, computation pattern and shard pattern are definitely coupled.

ver217:
> I have no other problems for this mr. But I worry that this change will reduce the flexibility of ColoTensor. Since every ComputePattern is deduced from input type, how will 2D or other parallel be added into the auto-tp1D?

We can add computations patters, like `TP2D`.

feifeibear:
> > I also doubt if we need to approve this MR right now. Decoupling the computation pattern of OP and the distribution pattern o TENSOR is the main purpose of ColoTensor. This MR merges them together again.
> 
> We have to figure out whether the shard pattern of parameters is static. If so, the computation pattern is deterministic. In this way, computation patterns and shard patterns are definitely coupled.

The thing will become complicated after involving ZeRO and Gemini. Let's make it simple and leave the compute and dist spec decoupled right now. Another module can be introduced to infer dist spec with compute pattern. It can be implemented as an independent module rather than simply removing the computation pattern like this.

ver217:
> > > I also doubt if we need to approve this MR right now. Decoupling the computation pattern of OP and the distribution pattern o TENSOR is the main purpose of ColoTensor. This MR merges them together again.
> > 
> > 
> > We have to figure out whether the shard pattern of parameters is static. If so, the computation pattern is deterministic. In this way, computation patterns and shard patterns are definitely coupled.
> 
> The thing will become complicated after involving ZeRO and Gemini. Let's make it simple and leave the compute and dist spec decoupled right now. Another module can be introduced to infer dist spec with compute pattern or vice versa. It can be implemented as an independent module rather than simply removing the computation pattern like this.

Yes, I agree we can use a new module. But now, if we don't do this, it's hard to run GPT. Since GPT shares weight among embedding and linear layer. Their computation patterns are different.

ver217:
I will update this MR to show how to support more TPs and ZeRO.

Wesley-Jzy (approved):

PR #1004, opened 2022-05-19
[Tensor] add test checkpoint

Wesley-Jzy:
1. add test_checkpoint
2. remove trick for old ColoTensor

### DIFF
--- a/tests/test_tensor/test_checkpoint.py
+++ b/tests/test_tensor/test_checkpoint.py
@@ -1,0 +1,30 @@
+from torch.utils.checkpoint import checkpoint
+import torch
+from colossalai.utils import ColoInitContext
+
+
+def test_checkpoint():
+    class ModelwithCheckPoint(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.proj = torch.nn.Linear(32, 32)
+            self.proj_ckpt = torch.nn.Linear(32, 2)
+
+        def forward(self, x):
+            x = self.proj(x)
+            x = checkpoint(self.proj_ckpt, x)
+            assert x.requires_grad
+            return x
+
+    with ColoInitContext():
+        model = ModelwithCheckPoint()
+
+    x = torch.rand(2,32)
+    out = model(x)
+    loss = torch.sum(out)
+    loss.backward()
+    assert loss.requires_grad
+
+
+if __name__ == '__main__':
+    test_checkpoint()
--- a/tests/test_tensor/test_gpt.py
+++ b/tests/test_tensor/test_gpt.py
@@ -22,58 +22,6 @@ from dataclasses import fields
 from tests.test_tensor._utils import tensor_equal
 
 
-def _post_init_colotensor(self):
-    class_fields = fields(self)
-    # Safety and consistency checks
-    if len(class_fields) == 0:
-        raise ValueError(f"{self.__class__.__name__} has no fields.")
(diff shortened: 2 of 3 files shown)


tests/test_tensor/test_checkpoint.py
```diff
@@ -0,0 +1,30 @@
+from torch.utils.checkpoint import checkpoint
+import torch
+from colossalai.utils import ColoInitContext
+
+
+def test_checkpoint():
+    class ModelwithCheckPoint(torch.nn.Module):
```
feifeibear, line 7: move the model define to common. maybe some other places use it again.

tests/test_tensor/test_checkpoint.py
```diff
@@ -0,0 +1,30 @@
...
+        def __init__(self):
+            super().__init__()
+            self.proj = torch.nn.Linear(32, 32)
```
feifeibear, line 10: use a smaller dim, make unitest faster.

tests/test_tensor/test_checkpoint.py
feifeibear: Why not use simpleNet and init is with checkpoint=True

tests/test_tensor/test_checkpoint.py
```diff
@@ -0,0 +1,30 @@
...
+            self.proj_ckpt = torch.nn.Linear(32, 2)
+
+        def forward(self, x):
+            x = self.proj(x)
+            x = checkpoint(self.proj_ckpt, x)
+            assert x.requires_grad
```
feifeibear, line 16: I do think these assets is enought to check the grad checkpointing....

feifeibear:
I concern the necessarity of this PR. Because it lacks an end-to-end grad/param checking for gradient checkpointing.

tests/test_tensor/test_checkpoint.py
Wesley-Jzy: I don't think it's a right usage for checkpoint. Using CheckPointModule as the last layer is wrong because our CheckPointModule uses forward as its function, so no process_output will be invoked. If it's the last layer, grad will disappear.

tests/test_tensor/test_checkpoint.py
Wesley-Jzy: Done

tests/test_tensor/test_checkpoint.py
Wesley-Jzy: In short, I mean the class uses forward() directly instead of apply(). So it's output without grad will not be post-processed by torch itself. It can only be a middle layer of a model.

Wesley-Jzy: So would you like a new class in util or an update for simple-net?

Wesley-Jzy:
> I concern the necessarity of this PR. Because it lacks an end-to-end grad/param checking for gradient checkpointing.

Checking the existence of grad only is because after reading the source code, I found the only useful code of checkpoint is to put fwd in no_grad() and re-run it in bwd.
But I know your conern. I will add correctness checking for it.

tests/test_tensor/test_checkpoint.py
Wesley-Jzy: I can use simple-net to do it. But current simple-net is wrong for checkpoint, I do some updates.

PR #1011, opened 2022-05-21
[hotfix] fix some bugs caused by size mismatch.

YuliangLiu0306:


colossalai/cli/benchmark/benchmark.py
```diff
@@ -63,6 +63,10 @@ def run_dist_profiling(rank: int, world_size: int, port_list: List[int], config_
         colossalai.launch(config=config, rank=rank, world_size=world_size, host='localhost', port=port, backend='nccl')
         timer = MultiTimer()
 
+        # 1D parallel should be skipped if batch size is not able to be divided exactly by 1D parallel size.
+        if config.parallel.tensor.mode == '1d' and hyperparams.batch_size % config.parallel.tensor.size != 0:
+            continue
```
FrankLeeeee, line 68: 1D tensor parallelism does not split the batch dimension.

colossalai/cli/benchmark/utils.py
```diff
@@ -48,9 +48,15 @@ def find_all_configs(device_cnt: int) -> List[Dict]:
     """
 
     def _is_square(num):
+        # 2D parallel should be implemented with at least 2 devices.
+        if num <= 1:
+            return False
```
FrankLeeeee, line 53: Can we show some logs so that the user knows what goes wrong?

colossalai/cli/benchmark/benchmark.py
```diff
@@ -63,6 +65,13 @@ def run_dist_profiling(rank: int, world_size: int, port_list: List[int], config_
         colossalai.launch(config=config, rank=rank, world_size=world_size, host='localhost', port=port, backend='nccl')
         timer = MultiTimer()
 
+        # 1D parallel should be skipped if in_features or out_features is not able to be divided exactly by 1D parallel size.
+        if config.parallel.tensor.mode == '1d' and hyperparams.dimension % config.parallel.tensor.size != 0:
+            logger.warning(
+                "1D parallel will be skipped because in_features or out_features is not able to be divided exactly by 1D parallel size.",
+                ranks=[0])
```
FrankLeeeee, line 72: Use `print` instead in CLI for consistency.

FrankLeeeee: e.g. print("Warning: xxx")

FrankLeeeee (approved):

PR #1015, opened 2022-05-24
[tensor] ColoTensor supports ZeRo

ver217:
Usage:
```python
chunk_size = 38 * 1024**2 if use_chunk else None
chunk_manager = ChunkManager(chunk_size, enable_distributed_storage=use_zero)
model = ColoDDPV2(model, chunk_manager)
```
`chunk_size=None` means chunk is not used.


colossalai/nn/parallel.py
```diff
@@ -3,8 +3,10 @@
 from colossalai.core import global_context as gpc
 from colossalai.context import ParallelMode
 from functools import partial
+from colossalai.zero.utils.zero_hook_v2 import ZeROHookV2
+from colossalai.tensor import ChunkManager, use_param_op_hooks, TensorState
 
-__all__ = ['ColoDDP']
+__all__ = ['ColoDDP', 'ColoDDPV2']
```
Wesley-Jzy, line 9: I'm not familiar with previous zero. What DDPV2 means?

ver217: `ColoDDP` is like torch ddp, which only supports traditional DP. `ColoDDPV2` supports ZeRO-DP and traditional DP, so I still name it `DDP`. `ColoDDP` and `ColoDDPV2` are both for `ColoTensor`, so I name the second one `V2`.

colossalai/tensor/chunk.py
```diff
@@ -0,0 +1,265 @@
...
+        self.data.data = self.data.to(get_current_device())
+        if is_all_reduce:
+            dist.all_reduce(self.data, group=gpc.get_group(ParallelMode.DATA))
+        else:
+            dist.reduce(self.data, self.src_rank, group=gpc.get_group(ParallelMode.DATA))
+        self._update_tensors_ptr()
+        self._update_tensors_state(TensorState.HOLD)
+
+    def tensor_trans_state(self, tensor: torch.Tensor, tensor_state: TensorState) -> None:
+        assert tensor != TensorState.FREE, 'Can only set a chunk of tesors to FREE'
```
Wesley-Jzy, line 110: typo: tensor

colossalai/tensor/chunk.py
```diff
@@ -0,0 +1,265 @@
...
+    def __init__(self,
+                 chunk_size: Optional[int],
+                 enable_distributed_storage: bool = False,
+                 init_device: Optional[torch.device] = None) -> None:
+        assert chunk_size is None or chunk_size > 0
+        self.chunk_size = chunk_size
+        self.enable_distributed_storage = enable_distributed_storage
+        self.device = init_device or get_current_device()
+        self.chunk_groups: Dict[str, Deque[Chunk]] = {}
+        self.tensor_chunk_map: Dict[torch.Tensor, Chunk] = {}
```
Wesley-Jzy, line 169: If a tensor is unused, where will the map be updated?

colossalai/tensor/chunk.py
ver217: fixed

colossalai/tensor/chunk.py
ver217: ChunkManager only manage used params. Unused params don't take part in computation and won't be updated. We can even remove them.

colossalai/tensor/chunk.py
```diff
@@ -0,0 +1,265 @@
...
+            tensor.data = self.data[tensor_info.offset:tensor_info.end].view(tensor.shape)
+
+    def _update_tensors_state(self, next_state: TensorState, prev_state: Optional[TensorState] = None):
+        for tensor_info in self.tensors_info.values():
+            if prev_state is None or tensor_info.state == prev_state:
+                tensor_info.state = next_state
+
+    def access(self) -> None:
+        if not self.is_src_rank:
+            self.data.storage().resize_(self.size)
```
feifeibear, line 88: how about the device? can you make sure the storage is on current device?

ver217: fixed

feifeibear:
The main concern is about suspending parameters.
For example
```
class Net(_)
  def __init__
       self.fc1 = Linear()
       self.suspend_param = torch.Paramer(..)
````
The self.suspend_param will occur in module.parameters(). So your DDPV2 will append it to the chunk manager.
Managing the state of the param will be a disaster for your design?

**Comment by ver217** [comment]
> The main concern is about suspending parameters. For example
> 
> ```
> class Net(_)
>   def __init__
>        self.fc1 = Linear()
>        self.suspend_param = torch.Paramer(..)
> ```
> 
> The self.suspend_param will occur in module.parameters(). So your DDPV2 will append it to the chunk manager. Managing the state of the param will be a disaster for your design?

Finally, we will use computation graph to derive the computation order of params and filter those unsed params. Chunk manager only manages used params

File: colossalai/tensor/chunk.py [hunk 5]
```diff
@@ -0,0 +1,264 @@
...
+        self.device = init_device or get_current_device()
+        self.chunk_groups: Dict[str, Deque[Chunk]] = {}
+        self.tensor_chunk_map: Dict[torch.Tensor, Chunk] = {}
+        self.accessed_chunks: Set[Chunk] = set()
+        self.lazy_release_tensors: List[torch.Tensor] = []
+        if enable_distributed_storage and chunk_size is None:
+            self.rank_load: Dict[str, torch.Tensor] = {}
+
+    def append_tensor(self, tensor: torch.Tensor, group_name: str) -> None:
+        assert tensor not in self.tensor_chunk_map
```
feifeibear, line 175: Can not handle shared param like tie weight?

colossalai/tensor/chunk.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self,
+                 chunk_size: Optional[int],
+                 enable_distributed_storage: bool = False,
+                 init_device: Optional[torch.device] = None) -> None:
+        assert chunk_size is None or chunk_size > 0
+        self.chunk_size = chunk_size
+        self.enable_distributed_storage = enable_distributed_storage
+        self.device = init_device or get_current_device()
+        self.chunk_groups: Dict[str, Deque[Chunk]] = {}
```
feifeibear, line 167: why use a deque? I think you use it as a List.

colossalai/tensor/param_op_hook.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @staticmethod
+    def backward(ctx, *grads):
+        for hook in _ParamOpHookWrapper.hooks:
+            hook.pre_backward(ctx.params)
+        return (None,) + grads
+
+
+@contextmanager
+def use_param_op_hooks(*hooks: ParamOpHook):
```
feifeibear, line 65: type hint is wrong.
It is a Tupe[ParamOpHook, ...]?

colossalai/nn/parallel.py
```diff
@@ -76,3 +78,54 @@ def zero_grad(self, set_to_none: bool = False) -> None:
...
+
+    def backward(self, loss: torch.Tensor):
+        with self.param_op_hook.switch_to_backward(), use_param_op_hooks(self.param_op_hook):
+            loss.backward()
+        self.chunk_manager.exec_lazy_release()
+        for p in self.module.parameters():
+            if self.chunk_manager.is_chunk_free(p) or not p.requires_grad:
+                p.grad = None
+            else:
+                p.grad = p.data
```
feifeibear, line 116: what does it mean?
reusing grad fp16 with param fp16?
should it be p.data = p.grad?

colossalai/tensor/chunk.py
```diff
@@ -0,0 +1,264 @@
...
+        if chunk in self.accessed_chunks:
+            return
+        chunk.access()
+        self.accessed_chunks.add(chunk)
+
+    def release_chunk(self, tensor: torch.Tensor) -> None:
+        if not self.enable_distributed_storage:
+            return
+        chunk = self.tensor_chunk_map[tensor]
+        if chunk not in self.accessed_chunks:
```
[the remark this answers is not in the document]
ver217: If a rank store a chunk, its initial state is HOLD, but the rank should broadcast.

colossalai/tensor/chunk.py
ver217: Can handle. It won't be appended twice.

colossalai/tensor/chunk.py
ver217: As it won't be accessed randomly by index, just for efficiency.

colossalai/tensor/param_op_hook.py
ver217: type hint for *args should be the type of each element.

colossalai/nn/parallel.py
```diff
@@ -76,3 +78,54 @@ def zero_grad(self, set_to_none: bool = False) -> None:
...
+
+    def grad_handle(self, p, grad):
+        empty_grad = torch.empty_like(grad)
+        free_storage(empty_grad)
+        with torch._C.DisableTorchFunction():
```
feifeibear, line 121: why use this ctx?

colossalai/nn/parallel.py
```diff
@@ -76,3 +78,54 @@ def zero_grad(self, set_to_none: bool = False) -> None:
...
+            self.chunk_manager.trans_tensor_state(p, TensorState.READY_FOR_REDUCE)
+            if self.dp_world_size > 1:
+                grad = grad / self.dp_world_size
+            self.chunk_manager.update_tensor(p, grad)
```
feifeibear, line 125: here, chunk reuse?

colossalai/nn/parallel.py
ver217: Set p.grad to correct ptr here. p.data saves grad, due to reuse.

colossalai/nn/parallel.py
ver217: yes

colossalai/nn/parallel.py
ver217: in case some ColoParameter in this functions trigger hooks. Actually them shouldn't trigger hooks in chunk manager. It leads to bugs in the previous code, so I add this ctx.

colossalai/nn/parallel.py
feifeibear: if a chunk is moved from gpu to cpu later. This line makes p.grad point to an old memory space.

colossalai/tensor/chunk.py
feifeibear: why not randomly? Eviction strategy will access the farthest reference chunk? It is definitely not sequential access.

colossalai/tensor/chunk.py
feifeibear: are they points to the same chunk memory slice?

colossalai/tensor/chunk.py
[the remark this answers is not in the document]
feifeibear: Cannot understand what you mean? You can say Chinese here.

colossalai/tensor/chunk.py
ver217: If you mean shared param like embedding weight and classifier weight in GPT, yes, they are actually the same parameter. If they are on different pipeline parallel ranks, they are different parameters and points to different chunks.

colossalai/tensor/param_op_hook.py
feifeibear: I see. Good to know.

colossalai/tensor/chunk.py
ver217: We will have a another list like `hold_cuda_list` which is in order.

colossalai/nn/parallel.py
ver217: No, move device will update p.data at the same time.

colossalai/tensor/chunk.py
feifeibear: I think this PR does not consider PP cases? I mean the first case. 
If I call append_tensor on the same tensor twice. Then will I trigger the assert? Because the tensor has already been in self.tensor_chunk_map after the first append_tensor call.

ver217: Yes. But we should not append a same parameter twice. Even though they are used in two layers, `module.parameters()` just yield one parameter once.

colossalai/nn/parallel.py
feifeibear: This line makes p.grad -> p.data (addr1)
Afterwards, you move chunk
let p.data (addr2)
However, p.grad still points to addr1.

colossalai/tensor/chunk.py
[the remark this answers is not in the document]
ver217: For example, dp size = 2 here, and rank0 stores chunk0. The initial state of chunk0 in rank0 is HOLD, which in rank1 is FREE. We want to access chunk0 now, even rank0 has chunk0, rank0 have to do `broadcast()`. We can determine whether a rank has a chunk by state, but we cannot know whether the `broadcast()` is done by state.

colossalai/nn/parallel.py
ver217: If so, just set p.grad again, or move this code snippets to `optimizer.step()` after device moving is done.

ver217: To check grad in unit test, I just set `p.grad` here.

feifeibear: Is it necessary to build a dict {param: chunk slice} to index grad and its true memory space (may reuse with param.data} ?

colossalai/tensor/chunk.py
feifeibear: I see. That make sense

colossalai/tensor/chunk.py
```diff
@@ -0,0 +1,264 @@
...
+        # this function only apply valid state transformation
+        # invalid calls will be ignored and nothing changes
+        if (self.tensors_info[tensor].state, tensor_state) not in STATE_TRANS:
+            # print(
+            #     f'WARNING: Rank{gpc.get_global_rank()} apply invalid state trans: {self.tensors_info[tensor].state} to {tensor_state}'
+            # )
+            return
+        self.tensors_info[tensor].state = tensor_state
+
+    def update_tensor(self, tensor: torch.Tensor, data_slice: torch.Tensor) -> None:
```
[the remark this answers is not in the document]
ver217: done

colossalai/nn/parallel.py
ver217: Not necessary now I think. p.grad should always point to the chunk slice memory of p, as we always reuse now. If not reuse, I think it's necessary.

feifeibear: OK, add it if necessary later.

feifeibear (approved):

PR #1016, opened 2022-05-24
[graph]add graph module to colossalai

YuliangLiu0306:


colossalai/graph/passes/adding_split_node_pass.py
```diff
@@ -0,0 +1,54 @@
+import torch
+
+from torch.fx import symbolic_trace
+from torch.fx.node import Node
+from torch.fx.passes.split_module import split_module
+
+
+def pipe_split():
+    pass
```
FrankLeeeee, line 9: why pass?

colossalai/graph/passes/adding_split_node_pass.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def balanced_split_node(gm: torch.fx.GraphModule, pp_size: int):
+    mod_graph = gm.graph
+    total_param_amount = 0
+    for param in mod_graph.owning_module.parameters():
+        total_param_amount += param.numel()
```
FrankLeeeee, line 16: Will calling `.numel()` work with ColossalAI context?

colossalai/graph/passes/adding_split_node_pass.py
```diff
@@ -0,0 +1,54 @@
+import torch
+
+from torch.fx import symbolic_trace
+from torch.fx.node import Node
+from torch.fx.passes.split_module import split_module
```
FrankLeeeee, line 5: Can you do a torch version check when importing `torch.fx`. Colossal-AI supports minimum torch 1.8 and I think `torch.fx` does not exist in the older versions.

colossalai/graph/passes/adding_split_node_pass.py
YuliangLiu0306: This function is just used to create a node on graph to annotate the split location. The split node will be erased after split_module pass, so it should never be used during runtime.

colossalai/graph/passes/adding_split_node_pass.py
YuliangLiu0306: Yes, the param with ColoInit will be a ColoTensor, it has numel method and works as expected.

colossalai/graph/__init__.py
```diff
@@ -0,0 +1,5 @@
+import torch
+from packaging import version
+
+assert version.parse(torch.__version__) >= version.parse(
+    '1.10'), "We require at least 1.10 pytorch version when using graph module for a stable torch.fx interface."
```
FrankLeeeee, line 5: Don't do this as if other modules import this module, then users with pytorch < 1.10 cannot use colossalai at all.

colossalai/graph/passes/shard_1d_pass.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def weight_split(weight, dim):
+    num_partition = gpc.get_world_size(ParallelMode.TENSOR)
+    shape = weight.shape
+    length = shape[dim] // num_partition
+    sharded_weight_list = []
+    for i in range(num_partition):
+        sharded_weight_list.append(weight.narrow(dim, i * length, length))
+    return sharded_weight_list[gpc.get_local_rank(ParallelMode.PARALLEL_1D)]
```
FrankLeeeee, line 36: I would suggest to utilize the `dist_spec` in ColoTensor.

colossalai/graph/__init__.py
YuliangLiu0306: ok

FrankLeeeee:
More documentation would help with readability and maintainability.

colossalai/graph/tests/test_parallel_1d.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+
+from functools import partial
```
FrankLeeeee, line 4: This file should be put into the `tests` folder, otherwise it will not be checked by CI. The graph modules can be moved to the `colossalai.fx` module as well.

colossalai/graph/passes/shard_1d_pass.py
FrankLeeeee: If `ColoTensor` is not used for now, let's add a TODO comment for future replacement of torch tensor with colotensor.

PR #1024, opened 2022-05-25
[p2p]add object list send/recv

YuliangLiu0306:


colossalai/communication/p2p.py
```diff
@@ -79,50 +79,108 @@ def _communicate(tensor_send_next: torch.Tensor = None,
...
-    if tensor_send_next is not None:
-        send_next_op = dist.P2POp(dist.isend, tensor_send_next, next_rank)
-        ops.append(send_next_op)
+        if isinstance(tensor_recv_next, torch.Tensor):
+            recv_next_op = dist.P2POp(dist.irecv, tensor_recv_next, next_rank)
+            ops.append(recv_next_op)
+        else:
+            for tensor_recv in tensor_recv_next:
+                recv_next_op = dist.P2POp(dist.irecv, tensor_recv, next_rank)
+                ops.append(recv_next_op)
```
FrankLeeeee, line 173: This code snippet is highly redundant, can you make it reusable?

colossalai/communication/p2p.py
```diff
@@ -79,50 +79,108 @@ def _communicate(tensor_send_next: torch.Tensor = None,
...
+
+    if object_send_next is not None:
+        if isinstance(object_send_next, torch.Tensor):
```
feifeibear, line 176: Will it be better to bandwidth utilisation if we concat tensors into a flatten tensor and set it out? But it will consumes more memory to prepare a flatten buff.

YuliangLiu0306: I think concat tensors will achieve a better performance in communication, but it may not make a big difference in pipeline parallel scenario, because there are just several tensors to transport, not numbers of tiny tensors.

colossalai/communication/p2p.py
```diff
@@ -79,50 +79,108 @@ def _communicate(tensor_send_next: torch.Tensor = None,
...
+            send_next_op = dist.P2POp(dist.isend, object_send_next, next_rank)
+            ops.append(send_next_op)
+        else:
+            for tensor_send in object_send_next:
+                send_next_op = dist.P2POp(dist.isend, tensor_send, next_rank)
+                ops.append(send_next_op)
```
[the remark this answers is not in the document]
YuliangLiu0306: The API broadcast_object_list sends the data to all ranks in group. I think this pr is something like send_object_list.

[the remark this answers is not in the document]
FrankLeeeee: One way to do this is to create a process group for each pair of adjacent stages, and broadcast from one rank to the other rank in this process group.

FrankLeeeee (approved):

PR #1025, opened 2022-05-25
[tensor] ColoTensor supports lazy init

ver217:
```python
# init a tensor without storage, x is freed
x = ColoTensor(torch.rand(5, 5), lazy_init=True)
# alloc memory for a freed tensor
x.malloc()
# free it again
x.free()
# check whether it's freed
print(x.is_freed)
```
Tensor's shape, device and dtype won't change after free()/malloc().
A freed tensor cannot take part in computation and communication

### DIFF
--- a/colossalai/tensor/colo_parameter.py
+++ b/colossalai/tensor/colo_parameter.py
@@ -13,7 +13,8 @@ class ColoParameter(ColoTensor):
     def __new__(cls,
                 data: torch.Tensor,
                 requires_grad: bool = True,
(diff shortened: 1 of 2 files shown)


feifeibear:
Explicitly calling malloc and free can not be named as lazy!
Lazy means when you use it, you implicitly allocate memory for it.
Such API is not user-friendly.

ver217:
> Explicitly calling malloc and free can not be named as lazy! Lazy means when you use it, you implicitly allocate memory for it. Such API is not user-friendly.

We can do this by allocate memory in __torch_function__, however, the weight initialization is still a problem.

PR #1041, opened 2022-05-30
[Tensor] add Parameter inheritance for ColoParameter

Wesley-Jzy:


colossalai/tensor/colo_parameter.py
```diff
@@ -4,22 +4,21 @@
...
-
-class ColoParameter(ColoTensor):
+class ColoParameter(ColoTensor, torch.nn.Parameter):
     r"""A kind of ColoTensor to be considered as a module parameter.
 
     """
 
     def __new__(cls,
-                data: torch.Tensor,
+                data: torch.Tensor = None,
```
feifeibear, line 13: = torch.empty(0)

feifeibear: if = None it is Optional[torch.Tensor]

Wesley-Jzy: Done

colossalai/tensor/colo_parameter.py
```diff
@@ -3,23 +3,23 @@
 import torch
 from colossalai.tensor import TensorSpec, distspec
 from copy import copy
+from typing import Optional
 
-
-class ColoParameter(ColoTensor):
+class ColoParameter(ColoTensor, torch.nn.Parameter):
```
feifeibear, line 8: Can ColoParameter use __torch_function__?
Can you add a test for this function?

feifeibear (approved):
LGTM

colossalai/tensor/colo_parameter.py
Wesley-Jzy: It works.
image

PR #1045, opened 2022-05-31
[hotfix] fix dist spec mgr

ver217:


colossalai/tensor/dist_spec_mgr.py
```diff
@@ -34,7 +34,7 @@ def _shard_as(tensor: torch.Tensor, old_dist_spec: _DistSpec, dist_spec: _DistSp
             chunk_size = divide(tensor.size(dim), dist_spec.num_partitions[i])
             chunk = chunk.narrow(dim, idx // num_parts * chunk_size, chunk_size)
             idx %= num_parts
-        return chunk.detach().contiguous()
+        return chunk.clone().detach().contiguous()
```
feifeibear, line 37: _shard_as is only called on model data before iteration, right?

ver217: It can also be called on activation, but it's not in-place operation for activation.

feifeibear: then clone will introduce non-trivial overhead. especially clone on CPU.

ver217: But without `clone()`, there may be side-effect. E.g. user modifies the original tensor after `_shard_as()`, then the content shard is modified as well.

feifeibear (approved):

Issue #1061, opened 2022-06-03
[PROPOSAL]: reorganize colotensor directories


feifeibear:
### Proposal

A new tensor data structure named colotensor is under development in directory colossalai/tensor. However, too many functions, including customized operators, optimizers, modules and so on, are placed in this directory, making this name tensor can not explains itself. I proposed to clean the directory and move irrelevant functions to other directories.

### Self-service

- [x] I'd be willing to do some initial work on this proposal myself.

PR #1062, opened 2022-06-03
reorgnize colotensor directory

feifeibear:
#1061 
I am cleaning the tensor directory during the holiday, so it has no impact on our working day development.
In the short, module,  _ops and optimizer relevant code are moved to colossalai.nn.
The directory tensor should has no dependency on colosssalai.nn to avoid possible circle import.


feifeibear:
The CI may fail because someone is using the test machine.
I will paste a screenshot when the machine is free.

feifeibear:
image

FrankLeeeee (approved):

Issue #1060, opened 2022-06-02
[BUG]: Tensorboard add_scalar error


super-dainiu:
### 🐛 Describe the bug

The original code for add_scalar is,
```
self.writer.add_scalar(f'{metric_name}/{mode}', val, trainer.cur_step)
```
But I believe the *val* should be modified to *float(val)* because some metrics might have somehow converted to the type of *str*.
```
self.writer.add_scalar(f'{metric_name}/{mode}', float(val), trainer.cur_step)
```
Otherwise, the following bug happens when I trained ViT on CIFAR-100
![image](https://user-images.githubusercontent.com/78588128/171671239-43b1ec70-7fe3-41a7-b939-fbfe722d6874.png)

PR #1071, opened 2022-06-06
[hotfix] fix issue #1060

feifeibear:


FrankLeeeee:
There is no issue 1080?

feifeibear:
> There is no issue 1080?

it is 1060

PR #1077, opened 2022-06-07
[tensor] refactor chunk mgr and impl MemStatsCollectorV2

ver217:
Refactor chunk mgr for easily monitoring memory usage.

### DIFF
--- a/colossalai/tensor/chunk.py
+++ b/colossalai/tensor/chunk.py
@@ -2,7 +2,7 @@ import torch
 import torch.distributed as dist
 from dataclasses import dataclass
 from enum import Enum
-from typing import Optional, Dict, Deque, Set, List
+from typing import Optional, Dict, Deque, Set, List, FrozenSet, Iterable
 from collections import deque
 from colossalai.core import global_context as gpc
 from colossalai.context import ParallelMode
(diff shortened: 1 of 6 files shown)


colossalai/tensor/chunk.py
```diff
@@ -236,10 +235,9 @@ def access_chunk(self, tensor: torch.Tensor) -> None:
         self.accessed_chunks.add(chunk)
         self.total_mem[chunk.device_type] += chunk.mem
 
-    def release_chunk(self, tensor: torch.Tensor) -> None:
+    def release_chunk(self, chunk: Chunk) -> None:
```
feifeibear, line 238: will the API changing affects our old code?

colossalai/tensor/chunk.py
```diff
@@ -340,3 +332,17 @@ def copy_chunk_group(self, dest_group_name: str, src_group_name: str):
         for dest_chunk, src_chunk in zip(self.chunk_groups[dest_group_name], self.chunk_groups[src_group_name]):
             if not dest_chunk.is_free:
                 dest_chunk.copy_(src_chunk)
+
+    def get_chunks(self, tensors: Iterable[torch.Tensor]) -> FrozenSet[Chunk]:
+        return frozenset([self.get_chunk(tensor) for tensor in tensors])
+
+    def add_extern_static_tensor(self, tensor: torch.Tensor) -> None:
```
feifeibear, line 339: I think you need to make sure the static tensor is not registered as a chunk managed tensor later.

ver217: Yes. Done.

colossalai/tensor/chunk.py
ver217: Chunk manager's methods should not be called by users directly. I think user's code won't be influenced by this PR.

feifeibear (approved):

PR #1086, opened 2022-06-09
[workflow] disable p2p via shared memory on non-nvlink machine

FrankLeeeee:


binmakeswell (approved):

feifeibear:
why do we need to disable p2p via SM?

FrankLeeeee:
> why do we need to disable p2p via SM?

This is only for CI as the test got stuck inside docker.

PR #1091, opened 2022-06-09
[Tensor] fix equal assert

Wesley-Jzy:


tests/test_tensor/test_model.py
```diff
@@ -353,5 +341,5 @@ def _test_pretrain_load(world_size):
 if __name__ == '__main__':
     # test_model_parameters()
     # test_colo_optimizer()
-    # test_model(4)
-    _test_pretrain_load(4)
+    test_model(4)
+    #_test_pretrain_load(4)
```
FrankLeeeee, line 345: why comment out this test?

Wesley-Jzy: It will not influence pytest. I just run error case by main().

FrankLeeeee: Should we keep the test_pretrain_load test case?

Wesley-Jzy: It's ok to keep it _test_pretrain_load.

Wesley-Jzy: Updated.

FrankLeeeee (approved):

PR #1097, opened 2022-06-10
[tensor] refactor param op hook

ver217:
`torch.autograd.Function` ensures the device of original inputs and grads are the same. Since zero hook may change the device of input in pre fwd, we execute pre fwd out of `autograd.Function`.

### DIFF
--- a/colossalai/tensor/param_op_hook.py
+++ b/colossalai/tensor/param_op_hook.py
@@ -5,6 +5,11 @@ from typing import List, Tuple, Any
 
 
 class ParamOpHook(ABC):
+    """Hook which is triggered by each operation when operands contain ColoParameter.
+    To customize it, you must inherit this abstract class, and implement ``pre_forward``,
+    ``post_forward``, ``pre_backward`` and ``post_backward``. These four methods take a list
+    of ColoParameter.
+    """
 
     @abstractmethod
     def pre_forward(self, params: List[torch.Tensor]) -> None:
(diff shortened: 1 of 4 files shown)


FrankLeeeee:
Can you add some documentation for better readability and maintainability.

ver217:
> Can you add some documentation for better readability and maintainability?

DONE

colossalai/tensor/param_op_hook.py
```diff
@@ -23,25 +28,78 @@ def post_backward(self, params: List[torch.Tensor]) -> None:
...
 
+    @staticmethod
+    @contextmanager
+    def use_hooks(*hooks: ParamOpHook):
+        """Change the param op hooks you use. Nested calling is allowed.
+
+        Example::
+            >>> with ParamOpHookManager.use_hooks(*hooks):
+            >>>     do_something()
+            >>>     with ParamOpHookManager.use_hooks():
```
feifeibear, line 45: why nest ParamOpHookManager context?

ver217: In case some users want to nest. In most cases, we don't need to nest.

feifeibear: I think users should not know what is OpHook. It is not a common-sense concept. Why do they need to have a manager?

ver217: They can extend profiler by this manager. E.g. before op sample sth, after op sample sth. Indeed, it's for advanced developers.

feifeibear (approved):

PR #1100, opened 2022-06-10
[hotfix]change to fit latest p2p

YuliangLiu0306:


colossalai/engine/schedule/_pipeline_schedule.py
```diff
@@ -86,7 +86,10 @@ def __init__(self,
 
         self.num_microbatches = num_microbatches
         self.dtype = torch.float
-        self.tensor_shape = tensor_shape
+        if isinstance(tensor_shape, torch.Size):
+            self.tensor_shape = tensor_shape
+        else:
+            self.tensor_shape = torch.Size(tensor_shape)
```
feifeibear, line 92: you have to make sure tensor_shape is iterable!
For example torch.Size(1) is invalid.

YuliangLiu0306: The type of tensor shape should be torch.Size, Tuple[int], or List[int]. I will add an assertion to make sure tensor_shape is not an int here.

FrankLeeeee (approved):

PR #1108, opened 2022-06-14
[pipeline] supported more flexible dataflow control for pipeline parallel training

### LINKED ISSUE #120: Compatabilities to various batch formats[FEATURE]
**Is your feature request related to a problem? Please describe.**
The implementation of ColossalAI seems only transfer tesors in batch (dict) values to device. However the organization of batch format can be various and highly customized (e.g. list-type batches, minibatch dict containing lists of tensors). In such cases the batchsize cannot be correctly determined and causing errors. 

**Describe the solution you'd like**
The `batch_size` in `BaseSchedule.load_batch()` should consider list-type batches and use `len()` instead of `size()`. The same applies for the `BaseSchedule._move_to_device()`, which should consider minibatches containing multiple tensors in a list.

FrankLeeeee:
# Summary

Fixed #120  to support more data formats in the engine schedules and implemented a more flexible tensor-argument mapping mechanism in pipeline schedules.

# Details

- Data in the format of list and tuple is now supported as the model inputs and outputs.
- The data_process_func now accepts the stage output as well so that the user can write their own logic to map the tensors (outputs from the previous stage and those produced by dataloader on the current stage) to the arguments of the model forward function on the current stage.

### DIFF
--- a/colossalai/engine/schedule/_non_pipeline_schedule.py
+++ b/colossalai/engine/schedule/_non_pipeline_schedule.py
@@ -4,9 +4,10 @@
 from typing import Iterable
 
 import torch
-
+import inspect
 from ._base_schedule import BaseSchedule
 from colossalai.utils import conditional_context
+from typing import Callable
 
 
 class NonPipelineSchedule(BaseSchedule):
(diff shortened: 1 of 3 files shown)


colossalai/engine/schedule/_non_pipeline_schedule.py
```diff
@@ -16,10 +17,32 @@ class NonPipelineSchedule(BaseSchedule):
     to update the parameters if it is in training mode.
 
     Args:
-        batch_data_process_func (Callable, optional): The preprocessing function which receives a batch of data,
+        batch_data_process_func (Callable, optional): The preprocessing function which receives a batch of data
```
YuliangLiu0306, line 20: The argument name should be data_process_func.

FrankLeeeee: Ok

YuliangLiu0306 (approved):

Issue #1107, opened 2022-06-14
[FEATURE]: 


feifeibear:
### Describe the feature

Supports static method `_set_params_and_buffers_to_ignore_for_model` in ColoDDPV2.

ColoDDPV2 refers
https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/distributed.py#L1665

_set_params_and_buffers_to_ignore_for_model refers
https://github.com/hpcaitech/ColossalAI/blob/main/colossalai/nn/parallel/data_parallel.py#L93

When using hybrid Parallel, we would like to exclude some params in DP and apply our own parallel approach on them. I believe providing such a method is reasonable for that purpose.

ver217:
What does this mean? I'm sure we don't need all-reduce/reduce grad in this case. But are these params/buffers replica? This determines whether ZeRO can be applied.

PR #1122, opened 2022-06-15
[ddp] add set_params_to_ignore for ColoDDP

ver217:
Implement #1107


feifeibear:
Can we have some unitests?

ver217:
> Can we have some unitests?

Yes, unit test is added

feifeibear (approved):

PR #1127, opened 2022-06-16
[ddp] add save/load state dict for ColoDDP

ver217:


colossalai/nn/parallel/data_parallel.py
```diff
@@ -205,3 +218,208 @@ def zero_grad(self, set_to_none: bool = False) -> None:
...
+            destination._metadata = OrderedDict()
+        destination._metadata[prefix[:-1]] = local_metadata = dict(version=self._version)
+        self._save_to_state_dict(destination, prefix, keep_vars)
+        for hook in self._state_dict_hooks.values():
+            hook_result = hook(self, destination, prefix, local_metadata)
+            if hook_result is not None:
+                destination = hook_result
+        return destination
+
+    def _save_to_state_dict(self, destination, prefix, keep_vars):
```
kurisusnowdeng, line 250: Is this saving only a single global dict, or saving on each data parallel rank?

ver217: * a single global dict,
* yes

kurisusnowdeng (approved):

PR #1128, opened 2022-06-16
[hotfix] fix zero init ctx numel

ver217:


colossalai/zero/init_ctx/init_context.py
```diff
@@ -169,11 +172,18 @@ def _post_context_exec(self):
         torch.set_rng_state(self.cpu_rng_state)
         torch.cuda.set_rng_state(self.cuda_rng_state)
 
+        params = frozenset(self.top_module.parameters())
```
Wesley-Jzy, line 175: Because Module's parameters() function uses set() to store result, I think it will not return duplicated param?

Wesley-Jzy (approved):

PR #1130, opened 2022-06-17
[pipeline] add double stream functionality to pipeline schedule

JThh:


JThh:
Need more experimental efforts

PR #1151, opened 2022-06-21
[gemini] refactor gemini mgr

ver217:
1. Gemini manager breakdown PCIE moving volume into H2D volume and D2H volume.
2. `_get_layout_info()` uses LRU cache.

### DIFF
--- a/colossalai/gemini/gemini_mgr.py
+++ b/colossalai/gemini/gemini_mgr.py
@@ -1,4 +1,5 @@
 import torch
+import functools
 from .memory_tracer.memstats_collector import MemStatsCollectorV2
 from typing import List, Optional, Tuple
 from time import time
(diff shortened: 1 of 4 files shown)


feifeibear (approved):

colossalai/gemini/gemini_mgr.py
```diff
@@ -57,22 +62,19 @@ def adjust_layout(self, chunks: Tuple[Chunk, ...], group_name: str) -> None:
...
-        self._cpu_gpu_move_volume += cuda_demand
+        self._h2d_volume += cuda_demand
 
-    @property
-    def cpu_gpu_move_volume(self):
-        return self._cpu_gpu_move_volume
-
-    # @functools.lru_cache(maxsize=None)
-    # TODO: test lru
+    @functools.lru_cache(maxsize=None)
```
feifeibear, line 70: why add this decorator?
I suppose it works for a recursion function.

ver217: layout info at each computation step won’t change during steady phase. use LRU to speed up this function

PR #1157, opened 2022-06-22
[graph] improve the graph building.

feifeibear:


feifeibear:
remove graph-related logic in torch function ops. Use a decorator `@register_colo_graph` to build the DNN compute graph.
`

colossalai/nn/_ops/linear.py
```diff
@@ -48,23 +48,21 @@ def colo_linear_1Dcol(input_tensor: ColoTensor, weight: ColoTensor, bias: Option
...
 
 
-def colo_linear_1d(mode: str, input_tensor: ColoTensor, weight: ColoTensor, bias: Optional[ColoTensor]) -> ColoTensor:
+def colo_linear_1d(mode: str, input_tensor: ColoTensor, weight: ColoTensor, bias: Optional[ColoTensor]) -> 'ColoTensor':
     assert mode in ('row', 'col')
     funcs = {'row': colo_linear_1Drow, 'col': colo_linear_1Dcol}
     return funcs[mode](input_tensor, weight, bias)
 
 
-@colo_op_impl(F.linear)
```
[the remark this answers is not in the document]
feifeibear: I tried some approaches to implement a nested decorator but no one works.

colossalai/nn/graph/utils.py
```diff
@@ -0,0 +1,50 @@
+import functools
+import torch
+from colossalai.tensor import ColoTensor
+from typing import Callable, List
+from colossalai.nn._ops._utils import convert_to_colo_tensor
+
+
+def register_colo_graph(input_pos: List[int], param_pos: List[int]) -> Callable:
```
FrankLeeeee, line 8: I think purely relying on `__torch_function__` to trace the computation graph has some limitations, e.g. memory consumption due to tensor materialization, does not capture control flow and some operations such as indexing like `out = inputs[0]`.

FrankLeeeee: We have implemented a `Proxy` to fulfill the requirements above.

feifeibear: OK, I see. This method can be covered by proxy. 
I want to make some quick changes to ColoTensor. For example, I want to build a table that contains <Op, start, finish> for all activation tensors and give hints for their memory allocation for inference.

feifeibear: Not exactly, an index in the topological graph.

FrankLeeeee: ok.

FrankLeeeee (approved):

colossalai/nn/graph/utils.py
feifeibear: I might need some coarse-grained information on the graph to try out my ideas. 
I also plan to be involved in the development of fx later.

PR #1163, opened 2022-06-23
remove gather out in parallel action

feifeibear:
the gather_out attribute of the ParallelAction is not friendly to the user. Therefore I remove such an attr. However, I force the to_replicate for Op's tp colomn. This might be optimized later.

### DIFF
--- a/tests/test_tensor/test_model.py
+++ b/tests/test_tensor/test_model.py
@@ -26,10 +26,10 @@ def init_1d_row_linear(weight):
         weight.set_spec(spec)
 
 
-def init_1d_col_linear(weight, gather_out=True):
+def init_1d_col_linear(weight):
     spec = TensorSpec(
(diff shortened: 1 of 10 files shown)


feifeibear:
`python test_gpt.py`
image

tests/test_tensor/test_model.py
```diff
@@ -98,7 +98,7 @@ def run_1d_hybrid_tp(model_name):
             if 'proj2' in name and 'weight' in name:
                 init_1d_row_linear(p)
             if 'classifier' in name and ('weight' in name or 'bias' in name):
-                init_1d_col_linear(p, gather_out=False)
+                init_1d_col_linear(p)
```
feifeibear, line 101: @Wesley-Jzy I make gather_out always True. It still pass the unitest.

ver217 (approved):

PR #1167, opened 2022-06-23
[Tensor] new context manager

feifeibear:
When using ColoTensor, we should have a light-weight launcher. At least, we do not need to launch with a config file.


colossalai/initialize.py
```diff
@@ -470,3 +470,12 @@ def initialize(model: nn.Module,
                     schedule=schedule)
 
     return engine, train_dataloader, test_dataloader, lr_scheduler
+
+
+### The Launcher for ColoTensor ###
+def colo_launch(rank, world_size, host, port, backend, local_rank=None, seed=47):
```
FrankLeeeee, line 476: I think we can directly set the `config` to be `None` by default instead of creating a new dummy launch method.

feifeibear: That's better. But it is confusing for the user to always set None as the 1st arg.

FrankLeeeee: I see. Do we plan to use this function for testing only and remove it in the future? My concern is that the users may confuse this method with the original launch method. Another way to improve this is to change the order of the parameters.

PR #1170, opened 2022-06-24
[zero] sharded optim supports loading local state dict

ver217:


colossalai/zero/sharded_optim/sharded_optim_v2.py
```diff
@@ -128,6 +129,7 @@ def __init__(self,
                 f"After init ShardedOptimizerV2 consumes {self.get_memory_usage()[0] / 1e6} MB CUDA Memory!", ranks=[0])
 
         self._use_memory_tracer = self.model.use_memory_tracer
+        self._process_loaded_states = disposable(self._process_loaded_states_)
```
feifeibear, line 132: run only once?

ver217: Yes, if users load state dict before training, we have to make sure dtype and device of optimizer states are right.

ver217:
> Can you add any unit testing?

OK, I will add

PR #1173, opened 2022-06-24
[tensor] prevent pickle dump error

ver217:


FrankLeeeee:
Is this PR still valid?

PR #1175, opened 2022-06-26
[hotfix]different overflow status lead to communication stuck.

YuliangLiu0306:


colossalai/amp/naive_amp/_fp16_optimizer.py
```diff
@@ -258,24 +258,25 @@ def step(self):
...
-        # Update params from main params.
-        self._update_fp16_param_from_fp32_param()
+            # Update params from main params.
+            self._update_fp16_param_from_fp32_param()
 
-        # Successful update.
-        return True, grad_norm
+            # Successful update.
+            return True, grad_norm
+        else:
```
feifeibear, line 278: why clip grad if overflow?

YuliangLiu0306: Different pipeline ranks may have different overflow status, if the overflow do not join the clip grad norm, the all reduce process will be stuck.

YuliangLiu0306: Another solution add an extra communication for overflow status, then we could skip the clip grad and step if some ranks overflowed.

feifeibear: I see!

feifeibear (approved):

PR #1178, opened 2022-06-27
[tensor] revert local view back

feifeibear:
Revert the .view()/size() back to the local one.
I think it is too risky to override them as the global ones.


feifeibear:
image

Wesley-Jzy (approved):

PR #1187, opened 2022-06-29
[context]use meta tensor to init model lazily.

YuliangLiu0306:


colossalai/utils/model/lazy_init_context.py
```diff
@@ -8,6 +8,13 @@
 import typing
 from typing import List, Callable
 
+
+def _substitute_init_recursively(cls, func):
+    for subcls in cls.__subclasses__():
+        _substitute_init_recursively(subcls, func)
+        func(subcls)
```
FrankLeeeee, line 15: Can we import this method from the ctx utils directly?

YuliangLiu0306: Done

colossalai/utils/model/utils.py
```diff
@@ -3,9 +3,9 @@
 from typing import Optional
 
 
-def _substitute_init_recursively(cls, func):
+def substitute_init_recursively(cls, func):
```
FrankLeeeee, line 6: Not sure changing this API will affect other contexts.

colossalai/utils/model/lazy_init_context.py
```diff
@@ -36,126 +38,128 @@ class LazyInitContext():
...
         has_device = 'device' in inspect.signature(func).parameters
 
-        def layer_lazy_init(*args, **kwargs):
-            if has_device:
-                kwargs['device'] = 'meta'
-            func(*args, **kwargs)
+        def layer_lazy_init(module, *args, **kwargs):
+            self._intercepted_init_func_cache.append(dict(func=func, module=module, args=args, kwargs=kwargs))
+            func(module, *args, **kwargs)
+            module.to('meta')
```
FrankLeeeee, line 107: I am wondering whether this will take the same amount of time as just init the model like normal. If so, I think we can check if kwargs contain `device`, if yes, we set `device` to `meta`, if no, instantiate it and convert to `meta`. This should speed up the process.

colossalai/utils/model/utils.py
YuliangLiu0306: I have already searched in our repo, and change them to fit the new api.

colossalai/utils/model/lazy_init_context.py
YuliangLiu0306: I see. I will make module with device kwargs by pass the normal init.

FrankLeeeee (approved):

PR #1192, opened 2022-06-30
[gemini] make warmup ratio AutoPlacementPolicy configurable

feifeibear:


feifeibear:
```
zero = dict(
    model_config=dict(
        tensor_placement_policy='auto',
        shard_strategy=TensorShardStrategy(),
        reuse_fp16_shard=True,
        **warmup_non_model_data_ratio**=0.9,
    ),
    optimizer_config=dict()
）
```

colossalai/zero/sharded_model/sharded_model_v2.py
```diff
@@ -76,7 +76,9 @@ def __init__(self,
                  fp32_reduce_scatter: bool = False,
                  tensor_placement_policy: str = 'cuda',
                  gradient_predivide_factor: Optional[float] = 1.0,
-                 reuse_fp16_shard: bool = False):
+                 reuse_fp16_shard: bool = False,
+                 *args,
```
ver217, line 80: It is never used. Remove it?

feifeibear: I think it doesn't matter. leaving it here is harmless. Also *args and **kwargs often come together to make the parameters passed in scalable.

ver217: `*args` is usaully in front of kwargs like `xxx=xxx`

ver217 (approved):

PR #1193, opened 2022-06-30
[context]support arbitary module materialization.

YuliangLiu0306:


tests/test_utils/test_materialize_arbitary_lazy_module.py
```diff
@@ -0,0 +1,23 @@
...
+    for param in model.parameters():
+        assert param.is_meta
+    for buffer in model.buffers():
+        assert buffer.is_meta
+    for module in model.children():
+        ctx.lazy_init_parameters(module)
+        for param in module.parameters():
+            assert not param.is_meta
+        for buffer in module.buffers():
+            assert not buffer.is_meta
```
FrankLeeeee, line 19: This test should check whether the model is initialized to the same parameters with and without lazy init given the same seed as well.

YuliangLiu0306: Ok, I will add  numerical check in this test.

FrankLeeeee (approved):

PR #1200, opened 2022-07-05
[Tensor] add cpu group to ddp

feifeibear:


colossalai/nn/parallel/data_parallel.py
```diff
@@ -70,8 +69,10 @@ def __init__(self,
         self.comm_stream: torch.cuda.Stream = torch.cuda.Stream()
         assert process_group
 
-        self.process_group = process_group.dp_process_group()
-        self.dp_world_size = self.process_group.size()
+        # self.process_group = process_group.dp_process_group()
```
feifeibear, line 72: I will remove this line after CI passed.

ver217 (approved):

PR #1203, opened 2022-07-05
[refactor] move process group from _DistSpec to ColoTensor.

feifeibear:
process group of a ColoTensor never change.
We should avoid setting dist spec with a process group as a parameter.

Some significant changes.
1. move process group to ColoTensor
2. colo context init initialize ColoTensor using a DP ProcessGroup
3. allow process group change only if pg is DP-only and dist spec is replicate.

### DIFF
--- a/colossalai/nn/graph/utils.py
+++ b/colossalai/nn/graph/utils.py
@@ -23,7 +23,7 @@ def register_colo_graph(input_pos: List[int], param_pos: List[int]) -> Callable:
         def wrapper(*args, **kwargs):
             param_list = []
             input_list = []
-            # TODO(jiaruifan) find teh pg
+            # TODO(jiaruifang) find the pg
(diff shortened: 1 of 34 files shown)


colossalai/nn/graph/utils.py
```diff
@@ -23,6 +23,7 @@ def register_colo_graph_decorator(func):
         def wrapper(*args, **kwargs):
             param_list = []
             input_list = []
+            # TODO(jiaruifan) find teh pg
```
ver217, line 26: typo

ver217 (approved):

colossalai/nn/graph/utils.py
feifeibear: DONE

feifeibear:
image

PR #1209, opened 2022-07-06
[fx] append tensor spec to weight

Itok2000u:
append tensor spec to weight.


Itok2000u:
Have not finished yet, still needs to change API of tensor spec.

PR #1224, opened 2022-07-07
[fx]get communication size between partitions

YuliangLiu0306:


colossalai/fx/passes/utils.py
```diff
@@ -0,0 +1,27 @@
+import torch
+from typing import Dict, Set
+from torch.fx.node import Node, map_arg
+
+
+def get_comm_size(parent_partition, child_partition):
```
FrankLeeeee, line 6: It is kind of confusing to me as partitions are not in the relationship of parent-child. Is `prev_partition` and `next_partition` better?

FrankLeeeee (approved):

PR #1246, opened 2022-07-11
[fx] methods to get fx graph property.

YuliangLiu0306:


colossalai/fx/passes/utils.py
```diff
@@ -25,3 +27,134 @@ def get_comm_size(prev_partition, next_partition):
...
+def is_leaf(graph: Graph, node: Node):
+    return node in get_leaf(graph)
+
+
+def get_top(graph: Graph):
+    """
+    Given a graph, return top nodes of this graph.
+
+    Note: This method will get the top nodes of given graph excluding ``placeholder`` nodes.
+    """
```
FrankLeeeee, line 61: Top nodes are quite confusing, will `get_placeholder` be better?

YuliangLiu0306: This method is used to get the root node after removing 'placeholder' and 'root' node from the origin fx graph.

FrankLeeeee: Ok, can you add this definition of top node in the docstring as it is not a common terminology?

YuliangLiu0306: Sure

FrankLeeeee (approved):

PR #1247, opened 2022-07-11
[tensor] redistribute among different process groups

feifeibear:


colossalai/tensor/colo_tensor.py
```diff
@@ -186,7 +186,28 @@ def _redistribute(self, dist_spec: _DistSpec) -> None:
...
+            dist_spec (_DistSpec): the new dist spec.
+            pg (Optional[ProcessGroup], optional): the new process group . Defaults to None.
+
+        Returns:
+            ColoTensor: a redistributed colotensor
+        """
+        if pg is not None and pg != self.get_process_group():
+            print('here _redistribute')
+            # if the pg is not equal, convert the current tensor to replicated
+            self._redistribute(ReplicaSpec())
```
1SAA, line 209: I think the updated operand, 'redistribute', always used for non-model data in training.
It natrual that we need to keep 'redistribute' as an autograd operand.
But here, '_redistribute' can't used as an autograd operand.

feifeibear: I see. I will discuss with you offline. 
I hope the ColoTensor can be used independently, we don't need to assume it's only used for training.

YuliangLiu0306 (approved):

PR #1248, opened 2022-07-12
mlp_pass

Itok2000u:
Add transform mlp pass, which splits linear layers either in column dimension or row dimension.


FrankLeeeee:
Please add a unit test to verify this works.

PR #1251, opened 2022-07-12
[fx] add balanced policy v2

YuliangLiu0306:


colossalai/fx/passes/adding_split_node_pass.py
```diff
@@ -39,6 +41,36 @@ def balanced_split_pass(gm: torch.fx.GraphModule, pp_size: int):
     return gm
 
 
+def balanced_split_pass_v2(gm: torch.fx.GraphModule, pp_size: int):
```
FrankLeeeee, line 44: add  a test case for this one to ensure that the numel is counted correctly.

FrankLeeeee (approved):

PR #1257, opened 2022-07-12
[utils] integrated colotensor with lazy init context

FrankLeeeee:


colossalai/utils/model/lazy_init_context.py
```diff
@@ -187,71 +189,61 @@ def lazy_init_parameters(self, model: torch.nn.Module, device='cpu', call_back:
...
+                real_tensor = ColoTensor.from_torch_tensor(real_tensor)
+            else:
+                real_tensor = ColoParameter.from_torch_tensor(real_tensor, requires_grad=tensor.requires_grad)
 
-        # check args for parameter replacement
-        for idx, arg in enumerate(args):
-            arg = _process_arg(arg)
-            args[idx] = arg
+            # convert to distribted mode
+            if hasattr(tensor, 'dist_spec'):
```
feifeibear, line 227: isinstance(tensor, ColoTensor) ?

FrankLeeeee: `dist_spec` is annotated during the IR passes are executed, the IR pass will only set this attribute without effect since `tensor` is a meta tensor. The materialized `real_tensor` is guaranteed to be a ColoTensor and can redistribute.

colossalai/utils/model/lazy_init_context.py
```diff
@@ -187,71 +189,61 @@ def lazy_init_parameters(self, model: torch.nn.Module, device='cpu', call_back:
...
-                submodule_name, param_name = param_full_name.rsplit('.', 1)
-                submodule = model.get_submodule(submodule_name)
+            # look for initialization function
+            if tensor in self._intercepted_nn_init_func_cache:
+                init_func, args, kwargs = self._intercepted_nn_init_func_cache[tensor][-1]
             else:
-                submodule = model
-                param_name = param_full_name
-            setattr(submodule, param_name, real_param)
+                raise RuntimeError(f"{param_full_name} is not associated with a initialization function")
```
YuliangLiu0306, line 213: Some parameters may be initialized in the way like 'self.bias = nn.Parameter(torch.zeros(3,3))'

FrankLeeeee: This is ok. All `torch.nn` modules call `reset_parameters` in `__init__`.

YuliangLiu0306: https://github.com/huggingface/transformers/blob/d4ebd4e112034b4a429ab7f813d7e168e7bb63c3/src/transformers/pytorch_utils.py#L88

FrankLeeeee: Ok I see.

PR #1309, opened 2022-07-14
[hotfix] fix ColoTensor GPT2 unitest

1SAA:


tests/test_tensor/test_gpt2.py
```diff
@@ -67,14 +70,16 @@ def run_gpt(init_spec_func, use_ddp):
...
     init_spec_func(model, pg)
     check_param_equal(model, torch_model, pg)
     model.train()
     torch_model.train()
-    set_seed(pg.tp_local_rank())
+    torch.distributed.barrier()
 
     for i, (input_ids, attn_mask) in enumerate(train_dataloader):
-        logits = model(input_ids, attn_mask)
+        colo_input = ColoTensor.from_torch_tensor(input_ids, ColoTensorSpec(pg))
```
feifeibear, line 81: do we really need to convert input to ColoTensor?
What if pass a torch tensor into the colo model?

1SAA: I'm not sure whether this line is necessary.
Since we have a problem in transformation between [dp: 4, tp: 1] and [dp: 1, tp: 4], I think we'd better add this line to ensure that this transformation would not be triggered. Otherwise, `to_replicate` may use PG, [dp: 4, tp: 1] and the tensor wouldn't be replicated when we want to get 4 copy from it.

feifeibear: if the dp degree is consist among the model parameters. I believe we should pass the torch tensor into the model.

feifeibear: Did you test the correctness in the original way?
logits = model(input_ids, attn_mask)

1SAA: > if the dp degree is consist among the model parameters. I believe we should pass the torch tensor into the model.

I don't think so. Acctually, passed data has its own ditribution information. In TP environment, passed data is replicated among TP process group. In DP environment, passed data is just a part of batched data. It is not aprropriate to regard them having no difference as an odinary tensor.

PR #1339, opened 2022-07-19
[checkpoint] use gather_tensor in checkpoint and update its unit test

1SAA:
- only gathers tensors to rank0 to reduce memory usage
- corrected and polished colo_tensor's checkpointing unit test

### DIFF
--- a/colossalai/utils/checkpoint/module_checkpoint.py
+++ b/colossalai/utils/checkpoint/module_checkpoint.py
@@ -1,8 +1,8 @@
 import torch
 import torch.distributed as dist
-from colossalai.tensor import ColoTensor, DistSpecManager
+from colossalai.tensor import ColoTensor
 from colossalai.nn.optimizer import ColossalaiOptimizer
-from copy import copy
+from colossalai.utils.checkpoint.utils import gather_tensor, scatter_tensor
 from typing import Optional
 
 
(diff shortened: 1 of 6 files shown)


colossalai/utils/checkpoint/module_checkpoint.py
```diff
@@ -22,37 +22,52 @@ def save_checkpoint(dire: str,
...
+
+    if rank == 0:
+        # sanity check
+        for k, v in model_state.items():
             if isinstance(v, ColoTensor):
+                assert v.save_ready
                 assert v.is_replicate()
+                delattr(v, 'save_ready')
+        # model saving
+        save_state = {'epoch': epoch, 'model': model_state}
```
feifeibear, line 41: In the next PR. You can merge model and optim in a single file.
```
torch.save({
            'epoch': EPOCH,
            'model_state_dict': net.state_dict(),
            'optimizer_state_dict': optimizer.state_dict(),
            'loss': LOSS,
            }, PATH)
```
like 
https://pytorch.org/tutorials/recipes/recipes/saving_and_loading_a_general_checkpoint.html

colossalai/utils/checkpoint/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    """Make colo_tensor replicated when the rank is 0
+    """
+    if not colo_tensor.is_replicate():
+        pg = colo_tensor.get_process_group()
+        # for the group which contains rank 0
+        if pg.tp_rank_list()[0] == 0:
+            old_dist_spec = colo_tensor.dist_spec
+            colo_tensor.to_replicate_()
+            if dist.get_rank() != 0:
+                colo_tensor.set_dist_spec(old_dist_spec)
```
feifeibear, line 17: This line triggers collective communication. 
Will there be potential blocking if rank 0 is excluded?

tests/test_utils/test_colo_checkpoint.py
```diff
@@ -155,43 +159,33 @@ def _run_checkpoint(model_name, init_spec_func, use_ddp, use_mp_reload, test_sch
...
     if rank == 0:
         remove('./checkpoint')
+    dist.barrier()
 
 
 def run_dist(rank, world_size, port, use_ddp, use_mp_reload, test_scheduler):
     colossalai.launch(config={}, rank=rank, world_size=world_size, host='localhost', port=port, backend='nccl')
     pg = ProcessGroup(tp_degree=world_size)
-    for model_name in ['simple_net', 'bert']:
+    # TODO(haichen) add BERT in the test
```
feifeibear, line 186: Inside a DP group, the input is replicated?

1SAA: Depends on which model is using, we do not unite the standard now.

feifeibear (approved):

PR #1342, opened 2022-07-19
[fx] refactor tracer to trace complete graph

YuliangLiu0306:
Previously, when we call ColoProxy.shape, a torch.Size instance will be returned, instead of another ColoProxy as expected. It may leads to the tracing process temporarily stop until the forward function uses other ColoProxys.
To solve this issue, in this PR, we will create a new ColoProxy for getattr function, then tracing process could get complete graph.


colossalai/fx/tracer/tracer.py
```diff
@@ -358,10 +364,58 @@ def wrapper(*args, **kwargs):
...
             # this is called directly when the inputs do not contain proxy
             # e.g. torch.ones(4) where the input is static
             return target(*args, **kwargs)
 
     return wrapper, target
+
+
+for method in magic_methods:
+
+    def _scope(method):
```
FrankLeeeee, line 383: Add a few lines of comments here would help.

YuliangLiu0306: Ok

colossalai/fx/tracer/meta_patch/patched_function/torch_ops.py
```diff
@@ -24,6 +24,11 @@ def torch_arange(*args, **kwargs):
     return torch.empty((end - start) // step, dtype=dtype, device="meta")
 
 
+@meta_patched_function.register(torch.finfo)
+def torch_finfo(*args, **kwargs):
+    return torch.finfo(*args)
```
FrankLeeeee, line 29: kwargs is not passed to torch.finfo?

YuliangLiu0306: Yes, I will delete the kwargs argument.

colossalai/fx/tracer/_tracer_utils.py
```diff
@@ -29,3 +31,18 @@ def _convert(val):
...
+
+
+def compute_meta_data_for_functions_proxy(target, args, kwargs):
+    args_metas, kwargs_metas = extract_meta(*args, **kwargs)
+
+    # fetch patched function
+    if meta_patched_function.has(target):
+        meta_target = meta_patched_function.get(target)
+    else:
+        meta_target = target
```
FrankLeeeee, line 43: Some functions are indexed by name, e.g. torch builtin add is registered with the name 'add' as it can cover `torch.add`, `torch.Tensor.add` and `object.__add__`, so add this to the if statement.

```python
if meta_patched_function.has(target):
        meta_target = meta_patched_function.get(target)
elif meta_patched_function.has(target.__name__):
        meta_target = meta_patched_function.get(target.__name__)
else:
        meta_target = target
```

colossalai/fx/tracer/meta_patch/patched_function/torch_ops.py
FrankLeeeee: Should it be `torch.finfo(*args, **kwargs)`?

FrankLeeeee:
[test_transform_mlp_pass.py](https://github.com/hpcaitech/ColossalAI/pull/1342/files#diff-fd74a32c8b9b90ab1d119e6657d9359d5eebf7967f2002dfb62776150933388a) has been deleted, resolve the conflict.

YuliangLiu0306:
image

FrankLeeeee (approved):

PR #1361, opened 2022-07-25
[fx] Add torchrec deepfm colotracer compatibility test

Cypher30:
Currently the torchrec.models.deepfm has no compatibility issue with colotracer, and I have done all the testing aligned to other tracer test.
I will add dlrm part later


Cypher30:
Wrong requirements

PR #1374, opened 2022-07-27
[checkpoint] add kwargs for load_state_dict

1SAA:


feifeibear (approved):

colossalai/utils/checkpoint/module_checkpoint.py
```diff
@@ -71,22 +71,23 @@ def save_checkpoint(path: str,
...
 
-def load_checkpoint(path,
+def load_checkpoint(path: str,
                     epoch: int,
                     model: torch.nn.Module,
                     optimizer: Optional[ColossalaiOptimizer] = None,
                     lr_scheduler: torch.optim.lr_scheduler._LRScheduler = None,
+                    load_state_dict_kwargs: Optional[Dict] = None,
                     *args,
                     **kwargs):
```
FrankLeeeee, line 81: It can be quite confusing to the user if there are two kwargs

1SAA: Yeah, there are two functions requiring kwargs.
I think kwargs is conveninent for users.
Should we remove kwargs and add two dict for paramters passing in?

FrankLeeeee: I think it will be better.

1SAA: Done.

PR #1379, opened 2022-07-27
[hotfix] fix some bugs during gpt2 testing

YuliangLiu0306:


FrankLeeeee:
Can you elaborate a bit on what this PR is doing?

YuliangLiu0306:
This PR mainly does two things:
1. Previously, I suppose the args of 'output' node is a tuple or list of nodes in gpt2 model, but it somethings has only one node with the increasing number of partitions. I change it to support more general case.
2. In old version eliminate_unused_output function, out_args contains both nodes and name of nodes, and we need to convert those 'name' elements to node. In this PR, the out_args contains only name of nodes, and convert them to node together.

feifeibear (approved):

PR #1403, opened 2022-08-03
remade clip_grad_norm_fp32() with the new ColoTensor API

CsRic:
split the clip_grad_norm function, adapted with ColoTensor.
only programmed the TP all-reduce part. bypass PP and ZeRO, see comments.


colossalai/utils/common.py
```diff
@@ -161,8 +162,95 @@ def _get_tensor_norm(norm: Union[float, torch.Tensor], move_to_cuda) -> torch.Te
...
+    enable_zero = False
+    total_norm = _norm_tensor_parallel_fp32(parameters, norm_type)
+    if enable_pipline_parallel:
+        total_norm = _norm_pipeline_parallel_fp32(total_norm, norm_type)
+    if enable_zero:
+        total_norm = _norm_data_parallel_fp32(total_norm, norm_type)
+    _clip_grad_fp32(parameters, max_norm, total_norm)
+    return total_norm
+
+def _clip_grad_norm_fp32_old(parameters, max_norm, norm_type=2):
```
ver217, line 253: Don't rename this function.

CsRic: this function is obsolete. I will just delete it.

colossalai/utils/common.py
```diff
@@ -161,8 +162,95 @@ def _get_tensor_norm(norm: Union[float, torch.Tensor], move_to_cuda) -> torch.Te
...
+
+    else:
+        for p in params:
+            if norm_type == 2.0:
+                local_norm = _calc_l2_norm([p.grad.data])**norm_type
+            else:
+                local_norm = _calc_lp([p.grad.data], norm_type)
+            local_norm_cuda = torch.cuda.FloatTensor([float(local_norm)])
+            if p.is_sharded():
+                dist.all_reduce(local_norm_cuda,
```
ver217, line 202: This leads to many unexpected communications. You may use a bucket.

colossalai/utils/common.py
ver217: Old API still need this function. Don't delete it.

colossalai/utils/common.py
CsRic: done

colossalai/utils/common.py
```diff
@@ -1,11 +1,14 @@
...
 # -*- encoding: utf-8 -*-
 import os
+from pprint import pp
 import random
 import socket
 from pathlib import Path
 from typing import Callable, List, Union
 import functools
+
+from sklearn.model_selection import ParameterSampler
```
ver217, line 11: Is it useless?

CsRic: yes. its a mistake

PR #1416, opened 2022-08-09
[tensor] added linear implementation for the new sharding spec

FrankLeeeee:
The previous PR #1405  implemented the sharding spec. This PR implements the linear distributed computation using the new sharding spec API. Currently only the forward part is implemented. When checking the correctness of the gradients, I found that the backward part of the autograd function `reduce_input` is not executed for some reason, thus, the correctness of the gradient is not guaranteed.

I will send a separate PR to fix this problem.

### DIFF
--- a/colossalai/nn/_ops/linear.py
+++ b/colossalai/nn/_ops/linear.py
@@ -4,6 +4,8 @@ from ._utils import GeneralTensor, convert_to_colo_tensor
 from colossalai.tensor.op_wrapper import colo_op_impl
 from ._utils import reduce_input, reduce_grad
 from colossalai.tensor import ComputePattern, ComputeSpec, ColoTensor, ShardSpec, ReplicaSpec, ColoTensorSpec
+from colossalai.tensor.sharding_spec import ShardingSpec
+from copy import deepcopy
 
 
 def colo_linear_1drow(input_tensor: ColoTensor, weight: ColoTensor, bias: Optional[ColoTensor]) -> 'ColoTensor':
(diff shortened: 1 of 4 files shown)


colossalai/nn/_ops/linear.py
```diff
@@ -86,8 +87,63 @@ def colo_linear_imp(input_tensor: GeneralTensor,
...
+        output_partition_dict = {}
+
+        # compute the matmul
+        partial_output = F.linear(input_tensor, weight)
+        ret_tensor = reduce_input(partial_output, pg)
+
+        # add bias
+        if bias is not None:
+            ret_tensor += bias
+    elif str(weight.sharding_spec.sharding_sequence) == '[S1, R]':
```
YuliangLiu0306, line 115: [S0, R] is a valid sharding sequence for weight, like Matmul([R, R], [R, S0]) -> [R, S0]

colossalai/nn/_ops/linear.py
```diff
@@ -86,8 +87,63 @@ def colo_linear_imp(input_tensor: GeneralTensor,
...
+        assert bias is None or str(bias.sharding_spec.sharding_sequence) == '[S1]'
+
+        input_parallel = reduce_grad(input_tensor, pg)
+        ret_tensor = F.linear(input_parallel, weight, bias)
+
+        # the output is a column-sharded tensor
+        last_dim = input_tensor.dim() - 1
+        output_partition_dict = {last_dim: [1]}
```
YuliangLiu0306, line 123: I think this rule is too simple to cover all cases. A counter-example is Matmul([S0, R], [R, S1]) ->  [S0, S1]

colossalai/nn/_ops/linear.py
```diff
@@ -86,8 +87,63 @@ def colo_linear_imp(input_tensor: GeneralTensor,
...
+    # handle column and row parallel linear
+    # by reusing the implementation above
+    if str(weight.sharding_spec.sharding_sequence) == '[R, S1]':
+        assert bias is None or str(bias.sharding_spec.sharding_sequence) == '[R]'
+        # the output is a replica
+        output_partition_dict = {}
+
+        # compute the matmul
+        partial_output = F.linear(input_tensor, weight)
+        ret_tensor = reduce_input(partial_output, pg)
```
YuliangLiu0306, line 110: I think ret_tensor in this case should also have sharding spec, such as Matmul([S0,S1], [S1, R]) -> [S0, R]

colossalai/nn/_ops/linear.py
```diff
@@ -86,8 +87,63 @@ def colo_linear_imp(input_tensor: GeneralTensor,
...
+        assert bias is None or str(bias.sharding_spec.sharding_sequence) == '[S1]'
+
+        input_parallel = reduce_grad(input_tensor, pg)
+        ret_tensor = F.linear(input_parallel, weight, bias)
+
+        # the output is a column-sharded tensor
+        last_dim = input_tensor.dim() - 1
+        output_partition_dict = {last_dim: [1]}
+    else:
+        raise RuntimeError("Different sharding spec is not supported yet")
```
YuliangLiu0306, line 125: [R, R] should be a valid sharding spec for weight.

YuliangLiu0306 (approved):

PR #1418, opened 2022-08-09
[tensor] add shape consistency feature to support auto spec transform

YuliangLiu0306:
This PR construct a ShapeConsistencyManager class to manage the shape consistency behaviour, it mainly does three things:
1. Give a pair of source spec and target spec, it will give you a valid transform path with greedy algorithm.
2. It will cached the transform path of all searched pairs, so most of time when you apply the shape consistency transform, the transform has already cached in manager, it won't cost additional time to search the path again.
3. It will give an apply function to implement the sharding spec to the tensor.

Note: The ShapeConsistencyManager class is just partially completed.

### DIFF
--- a/colossalai/tensor/shape_consistency.py
+++ b/colossalai/tensor/shape_consistency.py
@@ -27,12 +27,16 @@ class ShapeConsistencyManager:
         Therefore, all gather operation just remove the last element in shard list,
         e.g.: 
             all-gather(S01) -> S0
+
+        Argument:
+            target_pair(Tuple[int, List[int]]): The first element is the dimension of tensor to be sharded,
+            and the second element decribes which logical axis will be sharded in that dimension.
         '''
-        pos, shard_list = target_pair
+        _, shard_list = target_pair
         new_shard_list = shard_list[:-1]
         # TODO: compute comm cost
         comm_cost = 0
-        return (pos, new_shard_list), comm_cost
+        return new_shard_list, comm_cost
 
     def _all_to_all_simulator(self, f_target_pair, b_target_pair):
         '''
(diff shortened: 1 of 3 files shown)


colossalai/tensor/shape_consistency.py
```diff
@@ -0,0 +1,306 @@
+import torch
+from colossalai.tensor.sharding_spec import ShardingSpec, _DimSpec
+from enum import Enum
+from copy import deepcopy
+
+
+class CollectiveCommPattern(Enum):
+    ALLGATHER = 'all_gather'
+    ALLTOALL = 'all_to_all'
+    SHARD = 'shard'
```
FrankLeeeee, line 10: This enum is not used anywhere in the code?

colossalai/tensor/shape_consistency.py
```diff
@@ -0,0 +1,306 @@
...
+        We don't allow uncontiguous layout, such as all-gather(S012)->S02 is NOT allowed.
+        Therefore, all gather operation just remove the last element in shard list,
+        e.g.: 
+            all-gather(S01) -> S0
+        '''
+        pos, shard_list = target_pair
+        new_shard_list = shard_list[:-1]
+        # TODO: compute comm cost
+        comm_cost = 0
+        return (pos, new_shard_list), comm_cost
```
FrankLeeeee, line 35: `pos` is obtained from input and directly returned.

colossalai/tensor/shape_consistency.py
```diff
@@ -0,0 +1,306 @@
...
+
+class ShapeConsistencyManager:
+
+    def __init__(self, consistency_option=None):
+        self.consistency_option = consistency_option
+        self.total_communication_cost = 0
+        self.total_transform_steps = 0
+        self.cached_spec_pairs = {}
+
+    def _all_gather_simulator(self, target_pair):
```
FrankLeeeee, line 21: Add docstring for `target_pair`.

colossalai/tensor/shape_consistency.py
YuliangLiu0306: It will be used in apply cached_spec_dict, but this part is WIP.

colossalai/tensor/shape_consistency.py
YuliangLiu0306: I see. I will delete it.

FrankLeeeee (approved):

PR #1421, opened 2022-08-09
Add FreqAwareEmbeddingBag

feifeibear:


feifeibear:
Add FreqAwareEmbeddingBag from https://github.com/zxgx/FreqCacheEmbedding/blob/main/recsys/modules/embeddings/freq_aware_embedding.py#L14

PR #1425, opened 2022-08-09
[fx] modify the calculation of node_size in MetaInfoProp for activation checkpointing usages

super-dainiu:
I believe this change won't affect usage of node.meta['tensor_meta'], but I have adopted the expressions in https://github.com/pytorch/pytorch/blob/f9533560ccb2d789bc78caa360d021f72f5ebc7b/torch/fx/passes/graph_manipulation.py#L82 so as to compute the memory consumptions for activation and model parameters separately. This will make it easy for us to retrieve the activation memory cost directly and correctly.

```python
for node in graph.nodes:
    node_size = getattr(node, 'node_size')    # node_size is a NamedTuple(output_size, param_size)
    print(node_size.output_size)    # will get the memory cost of activation in bytes format
    print(node_size.param_size)     # will get the memory cost of activation in bytes format
```

### DIFF
--- a/colossalai/fx/passes/meta_info_prop.py
+++ b/colossalai/fx/passes/meta_info_prop.py
@@ -22,6 +22,12 @@ class TensorMetadata(NamedTuple):
     # behaviour by appending sharding spec into list.
 
 
+@compatibility(is_backward_compatible=False)
+class node_size(NamedTuple):
+    output_size: int
+    param_size: int
+
+
 def _extract_tensor_metadata(result: torch.Tensor) -> TensorMetadata:
     """
     Extract a TensorMetadata NamedTuple describing `result`.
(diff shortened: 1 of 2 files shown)


colossalai/fx/passes/meta_info_prop.py
```diff
@@ -22,6 +22,12 @@ class TensorMetadata(NamedTuple):
     # behaviour by appending sharding spec into list.
 
 
+@compatibility(is_backward_compatible=False)
+class node_size(NamedTuple):
```
feifeibear, line 26: You should name it as NodeSize. It is a class name!

super-dainiu: I actually adopted this lower-case convention from https://github.com/pytorch/pytorch/blob/master/torch/fx/passes/graph_manipulation.py. Anyway, I removed this class because we won't be using it anymore.

YuliangLiu0306 (approved):

PR #1433, opened 2022-08-10
[fx] add vanilla activation checkpoint search with test on resnet and densenet

super-dainiu:
### What's new?
Now you can use the greedy algorithm from [Training Deep Nets with Sublinear Memory Cost (Chen et al. 2016)](https://arxiv.org/abs/1604.06174) with two lines of code.
```
B = 5 * 1024 * 1024 * 1024  # An approximate memory budget of 5GB
model = resnet18()
input_sample = torch.rand(4, 3, 224, 224)
gm = symbolic_trace(model)
MetaInfoProp(gm).run(input_sample)
gm = chen_greedy(gm, B)     # gm will run with activation checkpoint
```


colossalai/fx/passes/algorithms/ckpt_solver_chen.py
```diff
@@ -0,0 +1,60 @@
+import torch
+from torch.fx import GraphModule
+
+
```
FrankLeeeee, line 4: Add a namespace code  here.
```
__all__ = ['chen_greedy', `chen_sqrtn`]
```

FrankLeeeee (approved):

PR #1436, opened 2022-08-11
[zero] add chunk size searching algorithm for parameters in different groups

1SAA:
A new allocation algorithm for a uniform chunk size for parameters in different dp groups.

### DIFF
--- a/colossalai/gemini/update/search_utils.py
+++ b/colossalai/gemini/update/search_utils.py
@@ -1,0 +1,96 @@
+from typing import Dict, List
+import numpy as np
+import torch.nn as nn
+from colossalai.tensor import ColoParameter
+
+
+def _filter_exlarge_params(model: nn.Module, size_dict: Dict[int, List[int]]) -> None:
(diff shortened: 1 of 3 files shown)


colossalai/gemini/update/search_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+    params_dict = clasify_params(model)
+    config_dict: Dict[int, Dict] = dict()
+
+    size_dict: Dict[int, List[int]] = dict()
+    for key in params_dict:
+        params_list = params_dict[key]
+        size_list = [p.numel() for p in params_list]
+        # let small parameters keep gathered in CUDA all the time
+        tot_size = sum(size_list)
```
ver217, line 67: Don't use abbreviation here.

ver217 (approved):

colossalai/gemini/update/search_utils.py
1SAA: done

PR #1439, opened 2022-08-11
[fx] Use colossalai checkpoint and add offload recognition in codegen

Cypher30:
As our future automatic parallelization might need to offload the checkpoint input for memory saving, I 

1.  Replace the origin torch checkpoint function with colossal.utils.checkpoint, which has the inference for offload option.
2. Add the feature that let codegen module could recognize the "activation_checkpoint" option of node in the graph that is the start point of checkpoint region.

In the test I use `setattr(node, 'activation_offload', True)` to manually annotate the node with offload option, currently the tracer cannot trace the offload behavior, I think we could add this feature in the future. I think it is ok to only allow our automatic parallelization strategy (or manually by user if needed) to annotate the node with the attribute (the model to be traced should not have the offload operation) right now.


tests/test_fx/test_codegen/test_activation_checkpoint_codegen.py
```diff
@@ -13,6 +16,28 @@
...
+        s = socket.socket()
+        try:
+            s.bind((host, port))
+            s.close()
+            return port
+        except socket.error:
+            pass
+        port = port + 1
+
+
```
FrankLeeeee, line 40: import this from `colossalai.utils`.

tests/test_fx/test_codegen/test_activation_checkpoint_codegen.py
```diff
@@ -52,14 +85,24 @@ def test_act_ckpt_codegen():
...
+                setattr(node, 'activation_offload', True)
+
+    # assert checkpoint function will be generated and
+    # the offload option is correct
     code = graph.python_code('self').src
-    assert 'checkpoint_0' in code and 'checkpoint_1' in code
+    assert 'colossalai.utils.checkpoint(checkpoint_0, False, x)' in code and 'colossalai.utils.checkpoint(checkpoint_1, True, x)' in code
+
+    # launch colossalai to make sure we could execute colossalai.utils.checkpoint currectly
+    colossalai.launch(config={}, rank=0, world_size=1, host='localhost', port=free_port(), backend='nccl')
```
FrankLeeeee, line 105: You can move this to the top of this function.

tests/test_fx/test_codegen/test_activation_checkpoint_codegen.py
```diff
@@ -84,13 +132,22 @@ def test_act_ckpt_python_code_torch11():
...
```
FrankLeeeee, line 150: Move this to the top of this function.

colossalai/fx/codegen/activation_checkpoint_codegen.py
```diff
@@ -566,9 +576,13 @@ def emit_node(node: Node):
...
         code = ''.join(body)
         code = '\n'.join('    ' + line for line in code.split('\n'))
+
+        # as we need colossalai.utils.checkpoint, we need to import colossalai
+        # in forward function
         fn_code = f"""
 {wrap_stmts}
 
 def forward({', '.join(orig_args)}){maybe_return_annotation[0]}:
+    import colossalai
```
FrankLeeeee, line 586: `import` should not occur in the forward function. One way is to make sure `colossalai` is imported during tracing activation checkpointing.

colossalai/fx/codegen/activation_checkpoint_codegen.py
```diff
@@ -368,7 +375,10 @@ def emit_node(node: Node):
             for name, value in self.additional_globals():
                 add_global(name, value)
 
+            # as we need colossalai.utils.checkpoint, we need to import colossalai
+            # in forward function
             prologue = self.gen_fn_def(free_vars, maybe_return_annotation[0])
+            prologue = prologue + "\n    import colossalai"
```
FrankLeeeee, line 381: this is the same.

colossalai/fx/codegen/activation_checkpoint_codegen.py
FrankLeeeee: This should be fixed in a separate PR.

FrankLeeeee (approved):

colossalai/fx/codegen/activation_checkpoint_codegen.py
```diff
@@ -99,13 +99,13 @@ def _gen_ckpt_output(output_vars: List[str]) -> str:
...
 
-def _gen_ckpt_usage(label, input_vars, output_vars):
+def _gen_ckpt_usage(label, activation_offload, input_vars, output_vars):
     """
     Generate the checkpoint function call code text
     """
     outputs = ', '.join(output_vars)
     inputs = ', '.join(input_vars)
-    return f'{outputs} = torch.utils.checkpoint.checkpoint(checkpoint_{label}, {inputs})'
+    return f'{outputs} = colossalai.utils.activation_checkpoint.checkpoint(checkpoint_{label}, {activation_offload}, {inputs})'
```
Cypher30, line 108: I have changed those usages from colossalai.utils.checkpoint to colossalai.utils.activation_checkpoint.checkpoint

FrankLeeeee: ok.

tests/test_fx/test_codegen/test_activation_checkpoint_codegen.py
```diff
@@ -84,21 +114,31 @@ def test_act_ckpt_python_code_torch11():
...
+    assert 'colossalai.utils.activation_checkpoint.checkpoint(checkpoint_0, False, x)' in code and \
+    'colossalai.utils.activation_checkpoint.checkpoint(checkpoint_1, True, x)' in code
 
     # recompile and verify the outputs are consistent
     gm = GraphModule(model, graph)
     gm.recompile()
     fx_out = gm(data)
     assert torch.equal(non_fx_out, fx_out)
 
+    gpc.destroy()
```
Cypher30, line 138: I have added those gpc.destroy() functions to destroy the colossalai.launch process, hope it works

PR #1442, opened 2022-08-11
[utils] Impl clip_grad_norm for ColoTensor and ZeroOptimizer

ver217:
```python
from colossalai.utils.common import clip_grad_norm
total_norm = clip_grad_norm(params, 1.0, norm_type=2.0)
```
Grads can be on CUDA and CPU. They can be either fp16 or fp32.


feifeibear (approved):

densechen:
How can I use this function under lightning?  It seems that "on_after_backward()" hook does not work...

Issue #1440, opened 2022-08-11
[BUG]: weight duplicate in ParallelFreqAwareEmbeddingBag


zxgx:
### 🐛 Describe the bug

I suppose [`ParallelFreqAwareEmbeddingBag._weight`](https://github.com/hpcaitech/ColossalAI/blob/3b26516c69a37ec9731e8dd0245436fd4c03120f/colossalai/nn/parallel/layers/cache_embedding/parallel_freq_aware_embedding.py#L63) should be a ColoTensor, otherwise it would be moved to GPU when some Model containing this embedding uses .cuda().
That's why I observed the memory leakage:
![4ZgYZqe5vr](https://user-images.githubusercontent.com/34452939/184090949-d8e15eb5-5d08-496c-9c22-3ee56180a9f1.jpg)

Besides, as I've indicated, this weight is also duplicated in CachedParamMgr in this [Line](https://github.com/hpcaitech/ColossalAI/blob/3b26516c69a37ec9731e8dd0245436fd4c03120f/colossalai/nn/parallel/layers/cache_embedding/cache_mgr.py#L33)

In a word, I believe another design of the weight shared between ParallelFreqAwareEmbeddingBag and CachedParamMgr  is in need, and I'm working on it.

PR #1448, opened 2022-08-12
reorganize the inheritance struct of FreqCacheEmbedding

zxgx:
1. Remove the redundant weight in FreqCacheEmbedding to resolve this issue: https://github.com/hpcaitech/ColossalAI/issues/1440
2. reorganize the inheritance structure of FreqCacheEmbedding to remove redundant code in ParallelFreqCacheEmbedding. 
3. I also leave an interface in CacheParamMgr, so that FreqCacheEmbedding is compatible with vanilla PyTorch EmbeddingBag. But this implementation should be delayed.

Uq3tUGyuNX

### DIFF
--- a/colossalai/nn/parallel/layers/cache_embedding/cache_mgr.py
+++ b/colossalai/nn/parallel/layers/cache_embedding/cache_mgr.py
@@ -23,14 +23,26 @@ class CachedParamMgr(torch.nn.Module):
 
         self.elem_size_in_byte = weight.element_size()
 
+        # weight configure
+        if cuda_row_num > 0:
+            # Enable cache
+            self._init_with_cache(weight)
+        else:
+            # Disable cache
+            self._init_without_cache()
+
+        self.num_hits_history = []
+        self.num_miss_history = []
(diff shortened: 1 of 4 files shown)


feifeibear (approved):

colossalai/nn/parallel/layers/cache_embedding/cache_mgr.py
```diff
@@ -23,14 +23,26 @@ def __init__(self, weight: torch.Tensor, cuda_row_num: int = 0, buffer_size: int
 
         self.elem_size_in_byte = weight.element_size()
 
+        # weight configure
+        if cuda_row_num > 0:
+            # Enable cache
+            self._init_with_cache(weight)
+        else:
+            # Disable cache
+            self._init_without_cache()
```
feifeibear, line 32: 这两个接口合成一个吧。
self._init_with_cache self._init_without_cache

zxgx: Resolved

PR #1453, opened 2022-08-14
[tensor] support runtime ShardingSpec apply

YuliangLiu0306:
In #1435 PR，ShapeConsistencyManager could generate transform path and communication action sequence using comm simulator, but the runtime ShardingSpec apply function is still NOT supported. This PR mainly do the following things:
1. To support DOT op on sharding operands, all_reduce operation is supported by CommSpec.
2. Implement covert_spec_to_action to generate appropriate action based on CommSpec.
3. Encapsulating the shape_consistency_apply interface to support runtime ShardingSpec transform.


colossalai/device/device_mesh.py
```diff
@@ -18,9 +19,18 @@ class DeviceMesh:
             communication cost (default: None)
         mesh_beta (List[float], optional): coefficients used for computing
             communication cost (default: None)
+        init_process_group_during_init (bool, optional): initialize logical process group
```
FrankLeeeee, line 22: This variable name is kind of  tongue-twisting, I think `init_process_group` will be enough.

colossalai/tensor/shape_consistency.py
```diff
@@ -439,3 +513,67 @@ def shape_consistency(self, source_spec, target_spec):
             total_steps += 1
 
         raise RuntimeError(f"Could not find a valid transform path with in {MAX_TRANSFORM_STEPS} steps.")
+
+    def shape_consistency_apply(self, tensor_with_sharding_spec, target_spec):
```
FrankLeeeee, line 517: keep this succinct, `apply` should be enough.

colossalai/device/device_mesh.py
YuliangLiu0306: hhhh，I will change it.

FrankLeeeee (approved):

PR #1451, opened 2022-08-12
[fx] fix test and algorithm bugs in activation checkpointing.

super-dainiu:
### What's new?
I regretted using `torch.randn(2, 3, 224, 224)` previously #1446 for testing because this consumes too much time on CI.
Also, I made some modifications to the search algorithm (mostly conditions for annotations) to avoid crashes in `ActivationCheckpointCodeGen`. 
### What's wrong
However, I did not figure out why tracing on `densenet121` got an error.
```bash
Traceback (most recent call last):
  File "/home/ColossalAI/tests/test_fx/test_ckpt_solvers/test_ckpt_torchvision.py", line 104, in <module>
    test_ckpt_solver()
  File "/home/ColossalAI/tests/test_fx/test_ckpt_solvers/test_ckpt_torchvision.py", line 73, in test_ckpt_solver
    check_backward_consistency(m, gm, solver, model_cls)
  File "/home/ColossalAI/tests/test_fx/test_ckpt_solvers/test_ckpt_torchvision.py", line 46, in check_backward_consistency
    loss.backward()
  File "/home/.local/lib/python3.9/site-packages/torch/_tensor.py", line 396, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
  File "/home/.local/lib/python3.9/site-packages/torch/autograd/__init__.py", line 173, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "/home/.local/lib/python3.9/site-packages/torch/autograd/function.py", line 253, in apply
    return user_fn(self, *args)
  File "/home/ColossalAI/colossalai/utils/activation_checkpoint.py", line 114, in backward
    outputs = ctx.run_function(*detached_inputs)
  File "<eval_with_key>.3", line 73, in checkpoint_2
  File "/home/.local/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/.local/lib/python3.9/site-packages/torch/nn/modules/activation.py", line 98, in forward
    return F.relu(input, inplace=self.inplace)
  File "/home/.local/lib/python3.9/site-packages/torch/nn/functional.py", line 1455, in relu
    result = torch.relu_(input)
RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.
```
The generated `nn.Module` is as follows. Problems occurred in checkpoint_2
```
import torch
from torch.nn import *
[...]

PR #1460, opened 2022-08-16
[utils] Add use_reetrant=False in utils.activation_checkpoint

Cypher30:
We encountered the situation that if the first operation in a checkpoint function is an in-place operation, it will raise error as in original checkpoint process, it calls `run_function` with the `detached_input`, which will be viewed as a leaf node with `requires_grad=True`, the autograd itself will not allow this thing to happen. We check the torch itself has the option to set use_reentrant=False that could address this problem, using the `torch.autograd.graph.saved_tensors_hooks` to avoid calling the re-computation with `detached_input`. So I add this feature inside our colossalai checkpoint, and add our activation offload process for `use_reetrant=False` case as the original torch checkpoint doesn't provide this. I also modify the activation_checkpoint test which fail in previous PR [[test] recovered activation checkpointig test #1459](https://github.com/hpcaitech/ColossalAI/pull/1459)


tests/test_utils/test_activation_checkpointing.py
```diff
@@ -62,3 +61,54 @@ def test_activation_checkpointing(cpu_offload):
     # other tests will fail if running together with this test
     # as other tests can't overwrite the seed set by this test
     reset_seeds()
+
+
+@pytest.mark.gpu
+@pytest.mark.parametrize("cpu_offload", [True, False])
+def test_activation_checkpointing_reentrant_False(cpu_offload):
+
```
FrankLeeeee, line 69: Add a `reset_seed` at the start of the function so that other tests will not affect this one.

FrankLeeeee: This applies to the test function above as well.

Cypher30: oooo! I will modify it

tests/test_utils/test_activation_checkpointing.py
```diff
@@ -62,3 +61,54 @@ def test_activation_checkpointing(cpu_offload):
     # other tests will fail if running together with this test
     # as other tests can't overwrite the seed set by this test
     reset_seeds()
+
+
+@pytest.mark.gpu
+@pytest.mark.parametrize("cpu_offload", [True, False])
```
FrankLeeeee, line 67: If these two test functions only differ by the `use_reentrant` variable, you can just add `parameterize` here instead of creating a duplicated function.

Cypher30: okay I will modify this

colossalai/utils/activation_checkpoint.py
```diff
@@ -136,14 +138,122 @@ def backward(ctx, *args):
         return (None, None) + grads
 
 
-def checkpoint(function, activation_offload, *args):
+def checkpoint(function, activation_offload, *args, use_reentrant: bool = True):
```
super-dainiu, line 141: You should test on densenet121 where bugs were reported in #1451.

Cypher30: I have done some new tests on use_reentrant=False to test if it could solve the problem in report in #1451, I ran the test in #1451 locally and it passed, you could do some final check on the code. I will change the codegen after this PR is merged

super-dainiu: I will later add more logic for ckpt_solver. Approve~

super-dainiu (approved):
Approve

PR #1463, opened 2022-08-16
[fx] Add use_reentrant=False to checkpoint in codegen

Cypher30:
The previous PR #1460 added `use_reentrant=False` feature in checkpoint, to address the problems we encountered in activation searching strategy in #1451. In this PR, I bring this feature to our ColoTracer codegen process. Once the activation_codegen detects that one checkpoint's input will be used in an in-place operation, it will use `use_reentrant=False` option when calling the `colossalai.utils.activation_checkpoint.checkpoint`, otherwise it will use `use_reentrant=True` which consists with the original checkpoint function.


FrankLeeeee (approved):

super-dainiu:
I suggest you should test on torchvision models instead of customized models.

Cypher30:
> I suggest you should test on torchvision models instead of customized models.

I think we could add some compatibility test in the future, I think the customized model could be short and efficiently covers the situation we might encounter

PR #1471, opened 2022-08-19
[autoparallel] introduced baseclass for op handler and reduced code redundancy

FrankLeeeee:
This PR introduces the base class for the operator handler for better API consistency. The three `generate_sharding_sepc` API is merged into one to remove redundancy as well.

### DIFF
--- a/colossalai/auto_parallel/solver/dot_handler.py
+++ b/colossalai/auto_parallel/solver/dot_handler.py
@@ -3,7 +3,7 @@ from .operator_handler import OperatorHanlder
 
 class DotHandler(OperatorHanlder):
     """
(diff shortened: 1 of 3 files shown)


colossalai/auto_parallel/solver/dot_handler.py
```diff
@@ -0,0 +1,12 @@
+from .operator_handler import OperatorHanlder
+
+
+class DotHandler(OperatorHanlder):
+    """
+    A OperatorHandler which deals with the sharding strategies of linear matrix multiplication.\
```
YuliangLiu0306, line 6: Delete the random slash at the end of this line.

FrankLeeeee: done.

PR #1481, opened 2022-08-23
[autoparallel] add cost graph class

YuliangLiu0306:
A graph data structure to simplify the edge cost graph. It has two main functions:
    1. To feed the quadratic resharding costs into solver, we need to linearize it. We build edge_cost in
    CostGraph, and it stored every combinations of strategies for a src-dst node pair in an 1D list.
    2. To reduce the searching space, we merge computationally-trivial operators, such as 
    element-wise operators, transpose, and reduction, into their following nodes. The merging infomation will
    be given by the StrategiesVector depending on the type of target node and following nodes.

Additional，some previous bugs has been fixed in this PR:
1. misspelling of variable name
2. fix bugs in handler tests

### DIFF
--- a/colossalai/auto_parallel/solver/cost_graph.py
+++ b/colossalai/auto_parallel/solver/cost_graph.py
@@ -1,4 +1,6 @@
 from colossalai.auto_parallel.solver.sharding_strategy import ShardingStrategy, StrategiesVector
+from typing import List
+from torch.fx.node import Node
 
 
 class CostGraph:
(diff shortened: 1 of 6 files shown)


FrankLeeeee (approved):

FrankLeeeee (approved):

FrankLeeeee:
This PR can be merged if the unit test is ok.

YuliangLiu0306:
image

PR #1486, opened 2022-08-24
[pipeline/rpc] implement distributed optimizer | test with assert_close

LSTM-Kirigaya:
1. support distributed optimizer
2. use assert_close to test correctness of optimizer


tests/test_pipeline/test_cuda_rpc_optimizer.py
```diff
@@ -0,0 +1,153 @@
...
+from colossalai.pipeline.rpc.PipelineBase import FillDrainPipelineEngine, OneFOneBPipelineEngine
+from colossalai.testing import assert_close
+
+
+def color_debug(text, prefix=' ', color='blue'):
+    color = color.upper()
+    print(getattr(Back, color), prefix, Style.RESET_ALL, text)
+
+
+class TestModel(nn.Module):
```
YuliangLiu0306, line 21: For the maintainace, I think you could integrate this and some redundant codes below into a file named 'rpc_test_utils.py', and reuse them in rpc tests.

PR #1487, opened 2022-08-24
[autoparallel] add more sharding strategies to conv

YuliangLiu0306:
In previous PR, I missed some cirtical sharding strategies for convolution operation and add them in this PR:
1. the sharding through more than one axis of device mesh on the same dimension, such as S01 on batch dimension or channel_in dimension.
2. The sharding strategy like SR x RR -> SR, this is a typical data parallel strategy. 
3. The sharding strategy like RS x SR -> RR, this is a kind of typical 1D tensor parallel strategies for convolution.

Note: Our device mesh do not have an 1D flatten device mesh, so the communication cost of 'RS01 x S01R -> RR' is temprorally set to 0. It will be fixed after implementation of 1D flatten device mesh.


FrankLeeeee:
The PR for 1D device mesh has been approved, please fix the communication cost for 1D mesh in this PR.

YuliangLiu0306:
> The PR for 1D device mesh has been approved, please fix the communication cost for 1D mesh in this PR.

Done

FrankLeeeee (approved):

PR #1489, opened 2022-08-24
[fx] fixed adapative pooling size concatenation error

FrankLeeeee:
Fixed the shape concatenation error in adapative pooling. This allows the torchvision ResNet to be traced successfully with meta args.

A sample test script is given below:

```python
import torch
import torchvision.models as tm
from colossalai.fx import ColoTracer

torch_model = tm.resnet152()
input = torch.rand(32, 3, 224, 224, device="meta")
tracer = ColoTracer()
graph = tracer.trace(torch_model, meta_args={"x": input})
```


Cypher30 (approved):
Great!

PR #1495, opened 2022-08-25
[fx] add more op patches for profiler and error message for unsupported ops.

super-dainiu:
### What's new?
Now our profiler can profile any model from torchvision library. I also added error message and guide for unsupported ops.
```python
>> model = tm.efficientnet_b0()
>> data = torch.rand(1, 3, 224, 224, device='meta')
>> gm = symbolic_trace(model)
>> MetaInfoProp(gm).run(data)

AssertionError: 
Colossal-AI hasn't supported profiling for <class 'torch.nn.modules.activation.SiLU'>, you might manually patch it with the following code.

from colossalai.fx.profiler import meta_profiler_module

@meta_profiler_module.register(YOUR_MODULE)
def profile_YOUR_MODULE(self: torch.nn.Module, input: torch.Tensor) -> Tuple[int, int]:
    flops = ...
    macs = ...
    return flops, macs
```
### Concerns
This profiler is not able to detect temporary memory cost within one node. For example, `nn.Conv2d()` has extra memory reserved for the allocator, and the amount varies according to different types of devices (https://dl.acm.org/doi/10.1145/3368089.3417050). Also, I did not consider the extra memory cost when computing `nn.MultiheadAttention`. Some problems can be solved by refractoring, but we have to withstand the inaccuracies.


Cypher30 (approved):
Great Work!

PR #1480, opened 2022-08-23
[fx] add profiler for fx nodes.

super-dainiu:
### What's new?
After patching all possible ops, we can now profile the memory cost and FLOPs with lines of code. We only support the original `torch.nn.functional` and `torch.nn`, but it is not too challenging to profile your own model using `MetaInfoProp`.

``` python
import torch
from colossalai.fx.profiler import profile_function, profile_module


input = torch.rand(100, 100, 100, 100, device='meta')
func = torch.nn.functional.relu
output, profile = profile_function(func)(input, inplace=False)
print(f"Profiling function {func},")
print(f"Param size: {profile.param / 1024**2:.3f} MB, Activation size: {profile.activation / 1024**2:.3f} MB, {profile.flops} FLOPs, {profile.macs} MACs")

output, profile = profile_function(func)(input, inplace=True)
print(f"Profiling function {func},")
print(f"Param size: {profile.param / 1024**2:.3f} MB, Activation size: {profile.activation / 1024**2:.3f} MB, {profile.flops} FLOPs, {profile.macs} MACs")

input = torch.rand(4, 3, 224, 224, device='meta')
mod = torch.nn.Conv2d(3, 128, 3)
output, profile = profile_module(mod)(input)
print(f"Profiling function {mod},")
print(f"Param size: {profile.param / 1024**2:.3f} MB, Activation size: {profile.activation / 1024**2:.3f} MB, {profile.flops} FLOPs, {profile.macs} MACs")

===============================================================================
Result:
Profiling function <function relu at 0x7f3b6f8ead30>,
Param size: 0.000 MB, Activation size: 381.470 MB, 100000000 FLOPs, 0 MACs
Profiling function <function relu at 0x7f3b6f8ead30>,
Param size: 0.000 MB, Activation size: 0.000 MB, 100000000 FLOPs, 0 MACs
Profiling function Conv2d(3, 128, kernel_size=(3, 3), stride=(1, 1)),
Param size: 0.014 MB, Activation size: 96.258 MB, 1387837440 FLOPs, 681302016 MACs
===============================================================================
```

Also using `MetaInfoProp`, we can trace the model using option `device='meta'` solely and get all the required results.
```python
from typing import Tuple
import torch
import torch.nn as nn
import torch.nn.functional as F
import torchvision.models as tm
from torch.fx import symbolic_trace
import torch.fx
from torch.optim import Adam
from torch.nn import CrossEntropyLoss
from colossalai.fx.passes.meta_info_prop import MetaInfoProp


[...]

PR #1496, opened 2022-08-25
[fx] Add activation checkpoint solver rotor

Cypher30:
# What's New?

In this PR, we introduce rotor algorithm to our activation checkpoint solver lists. By the great effort in #1480 , we are able to estimate the metrics we need to use rotor algorithm. Down below is a quick example to try this new feature

```python
import torch
import torch.nn as nn
from colossalai.fx import ColoTracer, ColoGraphModule
from colossalai.fx.passes.algorithms import linearize, solver_rotor
from torch.fx import GraphModule
from torchvision.models import resnet18
from colossalai.fx.codegen import ActivationCheckpointCodeGen

model = torch_resnet18()
input = torch.rand(32, 3, 224, 224, device="meta")
tracer = ColoTracer()
graph = tracer.trace(torch_model, meta_args={"x": input})
graph.set_codegen(ActivationCheckpointCodeGen())
gm = ColoGraphModule(model, graph)
gm = solver_rotor(gm, input, mem_limit=600 * 1024 * 1024)
gm.to_folder("./fx")
```

I also add the unit test for it in `test_ckpt_torchvion.py`, currently on our CI we will skip this test, as we haven't implemented our `ColoGraphModule` in torch11, I will attach the test results on torch12 later.


Cypher30:
Screen Shot 2022-08-25 at 20 04 51
Above the test result in `test_fx` with torch12

colossalai/fx/passes/algorithms/utils.py
```diff
@@ -0,0 +1,229 @@
...
+        result = Sequence(Function("Strip", self.function.name, *self.function.args, startOfFwdEnableChain))
+        for i in range(lastIndex + 1):
+            result.insert(ops[i])
+        result.insert(Loss())
+        for i in range(chainLength, startOfFwdEnableChain - 1, -1):
+            position = endOfFirstPhase + 1 + (chainLength - i)
+            assert type(ops[position]) is Backward
+            assert ops[position].index == i
+        for i in range(endOfFirstPhase + 1 + 1 + chainLength - startOfFwdEnableChain, len(ops)):
+            result.insert(ops[i])
```
super-dainiu, line 228: Why only this function has so many CamelCase names?

colossalai/fx/passes/algorithms/linearize.py
```diff
@@ -0,0 +1,89 @@
...
+    # If the output node use multiple nodes, there might be an
+    # empty node after the output node
+    if len(node_dict[node_idx]) == 0:
+        node_dict.pop[node_idx]
+        node_idx -= 1
+
+    # pop the last two nodes
+    node_dict.pop(0)
+    node_dict.pop(node_idx)
+    return node_dict
```
super-dainiu, line 89: I believe the following code works the same.
``` python
def linearize(gm: GraphModule) -> List[List[Node, ...]]:
    """
    In most existing frameworks of activation checkpoint, the forward graph is assumed to be linearized.
    """

    def is_sink():
        """
        If we can free all memories when executing a certain node, it is a sink.
        """
        return not sum((v for k, v in deps.items()))

    deps = {}
    linearized_nodes = []
    region = []
    for n in gm.graph.nodes:
        for n_par in n._input_nodes:
            deps[n_par] -= 1    # free memory and dependencies
        region.append(n)    # append the node to this region

        # we can call an end to the region if the current node is the only living node in memory without any dependencies
        if is_sink():
            linearized_nodes.append(region)
            region = []
        deps[n] = len(n.users)    # add dependencies for future executions
    return linearized_nodes
```
The `linearized_nodes: List[List[Node, ...]]` will contain regions of nodes like `[[node_0, node_1, node_2], [node_3, node_4], ...]`.

colossalai/fx/passes/algorithms/utils.py
```diff
@@ -0,0 +1,229 @@
+class Chain:
+
+    def __init__(self, fw, bw, cw, cbw, ftmp, btmp, check=True):
+        self.fweigth = fw
```
super-dainiu, line 4: `fweigth` not `fweight`? Is that typo from rotor?

colossalai/fx/passes/algorithms/linearize.py
Cypher30: Great idea, but currently my code maintains some flexibility if we need smaller structure of node, check the annotated region line 62-70, in your code above, the below case will be considered as two nodes:
              -n3-
n0 - n1 - n2 - n4
but if we activate line 62-70, n1 could be seen as a node too, I think currently I will maintain my linearize structure, as the `solver_rotor` was built with the `node_dict`, I will consider your simplified code in next PR.

colossalai/fx/passes/algorithms/utils.py
Cypher30: I will fix this and the above funny variable names

colossalai/fx/passes/algorithms/linearize.py
Cypher30: Great idea, I think it is much better than my code, but currently I deactivate line 62-70, so there will be less nodes in our linearized graph, which could might be better for our solver right now. I will consider your simplified structure later.

super-dainiu:
Approved.

super-dainiu (approved):

PR #1510, opened 2022-08-26
[fx] Fix activation codegen dealing with checkpointing first op

Cypher30:
# What's New

In our activation checkpoint solver, it might checkpoint the first op in the graph, in this PR I enable the detection of such situation in our codegen. If the graph needs to checkpoint the first op, we should set `use_reentrant=False` to make sure the backward process will be correct.


colossalai/fx/codegen/activation_checkpoint_codegen.py
```diff
@@ -179,6 +182,10 @@ def emit_code_with_activation_checkpoint(body, ckpt_func, nodes, emit_node_func,
                                 if "inplace" in user.kwargs:
                                     use_reentrant = not user.kwargs["inplace"]
 
+            # if all the inputs are leaf node, we need to set use_reentrat = False
```
super-dainiu, line 185: typo

tests/test_fx/test_codegen/test_activation_checkpoint_codegen.py
```diff
@@ -49,16 +49,20 @@ def __init__(self):
         self.relu = relu()
         self.linear2 = torch.nn.Linear(4, 4)
 
+    def ckpt2(self, x):
+        return F.relu(x, inplace=True)
+
+    def ckpt3(self, x):
+        return self.linear2(x)
+
     def forward(self, x):
```
super-dainiu, line 58: Shall we try the multiple inputs test here? For example,
```python
def forward(self, x1, x2):
      y1 = checkpoint(self.layer1, x1)
      y2 = checkpoint(self.layer2, x2, y1)
      return y1 + y2
```

Cypher30: Yeah, I think it's better, I will add it later

Cypher30: I've added tests as you suggested

colossalai/fx/codegen/activation_checkpoint_codegen.py
Cypher30: fixed

PR #1515, opened 2022-08-29
[fx] hack __torch_dispatch__ for meta tensor and autograd.

super-dainiu:
### What's new?
When I tried to run autograd with meta tensor input on `vit_b_16`, I discovered that some aten ops are not registered for meta backend. So following the suggestions in [Function to automatically calculate Conv shape · Issue #79512 · pytorch/pytorch · GitHub](https://github.com/pytorch/pytorch/issues/79512), I tried to patch native_layer_norm.default for meta backend.

```python
@register_meta(aten.native_layer_norm.default)
def meta_ln(
    input: torch.Tensor, 
    normalized_shape, weight, bias, eps
):
    n_input = input.size(1)

    output = torch.empty_like(input)
    running_mean = torch.empty((n_input), device='meta')
    running_var = torch.empty((n_input), device='meta')
    return output, running_mean, running_var

@register_meta(aten.native_layer_norm_backward.default)
def meta_ln_backward(
    dY: torch.Tensor,
    input: torch.Tensor, 
    normalized_shape, mean, rstd, weight, bias, grad_input_mask
):
    dX = torch.empty_like(input)
    dgamma = torch.empty_like(weight)
    dbeta = torch.empty_like(bias)
    return dX, dgamma, dbeta
```
However, even if patching is successful, the autograd dispatcher refuses to use my patched op for meta backend.

``` bash  
RuntimeError: 0 INTERNAL ASSERT FAILED at "../aten/src/ATen/core/boxing/KernelFunction.cpp":23, please report a bug to PyTorch. aten::native_layer_norm has kernels registered to both CompositeImplicitAutograd and a backend mapped to AutogradOther. This makes the backend kernel unreachable; the dispatcher will always prefer the CompositeImplicitAutograd lowering (see Note [Ambiguity in AutogradOther kernel]). If you want to override CompositeImplicitAutograd, please open an issue to request a dedicated Autograd dispatch key for the backend.   
If you only want to run inference instead of training, add `c10::InferenceMode mode;` before model.forward(). Note this guard is only available in C++ but not Python at present.  
```

So as discussed in [CompositeImplicitAutograd operators should not perform operations that do not dispatch · Issue #61669 · pytorch/pytorch · GitHub](https://github.com/pytorch/pytorch/issues/61669), failing due to CompositeImplicitAutograd is inevitable for PyTorch version 1.12.0 and below. I somehow managed to develop another version of autograd with meta tensor.

``` python 
class MetaTensor(torch.Tensor):

    elem: torch.Tensor
 
    __slots__ = ['elem']
 
    @staticmethod
    def __new__(cls, elem):
        r = torch.Tensor._make_wrapper_subclass(
            cls, elem.size(),
            strides=elem.stride(), storage_offset=elem.storage_offset(),
            dtype=elem.dtype, layout=elem.layout,
            device='cpu', requires_grad=elem.requires_grad
        )    # deceive the frontend for aten selections
        r.elem = elem
        return r

    @ classmethod
    def __torch_dispatch__(cls, func, types, args=(), kwargs=None):
        def unwrap(x):
            return x.elem.to('meta') if isinstance(x, MetaTensor) else x
                
        args = tree_map(unwrap, args)
        kwargs = tree_map(unwrap, kwargs)
        out = func(*args, **kwargs)
        
        def wrap(x):
            return MetaTensor(x) if isinstance(x, torch.Tensor) else x
           
        return tree_map(wrap, out)

```
Since previous works of the PyTorch team have supported aten ops on meta backend, we can simply hack the autograd dispatcher, deceiving it that we are running on CPU. In this way, the dispatcher will not use CompositeImplicitAutograd anymore, and our patched ops can be used for meta backend. So now we can do forward and backward with meta tensor only, and trace a large model with batch_size=1e10 in milliseconds.

```python 

model = vit_b_16()
data = MetaTensor(torch.rand(int(1e10), 3, 224, 224, device='meta'))
model.to('meta')(data).sum().backward()

```

With this amazing `__torch_dispatch__`, I replaced the previously patched version of tracing into **REALLL** meta tracing.

### Concerns
Indeed, `__torch_dispatch__` is not compatible with PyTorch 1.11.0 and below.


colossalai/fx/profiler/__init__.py
```diff
@@ -1,3 +1,8 @@
+try:
+    from ._meta_registrations import *
+except:
+    print(f'_meta_registrations seems to be incompatible with PyTorch {torch.__version__}.')
```
super-dainiu, line 4: Please let me check the compatibility on CI.

colossalai/fx/passes/meta_info_prop.py
```diff
@@ -103,7 +102,7 @@ def extract_tensor_meta(obj):
             else:
                 return TensorMetadata(None, None, False, None, 0, False)
 
-        meta = map_aggregate(result, extract_tensor_meta)
+        meta = tree_map(extract_tensor_meta, result)
```
super-dainiu, line 105: I replace all map_aggregate with tree_map because it is not native for PyTorch development.

Cypher30 (commented):
Great work! As we have debated, you could add the MetaTensor to proxy and try to record the atens inside MetaTensor data structure~

Cypher30 (approved):
We could pass this PR first~

FrankLeeeee (changes requested):
You should fix the broken unit tests before merging this PR.

colossalai/fx/profiler/meta_tensor.py
```diff
@@ -0,0 +1,46 @@
...
+from torch.utils._pytree import tree_map, tree_flatten
+
+class MetaTensor(torch.Tensor):
+    """
+    A wrapping tensor that hacks `torch.autograd` without patching more `torch.ops.aten` ops.
+    """
+
+    _tensor: torch.Tensor
+ 
+    __slots__ = ['elem']
```
FrankLeeeee, line 11: what is this for?

colossalai/fx/profiler/__init__.py
```diff
@@ -1,3 +1,9 @@
+try:
+    from ._meta_registrations import *
+except:
+    import torch
+    print(f'_meta_registrations seems to be incompatible with PyTorch {torch.__version__}.')
+from .meta_tensor import *
 from .registry import *
 from .profiler_function import *
 from .profiler_module import *
```
FrankLeeeee, line 9: Do not use `*` in import as it might mess up the import namespace. You should follow the steps below.
1. give concrete name for import, i.e. `from a import b` instead of `from a import *`
2. add `__all__` for namespace constraints, e.g. `__all__ = ['b']`

colossalai/fx/profiler/_meta_registrations.py
```diff
@@ -0,0 +1,216 @@
...
+
+        tree_map(add_func, op)
+        return f
+
+    return wrapper
+
+
+# https://github.com/pytorch/pytorch/pull/79834
+@register_meta(aten.convolution.default)
+def meta_conv(
```
FrankLeeeee, line 36: Do test the correctness of these patched meta ops.

colossalai/fx/profiler/meta_tensor.py
```diff
@@ -0,0 +1,46 @@
+import torch
+from torch.utils._pytree import tree_map, tree_flatten
+
+class MetaTensor(torch.Tensor):
```
FrankLeeeee, line 4: This might require some refactoring afterwards, either integrate it inside `ColoTensor` or put it directly under `colossalai.fx.tensor`. I can accept as it is now but we can discuss this later.

colossalai/fx/profiler/meta_tensor.py
FrankLeeeee: I mean `__slots__` here.

super-dainiu: Oops I forgot to update the attribute name to `._tensor`

colossalai/fx/profiler/meta_tensor.py
super-dainiu: I would consider MetaTensor as an experimental thing. After I develop all possible features, I may refactor this in accordance with its usage and hierarchy.

colossalai/fx/profiler/_meta_registrations.py
super-dainiu: Do you mean shape consistency? I may come up with another PR doing this.

FrankLeeeee: Sure.

colossalai/fx/profiler/meta_tensor.py
FrankLeeeee: Sure.

colossalai/fx/profiler/meta_tensor.py
FrankLeeeee: what is it for?

super-dainiu: The special attribute [__slots__](https://docs.python.org/3/reference/datamodel.html#slots) allows you to explicitly state which instance attributes you expect your object instances to have, with the expected results:

1. faster attribute access.
2. space savings in memory.

PR #1516, opened 2022-08-29
[autoparallel] added liveness analysis

FrankLeeeee:
This PR Added liveness analysis based on the computation graph. The following context clarifies what is included in this PR.

## What is a live variable

A live variable is a variable whose lifecycle is still in effect (not terminated yet). In other words, if there is a node in a graph 
 and it will not be visited again during traversal, it is no longer a live variable. In the context of deep learning, a live variable can be an input, activation or a weight parameter.

## What is a life cycle stage

In this PR, we travel the computation graph node by node. Each node is transformed into one line of code and thus we deem each node as a stage. Information such as live variables and memory cost is stored with respect to each stage.

```
c = a + b  # node 0 -> stage 0
e = c + d  # node 1 -> stage 1
```

## Live variables for a stage

In a stage (node), what are the live variables produced/destroyed? In the current implementation, we only care about the forward graph, thus, no variable is destroyed. New live variables are added in the following cases:
1. this node produces a new output
2. this node is a `call_module` node and has some parameters which are not used in computation before
3. this node receives an input which has not been recorded before

A stage may also produce an aliased live variable, meaning two live variables refer to the same underlying memory space. This occurs when an operator is an in-place operator such as in-place relu. In this case, we keep both variables in the `all_live_variables` list and only keep one in the `unique_live_variable` list. Only one variable will be counted towards the memory cost of the current stage.

### DIFF
--- a/colossalai/auto_parallel/solver/graph_analysis.py
+++ b/colossalai/auto_parallel/solver/graph_analysis.py
@@ -1,0 +1,210 @@
+from operator import contains
+from xml.dom import NotFoundErr
+import torch
+from dataclasses import dataclass
+from torch.fx.node import Node
+from torch.fx.graph import Graph
+from torch.fx.graph_module import GraphModule
+from collections import OrderedDict as ODict
+from typing import List, OrderedDict, Type, Union, Any
+from colossalai.fx.passes.utils import get_node_module
+
+__all__ = ['LiveVariable', 'LiveVariableVector', 'LiveStage', 'GraphAnalyser', 'compute_tensor_size']
+
+
+@dataclass
+class LiveVariable:
+    """
+    LiveVariable is a data structure to store the meta information of a variable for liveness analysis.
+    """
+    name: str
+    meta: Union[Any, List[Any]]
+    is_inplace: bool
+
+
+class LiveVariableVector(list):
+    """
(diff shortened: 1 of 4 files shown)


colossalai/auto_parallel/solver/graph_analysis.py
```diff
@@ -0,0 +1,210 @@
...
+                    live_var_from_arg = LiveVariable(name=arg_name, meta=meta, is_inplace=False)
+                    all_live_variables.append(live_var_from_arg)
+                    checked_variables.append(live_var_from_arg)
+                    unique_live_vars.append(live_var_from_arg)
+
+            # TODO: add the logic to remove live variables
+            # this should be completed if we are able to trace the backward compute graph
+
+            ###########################
+            # compute the memory cost #
```
YuliangLiu0306, line 177: This memory cost without sharding strategy info won't be used in solver, and it is hard to enumerate the memory cost of all possible strategy combinations. So I think we could just leave the memory cost computing to solver as memory constraint.

FrankLeeeee: Sure.

FrankLeeeee: I have removed the code related to memory cost.

YuliangLiu0306 (approved):

PR #1527, opened 2022-08-31
[fx] add test for meta tensor.

super-dainiu:
I added test for meta tensor about its shape, dtype, and stride. Also, I checked the correctness of back-propagation.
image


Cypher30 (approved):
Great Work! In the future if we are able to retrieve the actual code with both forward and backward (and if this feature is needed by us), we could add some compatibility test that verify the forward & backward consistency.

Cypher30 (changes requested):
Please address the CI failure~

tests/test_fx/test_meta/test_aten.py
```diff
@@ -0,0 +1,88 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from colossalai.fx.profiler import MetaTensor
+
+import pytest
+
+try:
+    meta_lib = torch.library.Library("aten", "IMPL", "Meta")
+    incompatible = False    # version > 1.12.0
```
FrankLeeeee, line 11: Use constant for this value  i.e. `INCOMPATIBLE = False`

super-dainiu: Solved.

Cypher30 (approved):

PR #1531, opened 2022-09-01
[fx] Modify solver linearize and add corresponding test

Cypher30:
# What's New

In PR #1496, @super-dainiu suggests that we could use more elegant linearize code and node structure, but at that time we still worried about the time overhead for solver to solve the sequence. After some experiments, I'm sure that this old linearize could be left and use the suggested structure. I also add a test to check if the solver's sequence and annotated result is correct. Lastly, I fix a couple of lines of old inefficient code in `activation_checkpoint_codgen.py` and `ckpt_solver_rotor.py`.


Cypher30:
Screen Shot 2022-09-01 at 16 35 46
Currently we use MetaTensor implemented by @super-dainiu, which could not be used in torch11, the above is the result of `test_fx`.

tests/test_fx/test_ckpt_solvers/test_linearize.py
```diff
@@ -0,0 +1,139 @@
+from ast import For
+from binhex import REASONABLY_LARGE
+from typing import Callable
```
super-dainiu, line 3: What are these imports?

tests/test_fx/test_ckpt_solvers/test_linearize.py
```diff
@@ -0,0 +1,139 @@
+from ast import For
+from binhex import REASONABLY_LARGE
```
FrankLeeeee, line 2: remove unused imports.

tests/test_fx/test_ckpt_solvers/test_linearize.py
Cypher30: I will clean those imports, probably caused by viscose

tests/test_fx/test_ckpt_solvers/test_linearize.py
Cypher30: Removed~

super-dainiu (approved):

PR #1541, opened 2022-09-05
[hotfix] change namespace for meta_trace.

super-dainiu:
The `namespace` for this tracer should be default as `graph._graph_namespace`, so that variable names are correct in the display.


Cypher30 (approved):

Cypher30 (commented):
Maybe you could fix it within other PR?

PR #1542, opened 2022-09-05
[fx] Add common node in model linearize

Cypher30:
# What's New

We were inspired by the common node idea in [merak](https://arxiv.org/abs/2206.04959), and modify the naive process of defining a common node in [merak's source code](https://github.com/HPDL-Group/Merak/blob/91864243b21ccfef1d58746623e3a6cebdc630ae/Merak/autoshard/convert.py#L134). Now we are able to address the attention mask problem in hugginface BERT model, separating each attention block and MLP in the encoder.

Moreover, I also fixed the problem that our solver might repeatedly count the memory for combination like "BN-RELU", as the current linearize strategy will generate more isolated in-place RELU node in the linearized node list.


colossalai/fx/passes/algorithms/ckpt_solver_rotor.py
```diff
@@ -7,6 +7,7 @@
 from .utils import *
 from colossalai.fx.profiler import profile_function, profile_module
 from colossalai.fx.passes.meta_info_prop import MetaInfoProp
+import pdb
```
super-dainiu, line 10: remove this

colossalai/fx/passes/algorithms/linearize.py
```diff
@@ -2,11 +2,12 @@
 from torch.fx import GraphModule, Node
 
 
-def linearize(gm: GraphModule) -> List[List[Node]]:
+def linearize(gm: GraphModule, cnode: List[str] = None) -> List[List[Node]]:
```
super-dainiu, line 5: where do you get the cnode?

colossalai/fx/passes/algorithms/ckpt_solver_rotor.py
Cypher30: okay, forget lol

colossalai/fx/passes/algorithms/linearize.py
Cypher30: yes, it should be added to solver

super-dainiu: I mean how do you get the list of cnode?

Cypher30: The first list of common nodes should be input by users, and the other common nodes will be propagated in the following codes.

super-dainiu (approved):

PR #1586, opened 2022-09-12
[fx] Improve linearize and rotor solver

Cypher30:
# What's New

In this PR, I improve the checkpoint pass in the following two ways:

## 1. Improve linearize common node detection

Previously we have add common node mechanism to linearize, however, in some model such as GPT2, the common nodes are not derived from placeholder. In fact, the output of operation such as `getattr` is a common node. In this PR, I add the `COPS` to pick those operations.

## 2. Improve rotor solver

In my last PR #1585 , I have added the nested activation checkpoint codegen so that we could support the rotor solver's strategy. In this PR, I update the solver annotation process so that it is able to detect nested activation checkpoint. I also modify the DP process as there is something seems not comply with the paper. Lastly, I carefully modify the estimation part so that it could be more sensible.


colossalai/fx/passes/algorithms/ckpt_solver_rotor.py
```diff
@@ -128,8 +129,8 @@ def _compute_output_size(node: List[Node]) -> int:
         int: output size
     """
 
-    return node[-1].meta['tensor_meta'].numel * torch.tensor([],
-                                                             dtype=node[-1].meta['tensor_meta'].dtype).element_size()
+    return node[-1].meta['tensor_meta'].numel * \
+    torch.tensor([], dtype=node[-1].meta['tensor_meta'].dtype).element_size()
```
FrankLeeeee, line 133: Did you use pre-commit?

colossalai/fx/passes/algorithms/linearize.py
```diff
@@ -1,6 +1,25 @@
-from typing import List
+from typing import List, Any
 from torch.fx import GraphModule, Node
 
+# List of target name that could be seen as common node
```
FrankLeeeee, line 4: Include the definition of `common node` in the documentation for future reference.

Cypher30: okay, I will add the definition

colossalai/fx/passes/algorithms/ckpt_solver_rotor.py
Cypher30: Ahhh I probably forgot

FrankLeeeee (approved):

PR #1598, opened 2022-09-14
[fx] Add offload codegen

Cypher30:
# What's New

Previously we have offload option in activation checkpoint region to support the offload input process, however, in the upcoming activation solver, we might offload the input of a node that is not inside any checkpoint region. Therefore, I use saved_tensors_hooks for this kind of offload manner.

As we haven't implemented the torch11 version of `ColoGraphModule`, I skip the unit test for this part and attach the results on torch12 below
Screen Shot 2022-09-14 at 14 35 33


FrankLeeeee:
Paste an example of the generated code here.

tests/test_fx/test_codegen/test_offload_codegen.py
```diff
@@ -0,0 +1,160 @@
...
+    for node in graph.nodes:
+        if node.name == "linear2":
+            setattr(node, "activation_offload", True)
+        if node.name == "linear3":
+            setattr(node, "activation_offload", True)
+            setattr(node, "activation_checkpoint", [0])
+        if node.name == "linear4":
+            setattr(node, "activation_checkpoint", [0])
+
+    gm = ColoGraphModule(copy.deepcopy(model), graph)
```
FrankLeeeee, line 137: I think copy is not needed.

Cypher30: Copy is needed because we want to test backward gradients.

Cypher30:
> Paste an example of the generated code here.

Here is an example for the generated code:
```python
def pack_hook(self, x):
    if getattr(x, "offload", None):
        return (x.device, x.cpu())
    else:
        return x

def unpack_hook(self, packed):
    if isinstance(packed, tuple):
        device, tensor = packed
        return tensor.to(device)
    else:
        return packed

def checkpoint_0(self, linear2):
    linear3 = self.linear3(linear2);  linear2 = None
    linear4 = self.linear4(linear3);  linear3 = None
    return linear4

def forward(self, x):
    linear1 = self.linear1(x);  x = None
    setattr(linear1, 'offload', True)
    with torch.autograd.graph.saved_tensors_hooks(self.pack_hook, self.unpack_hook):
        linear2 = self.linear2(linear1);  linear1 = None
    linear4 = colossalai.utils.activation_checkpoint.checkpoint(self.checkpoint_0, True, linear2, use_reentrant=False)
    linear5 = self.linear5(linear4);  linear4 = None
    return linear5
```

tests/test_fx/test_codegen/test_offload_codegen.py
```diff
@@ -0,0 +1,160 @@
...
+    "setattr(linear1, 'offload', True)" in code and \
+    "with torch.autograd.graph.saved_tensors_hooks(self.pack_hook, self.unpack_hook):" in code and \
+    "colossalai.utils.activation_checkpoint.checkpoint(self.checkpoint_0, True, linear2, use_reentrant=False)" in code
+
+    _test_fwd_and_bwd(model, gm, data)
+    gpc.destroy()
+
+
+@pytest.mark.skipif(with_codegen, reason='torch version is equal to or higher than 1.12.0')
+@pytest.mark.skip(reason="currently torch11 ColoGraphModule is not implemented")
```
FrankLeeeee, line 154: why need two skips?

Cypher30: Sure I could remove the first one~

tests/test_fx/test_codegen/test_offload_codegen.py
FrankLeeeee: ok i see.

FrankLeeeee (approved):

colossalai/fx/codegen/activation_checkpoint_codegen.py
```diff
@@ -307,8 +342,27 @@ def emit_code_with_nested_activation_checkpoint(body, ckpt_func, nodes, emit_nod
...
+
+            # if a node is outside of checkpoint region and want to offload
+            # it's input activation, we will use torch.saved_tensors_hooks
+            # to complete the offload process.
+            if getattr(node, "activation_offload", False):
+                if not is_hook_inserted:
+                    pack_hook, unpack_hook = _gen_saved_tensors_hooks()
+                    ckpt_func.insert(0, "\n".join([pack_hook, unpack_hook]) + "\n")
+
+                for par in node._input_nodes:
```
super-dainiu, line 354: I think `node._input_nodes` should be `node.all_input_nodes`

Cypher30: Sure it will be better~

Cypher30: fixed

PR #1600, opened 2022-09-15
[autoparallel] add bcast op handler

YuliangLiu0306:
Bcast op in this scenario means the arguments of this kind of operation can be automatically expanded to be of equal sizes (without making copies of the data). To keep the computing correctness, we should make sure on each dimension two arguments have the same DimSpec. Additionally, the BcastOpHandler will enumerate all the possible sharding specs for the output and generate the input sharding spec based on the output sharding spec and input tensor meta information.

### DIFF
--- a/colossalai/auto_parallel/solver/_utils.py
+++ b/colossalai/auto_parallel/solver/_utils.py
@@ -4,6 +4,10 @@ from torch.fx.node import Node
 from colossalai.tensor.sharding_spec import ShardingSpec
 from colossalai.device.device_mesh import DeviceMesh
 from typing import Union, Dict, List, Optional
+import warnings
+from functools import reduce
+import functools
+import operator
 
 
 def generate_sharding_spec(input_: Union[Node, torch.Tensor], device_mesh: DeviceMesh,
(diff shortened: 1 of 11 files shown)


colossalai/auto_parallel/solver/_utils.py
```diff
@@ -4,6 +4,8 @@
 from colossalai.tensor.sharding_spec import ShardingSpec
 from colossalai.device.device_mesh import DeviceMesh
 from typing import Union, Dict, List, Optional
+from functools import reduce
+import operator
```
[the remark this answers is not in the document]
YuliangLiu0306: Removed

tests/test_auto_parallel/test_bcast_handler.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __init__(self, c_in, c_out):
+        super().__init__()
+        self.conv1 = nn.Conv2d(c_in, c_out, kernel_size=3, padding=1)
+        self.conv2 = nn.Conv2d(c_in, c_out, kernel_size=3, padding=1)
+
+    def forward(self, x):
+        x1 = self.conv1(x)
+        x2 = self.conv2(x)
+        x = x1 + x2
```
FrankLeeeee, line 22: These two tensors are of the same shape, it is better to test for different but broadcastable shapes

colossalai/auto_parallel/solver/op_handler/bcast_op_handler.py
```diff
@@ -0,0 +1,165 @@
...
+                                     entire_shape=shape,
+                                     dim_partition_dict=processed_dim_partition_dict)
+
+        return sharding_spec
+
+    def _generate_resharding_costs(self, sharding_specs):
+        # The resharding_cost of weight is counted due to sharing weight cases.
+        dtype = self.node._meta_data.dtype
+        nodes = self.predecessor_node
+        # The resharding_cost of weight is counted due to sharing weight cases.
```
FrankLeeeee, line 53: The two comments are duplicated and seems irrelevant?

tests/test_auto_parallel/test_bcast_handler.py
YuliangLiu0306: Sure, I will construct a more general test case.

FrankLeeeee (approved):

PR #1608, opened 2022-09-19
[fx] Add pofo solver

Cypher30:
# What's New?

The new pofo solver is based on paper [Efficient Combination of Rematerialization and Offloading for Training DNNs](https://proceedings.neurips.cc/paper/2021/hash/c8461bf13fca8a2b9912ab2eb1668e4b-Abstract.html) and it's code given in the supplemental. Currently we doesn't use the whole set up in the original paper and reuse rotor solver for the backward sequence as suggested in supplemental. The solver now is able to find strategy with offload. I will redesign the codegen later and add graph annotation to this solver when the codegen is compatible.


colossalai/fx/passes/algorithms/ckpt_solver_pofo.py
```diff
@@ -0,0 +1,396 @@
...
+        return self._get_value(state, self.what, INF)
+
+    def set_value(self, state, opt, what):
+        i, act_size, df, db, input_has_bar = state
+        self.opt[input_has_bar][i][act_size][(df, db)] = opt
+        self.what[input_has_bar][i][act_size][(df, db)] = what
+
+
+class PofoSolver:
+
```
FrankLeeeee, line 81: You should include a docstring here. You can move your PR description here so that it will appear on our API  documentation.

colossalai/fx/passes/algorithms/ckpt_solver_pofo.py
```diff
@@ -0,0 +1,396 @@
+from typing import List, Tuple
+import copy
+import torch
+from torch.fx import GraphModule, Node
+from colossalai.fx.graph_module import ColoGraphModule
+from colossalai.fx.profiler import parameter_size
+import math
+from .linearize import linearize
+from .utils import *
```
FrankLeeeee, line 9: do not use `*`.

colossalai/fx/passes/algorithms/utils.py
```diff
@@ -1,3 +1,10 @@
+import math
+
+
+def _discretize(mem_unit, values):
+    return [math.ceil(value / mem_unit) for value in values]
+
+
 class Chain:
```
FrankLeeeee, line 8: These classes in the `utils` file are not really `utils`, they are more like to be in  a file like `operation.py` or `action.py`.

FrankLeeeee (approved):

PR #1610, opened 2022-09-20
[autoparallel] refactored the data structure for sharding strategy

FrankLeeeee:


colossalai/auto_parallel/solver/sharding_strategy.py
```diff
@@ -37,6 +37,45 @@ class ShardingStrategy:
...
+    TrainCycleItem is a dataclass to store the items which have different values for the forward and backward pass
+    in a training iteration.
+
+    Args:
+        fwd (Any): the item for the forward pass
+        bwd (Any): the item for the backward pass
+    """
+    fwd: Any
+    bwd: Any
+
```
YuliangLiu0306, line 52: I think we could add a 'total' attribute into TrainCycleItem, I find it is useful during preparing costs information for solver.

FrankLeeeee: ok

YuliangLiu0306 (approved):

PR #1612, opened 2022-09-20
[autoparallel] added new node handler

FrankLeeeee:
A new `NodeHandler` is implemented to replace the original `OperatorHandler`.  This new `NodeHandler` has several improvements over the `OperatorHandler`.
1. It defines the mapping between the logical operand and the actual tensor via the `get_operand_mapping` method. This unifies the way of writing handles for function/module nodes.
2. It has an additional abstraction of StrategyGenerator, which allows clear grouping and re-use of different sharding strategies.
3. It has a consistent template for `register_strategy` to avoid code redundancy.

### DIFF
--- a/colossalai/auto_parallel/solver/sharding_strategy.py
+++ b/colossalai/auto_parallel/solver/sharding_strategy.py
@@ -1,4 +1,7 @@
 from dataclasses import dataclass
+from abc import ABC, abstractmethod
+from enum import Enum
+from colossalai.device.device_mesh import DeviceMesh
 from colossalai.tensor.sharding_spec import ShardingSpec
 from typing import Dict, List, Union, Tuple, Any
 from torch.fx.node import Node
@@ -37,6 +40,20 @@ class ShardingStrategy:
     input_shardings: List[ShardingSpec] = None
 
 
+class OperandType(Enum):
+    """
+    An operand can come from the argument list of an operator or the parameter list of a module.
+    """
+    ARG = 0
+    PARAM = 1
+
+
+@dataclass
+class Operand:
+    name: str
+    type: OperandType
+
+
 @dataclass
 class TrainCycleItem:
     """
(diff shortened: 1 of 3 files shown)


colossalai/auto_parallel/solver/sharding_strategy.py
```diff
@@ -37,20 +40,36 @@ class ShardingStrategy:
...
     TrainCycleItem is a dataclass to store the items which have different values for the forward and backward pass
     in a training iteration.
 
     Args:
-        fwd (Any): the item for the forward pass
-        bwd (Any): the item for the backward pass
-        total (Any): the total value for the forward and backward pass
+        fwd (float): the item for the forward pass
+        bwd (float): the item for the backward pass
     """
```
YuliangLiu0306, line 66: I prefer the previous version, because the fwd cost may be a tuple for memory cost(activation memory cost, weight memory cost). And the communication cost and computation cost can benefit from this. For example, the weight memory cost, not the total cost, is just what we need to compute the communication cost of batch sharding cases.

colossalai/auto_parallel/solver/sharding_strategy.py
```diff
@@ -37,20 +40,36 @@ class ShardingStrategy:
...
-    fwd: Any
-    bwd: Any
-    total: Any
+    fwd: float
+    bwd: float
+
+    @property
+    def total(self):
+        return self.fwd + self.bwd
```
YuliangLiu0306, line 72: And we can't samply add the fwd and bwd cost together, such as total memory cost is not the sum of bwd memory cost and fwd memory cost

colossalai/auto_parallel/solver/sharding_strategy.py
FrankLeeeee: Ok I see

FrankLeeeee: recovered.

YuliangLiu0306 (approved):

PR #1616, opened 2022-09-20
[autoparallel] added new linear module handler

FrankLeeeee:
This PR implements the new version of linear module handler by inheriting the new node handler and the new data structures for sharding strategy. The various strategy generators are to be implemented in the next PR.

### DIFF
--- a/colossalai/auto_parallel/solver/sharding_strategy.py
+++ b/colossalai/auto_parallel/solver/sharding_strategy.py
@@ -1,6 +1,10 @@
 from dataclasses import dataclass
 from abc import ABC, abstractmethod
 from enum import Enum
+import operator
+import torch
+from functools import reduce
+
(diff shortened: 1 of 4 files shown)


colossalai/auto_parallel/solver/sharding_strategy.py
```diff
@@ -69,6 +90,20 @@ class TrainCycleItem:
     total: Any
 
 
+class CommunicationType(Enum):
```
YuliangLiu0306, line 93: I think we could unify all the communication related operation to CommSpec, which could give communication cost information using get_comm_cost api, and support runtime application with convert_spec_to_action. The resharding costs is estimated by ShapeConsistencyManager which uses CommSpec internally.

YuliangLiu0306 (approved):

PR #1618, opened 2022-09-20
[fx] Modify offload codegen

Cypher30:
# What's New

In previous PR #1598 , I add offload mechanism into our codegen, however, after implementing #1608 , I found that the old version might not satisfies our needs. In this PR, I consider the offload outside the activation checkpoint region to be formulated in the following form: Each node will be assigned one offload label (if needed) in the form of `(idx, offload_input, offload_bar)`, the `idx` term is similar to activation codegen, indicates the number of offload region (starts from 0 and in the future I will add some assertion to make sure the index increases correctly); `offload_input` indicates whether we need to offload the input given to this region, and `offload_bar` indicates whether we need to offload all the intermediate results inside this offload region. I will show you the annotation process in the next PR that dealing with the pofo sequence annotation.

As we skip the test in torch11, the followings are the test results in torch12
Screen Shot 2022-09-20 at 20 21 15


FrankLeeeee:
Paste an example here.

Cypher30:
> Paste an example here.

Currently there will be four kinds of node that will be annotated with `activation_offload` attribute. And there will be two kinds of annotations, bool or tuple in the form of `(region_idx, offload_input, offload_intermediate)`, the details are as follows:
1. The beginning of a checkpoint region, if it needs to offload the input of checkpoint region, it will be annotated with `setattr(node, "activation_offload", True)`
2. An offload region that enable gradient, needs to offload the input into the region and all the intermediate results, it will be annotated with  `setattr(node, "activation_offload", (idx, True, True))`
3. An offload region that enable gradient, only needs to offload the input, will be annotated with `setattr(node, "activation_offload", (idx, True, False))`
4. An offload region that enable gradient, only needs to offload it's intermediate results, it will be annotated with `setattr(node, "activation_offload", (idx, False, True))`

For example, in the test, I annotate graph with
```python
    for node in graph.nodes:
        if node.name == "linear0":
            setattr(node, "activation_offload", (0, True, False))
        if node.name == "linear1":
            setattr(node, "activation_offload", (0, True, False))
        if node.name == "linear2":
            setattr(node, "activation_offload", (1, True, True))
        if node.name == "linear4":
            setattr(node, "activation_offload", (2, False, True))
        if node.name == "linear5":
            setattr(node, "activation_checkpoint", [0])
            setattr(node, "activation_offload", True)
```

The generated code will be
```python
def pack_hook_input(self, x):
    if getattr(x, "offload", False):
        return (x.device, x.cpu())
    else:
        return x
 
def pack_hook_no_input(self, x):
    if getattr(x, "offload", True):
        return (x.device, x.cpu())
    else:
        return x

def unpack_hook(self, packed):
    if isinstance(packed, tuple):
        device, tensor = packed
        return tensor.to(device)
    else:
        return packed

def checkpoint_0(self, linear4):
    linear5 = self.linear5(linear4);  linear4 = None
    return linear5

def forward(self, x):
    setattr(x, 'offload', True)
    with torch.autograd.graph.saved_tensors_hooks(self.pack_hook_input, self.unpack_hook):
        linear0 = self.linear0(x);  x = None
        linear1 = self.linear1(linear0);  linear0 = None
    with torch.autograd.graph.save_on_cpu(pin_memory=True):
        linear2 = self.linear2(linear1);  linear1 = None
    linear3 = self.linear3(linear2);  linear2 = None
    setattr(linear3, 'offload', False)
    with torch.autograd.graph.saved_tensors_hooks(self.pack_hook_no_input, self.unpack_hook):
        linear4 = self.linear4(linear3);  linear3 = None
    linear5 = colossalai.utils.activation_checkpoint.checkpoint(self.checkpoint_0, True, linear4, use_reentrant=False)
    linear6 = self.linear6(linear5);  linear5 = None
    return linear6
```

colossalai/fx/codegen/activation_checkpoint_codegen.py
```diff
@@ -118,6 +142,51 @@ def _find_ckpt_regions(nodes: List[Node]):
     return ckpt_regions
 
 
+def _find_offload_regions(nodes: List[Node]):
+    """This function is to find the offload regions
+    In pofo algorithm, during annotation, we will annotate the offload region with the 
+    tuple in the form of (idx, offload_input, offload_bar). idx indicates the offload
+    region's index, offload_input is a bool type indicates whether we need to offload
+    the input, offload_bar is a bool type indicates whether we need to offload all the
+    intermediate x_bars of this region.
```
FrankLeeeee, line 151: This docstring mentions that this is for pofo algorithm, does it work for the other algo as well?

FrankLeeeee (approved):

colossalai/fx/codegen/activation_checkpoint_codegen.py
Cypher30: The other algorithms (e.g. rotor or chen_greedy) don't make use of offload mechanism, and I think this function is definitely not that general for the future activation solver because in the future we might not be able to hold the same assumption, but currently it is needed by the pofo solver.

PR #1619, opened 2022-09-21
[fx/profiler] tuned the calculation of memory estimation

super-dainiu:
### What's modified?
To do `MetaInfoProp` on arbitrary physical devices, we need to wrap a `torch.Tensor` with `MetaTensor`, which has a property of `fake_device`. `fake_device` is used by `__torch_dispatch__` to find an `torch.ops.aten` for execution, while in real execution, the tensor is moved back to `device='meta'`. So now you should do `MetaInfoProp` on a `tm.resnet18` on CPU with the following script.
```python
model = tm.resnet18()
input = MetaTensor(torch.rand(10000, 3, 224, 224, device='meta'), fake_device='cpu')
MetaInfoProp(gm).run(input)
```
And on GPU with the following script.
```python
model = tm.resnet18().cuda()    # don't forget to move your model to cuda as well!
input = MetaTensor(torch.rand(10000, 3, 224, 224, device='meta'), fake_device='cuda')
MetaInfoProp(gm).run(input)
```
Now you might observe some different patterns in estimated memory, because `torch.autograd` behaves differently on different devices.
### Improvements
1. The computation graph is completely different when executing on 'cpu' and 'cuda'.
2. Tensor `dtype` may change during execution.
3. Not every tensor that passes through `torch.autograd.graph.saved_tensors_hooks` is saved to device. `tensor.data_ptr()` marks its physical address. No duplicated `tensor.data_ptr()` should be saved.
    1. One stupid thing is that we always have `tensor.data_ptr()=0` on `device='meta'`
    2. Luckily in Python frontend, we can still track the same `tensor` with the unique identifier of `tensor.data_ptr` as a `torch._C function`.
4. See the graph below for illustrations.
image
image

    1. Whether or not fwd_out should be saved is determined by the node's users. If and only if any of the node's users have saved `fwd_in` can this node have `fwd_out`. 
5. `nn.MultiheadAttention` has `(x, x, x)` as input, but only one `x` should be saved for backward.
### Concerns
Backward memory is indeed too hard to estimate.

### DIFF
--- a/colossalai/_meta_registrations.py
+++ b/colossalai/_meta_registrations.py
@@ -175,6 +175,11 @@ def meta_hardswish(input: torch.Tensor):
     return torch.empty_like(input)
 
 
+@register_meta(aten.hardtanh.default)
+def meta_hardtanh(input: torch.Tensor, min, max):
+    return torch.empty_like(input)
+
+
 @register_meta(aten.hardswish_backward.default)
 def meta_hardswish_backward(grad_out: torch.Tensor, input: torch.Tensor):
     grad_in = torch.empty_like(input)
@@ -189,7 +194,7 @@ def meta_hardtanh_backward(grad_out: torch.Tensor, input: torch.Tensor, min_val:
 
 @register_meta(aten.roll.default)
 def meta_roll(input: torch.Tensor, shifts, dims):
-    return torch.empty_like(input)
+    return input
 
 
 @register_meta(aten.native_batch_norm.default)
@@ -211,13 +216,39 @@ def meta_bn_backward(dY: torch.Tensor, input: torch.Tensor, weight: torch.Tensor
     return dX, dgamma, dbeta
 
 
-@register_meta(aten.native_layer_norm.default)
-def meta_ln(input: torch.Tensor, normalized_shape, weight, bias, eps):
+# https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cudnn/BatchNorm.cpp
+@register_meta(aten.cudnn_batch_norm.default)
+def meta_cudnn_bn(input: torch.Tensor, weight, bias, running_mean, running_var, training, momentum, eps):
     n_input = input.size(1)
 
     output = torch.empty_like(input)
     running_mean = torch.empty((n_input), device='meta')
     running_var = torch.empty((n_input), device='meta')
+    reserve = torch.empty((0), dtype=torch.uint8, device='meta')
+    return output, running_mean, running_var, reserve
+
+
+# https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cudnn/BatchNorm.cpp
+# NB: CuDNN only implements the backward algorithm for batchnorm
+# in training mode (evaluation mode batchnorm has a different algorithm),
+# which is why this doesn't accept a 'training' parameter.
+@register_meta(aten.cudnn_batch_norm_backward.default)
+def meta_cudnn_bn_backward(dY: torch.Tensor, input: torch.Tensor, weight: torch.Tensor, running_mean, running_var,
+                           save_mean, save_invstd, eps, reserve):
+    dX = torch.empty_like(input)
+    dgamma = torch.empty_like(weight)
+    dbeta = torch.empty_like(weight)
+    return dX, dgamma, dbeta
+
+
+@register_meta(aten.native_layer_norm.default)
+def meta_ln(input: torch.Tensor, normalized_shape, weight, bias, eps):
+    bs = input.size(0)
+    n_input = input.size(1)
+
+    output = torch.empty_like(input)
+    running_mean = torch.empty((bs, n_input, 1), device='meta')
+    running_var = torch.empty((bs, n_input, 1), device='meta')
     return output, running_mean, running_var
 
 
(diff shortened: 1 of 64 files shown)


colossalai/_meta_registrations.py
```diff
@@ -211,13 +216,39 @@ def meta_bn_backward(dY: torch.Tensor, input: torch.Tensor, weight: torch.Tensor
...
+
+@register_meta(aten.native_layer_norm.default)
+def meta_ln(input: torch.Tensor, normalized_shape, weight, bias, eps):
+    bs = input.size(0)
+    n_input = input.size(1)
+
+    output = torch.empty_like(input)
+    running_mean = torch.empty((bs, n_input, 1), device='meta')
+    running_var = torch.empty((bs, n_input, 1), device='meta')
     return output, running_mean, running_var
```
super-dainiu, line 252: A sad truth is autograd on cuda use `cudnn_batch_norm`

colossalai/_meta_registrations.py
```diff
@@ -338,6 +369,23 @@ def meta_embedding_dense_backward(grad_output: torch.Tensor, indices: torch.Tens
...
+    # notice that mask is bool
+    output = torch.empty_like(input)
+    mask = torch.empty_like(input, dtype=torch.bool)
+    return output, mask
+
+
+# https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/Dropout.cpp
+@register_meta(aten.native_dropout_backward.default)
+def meta_native_dropout_backward_default(grad: torch.Tensor, mask: torch.Tensor, scale: float):
+    return torch.empty_like(grad)
```
super-dainiu, line 391: Another sad truth is autograd on cuda use `native_dropout`

colossalai/fx/passes/algorithms/ckpt_solver_rotor.py
```diff
@@ -123,7 +124,9 @@ def _fwd_xbar(node: List[Node]) -> int:
 
     xbar = 0
     for n in node:
-        xbar += n.meta['fwd_mem_tmp'] + n.meta['fwd_mem_out']
+        xbar += n.meta['fwd_mem_tmp']
+        if any(map(lambda x: x.meta['save_fwd_in'], n.users)):
+            xbar += n.meta['fwd_mem_out']
```
super-dainiu, line 129: Not every node has `fwd_mem_out`.

colossalai/fx/passes/meta_info_prop.py
```diff
@@ -224,7 +222,7 @@ def output(self, target: 'Target', args: Tuple[Argument, ...], kwargs: Dict[str,
             result (Any): The argument value that was retrieved
             meta_info (MetaInfo): The memory cost and FLOPs estimated with `MetaTensor`.
         """
-        return args[0], GraphInfo(fwd_mem_in=activation_size(args[0]))
+        return args[0], GraphInfo(save_fwd_in=True)
```
super-dainiu, line 225: `output` node saves the last `fwd_mem_out` for sure

colossalai/fx/profiler/dataflow.py
```diff
@@ -99,25 +100,25 @@ def _peak_memory(deps: Dict[Node, int]):
...
             if len(n.users):
                 graph_info.bwd_mem_tmp = max(graph_info.bwd_mem_tmp, _peak_memory(deps))
             else:
                 # TODO: some of the bwd_mem_out might be model parameters.
                 # basically a backward node without user is a `grad_out` node
-                graph_info.bwd_mem_out += activation_size(n.meta['out'])
+                graph_info.bwd_mem_out += activation_size(n.meta['saved_tensor'])
         for input_n in n.all_input_nodes:
             if input_n in deps:
                 deps[input_n] -= 1
```
super-dainiu, line 124: these analysis are still naive, so sad

colossalai/fx/profiler/memory.py
```diff
@@ -1,88 +1,10 @@
...
-
-    # TODO: list all call_methods that are not inplace here
-    NON_INPLACE_METHOD = [
-        'chunk',
-        'contiguous',
-        'expand',
-        'mean',
-        'split',
-    ]
-    __all__ += ['INPLACE_OPS', 'INPLACE_METHOD', 'NON_INPLACE_METHOD']
```
super-dainiu, line 84: move them to `constant.py`

colossalai/fx/profiler/opcount.py
```diff
@@ -201,6 +202,8 @@ def zero_flop_jit(*args):
     # normalization
     aten.native_batch_norm.default: batchnorm_flop_jit,
     aten.native_batch_norm_backward.default: batchnorm_flop_jit,
+    aten.cudnn_batch_norm.default: batchnorm_flop_jit,
+    aten.cudnn_batch_norm_backward.default: partial(batchnorm_flop_jit, training=True),
```
super-dainiu, line 206: notice that cudnn_batch_norm_backward is different because it is only used when `training=True`

colossalai/fx/profiler/profiler.py
```diff
@@ -21,7 +27,17 @@ def is_autogradable(x):
     return isinstance(x, torch.Tensor) and x.is_floating_point()
 
 
-def _profile(target: Callable, *args, **kwargs) -> Tuple[Any, ...]:
+# super-dainiu:
+# x.detach() will change the unique identifier of data_ptr
+# we need to handle this in a stupid way
```
super-dainiu, line 32: x.detach() will change the unique identifier of data_ptr. we need to handle this in a stupid way

colossalai/fx/profiler/profiler.py
```diff
@@ -1,15 +1,21 @@
...
+from .memory import activation_size
+from .constant import ALIAS_ATEN
 from .tensor import MetaTensor
 from .opcount import flop_mapping
 
 __all__ = ['profile_function', 'profile_module', 'profile_method']
 
+# super-dainiu: this cache should be global, otherwise it cannot
+# track duplicated tensors between nodes
+cache = set()
```
super-dainiu, line 17: this cache should be global, otherwise it cannot track duplicated tensors between nodes

colossalai/fx/profiler/profiler.py
```diff
@@ -228,13 +277,17 @@ def profile_module(module: torch.nn.Module) -> Callable:
...
-        # skip the autograd profiling.
-        if getattr(module, 'inplace', False):
-            args = tree_map(lambda x: x.to('meta'), args)
-            kwargs = tree_map(lambda x: x.to('meta'), kwargs)
-            out = func(*args, **kwargs)
-            return out, GraphInfo(out.numel(), out.numel(), 0, 0, 0, 0)
+        # still run the profiling but discard some results regarding `module`.
+        inplace = getattr(module, 'inplace', False)
+        if inplace:
+            module.inplace = False
```
super-dainiu, line 282: no longer skip profiling for `inplace=True`

super-dainiu:
image

FrankLeeeee (approved):

PR #1623, opened 2022-09-21
[feature] new zero implementation

1SAA:
# _A New ZeRO Implementation_

## Backgrounds
In the current version, our ZeRO has a performance issue. The reason is that our asymmetric distribution of chunks makes one process hinder other processes when reading the content of the chunk  located in the CPU memory. It really prolongs the time of the data transmission and undermines the efficiency of ZeRO.

## Implementation
In order to solve this problem, I refactored the class `Chunk`. The new chunk is distributed evenly to all processes. All processes can move the data from the CPU memory to the CUDA memory in the same time. Furthermore, I provide an option to enable the pin memory for chunks. Now all chunks can have a copy in the pinned CPU memory. The above optimizations prominently improved the efficiency of data movements between CPU and CUDA.

## Another Advantage
The new ZeRO supports the true hybrid parallelism. It creates different chunk groups for parameters which have different DP communication groups. This brings huge flexibility to our up coming automatic configuration of parallelism.


colossalai/zero/zero_optimizer.py
```diff
@@ -61,11 +58,20 @@ def __init__(self,
...
-        self.fp16_param_to_fp32_param: Dict[torch.Tensor, torch.Tensor] = {}
+        self.param_to_range: Dict[Parameter, Tuple[int, int]] = dict()
+        self.param_to_chunk32: Dict[Parameter, Chunk] = dict()
+        self.chunk16_set: Set[Chunk] = set()
+
         for p, fp32_p in zip(module.parameters(), module.fp32_params):
-            self.fp16_param_to_fp32_param[p] = fp32_p
+            chunk_16 = self.chunk_manager.get_chunk(p)
+            chunk_32 = self.chunk_manager.get_chunk(fp32_p)
+            chunk_32.init_pair(chunk_16)
```
ver217, line 70: Why not init pair in ZeroDDP?

colossalai/gemini/chunk/chunk.py
```diff
@@ -306,46 +398,56 @@ def copy_tensor_to_chunk_slice(self, tensor: torch.Tensor, data_slice: torch.Ten
...
+            assert friend_chunk.is_gathered is True
+            self.chunk_total.copy_(friend_chunk.chunk_total)
+            self.optim_sync_flag = True
+        elif friend_chunk.device_type == 'cuda' and self.device_type == 'cuda':
+            self.cuda_shard.copy_(friend_chunk.cuda_shard)
+            self.optim_sync_flag = True
+            self.cpu_vis_flag = False
         else:
-            assert self.cuda_shard is not None    # only check in CUDA
-            valid_tensor = self.cuda_shard[:self.valid_end]
```
ver217, line 334: Where do you copy fp32 param to fp16 param when both of them are on CPU?

colossalai/zero/zero_optimizer.py
1SAA: It's a problem for a legacy reason. I'll move it to ZeroDDP.

colossalai/gemini/chunk/chunk.py
1SAA: See `shard_move` function in the `Chunk` class.

feifeibear (approved):

PR #1629, opened 2022-09-22
[autoparallel] add layernorm handler

YuliangLiu0306:


YuliangLiu0306:
This PR provides a LayerNormHandler to generator strategies for layer norm operation.  

Note:  To keep the math consistency, LayerNormHandler won't generate strategies with sharding on hidden dimension.

colossalai/auto_parallel/solver/strategies_constructor.py
```diff
@@ -349,35 +359,72 @@ def build_strategies_and_cost(self):
...
+                                                         memory_cost=memory_cost)
+                    strategies_vector.append(sharding_strategy)
+
+                # op list to be processed to support gpt2
+                elif target in (builtins.getattr, operator.le, torch.addmm, operator.pow, torch.where, torch.softmax,
+                                torch.nn.functional.softmax, torch.pow, torch.tanh):
+                    pass
                 # other function
                 else:
                     raise RuntimeError(f'{target} function is NOT supported now.')
```
FrankLeeeee, line 410: This should be offloaded to node handler to produce the strategy. We can do this later.

FrankLeeeee (approved):

colossalai/auto_parallel/solver/strategies_constructor.py
YuliangLiu0306: Yes, if we assign an op handler for all kind of op, this should be offloaded. Otherwise, some strategies of simple operations will be generated in StrategiesConstructor.

PR #1636, opened 2022-09-23
[autoparallel] change the following nodes strategies generation logic

YuliangLiu0306:
This PR changes the following nodes(nodes will be merged into other nodes) strategies generation logic, new version handlers for following nodes will **keep the duplicated strategies**. **To avoid the duplicated  check in StrategiesConstructor,** handler will **attatch the index of input strategy on new strategy name.** 

**This behavior will not increase the seaching space**, because no matter how many strategies generated by handler, **the following nodes will be merged and solver won't create a new variable for it.**


FrankLeeeee (approved):

YuliangLiu0306:
image

PR #1637, opened 2022-09-23
[autoparallel] Add pofo sequence annotation

Cypher30:
# What's New?

Previously I add pofo solver in #1608, and in this PR I finish annotate graph module with pofo sequence, with the help of PR #1618 , finishing the complete pipeline of pofo solver.

The following is a brief summary of our rotor/pofo solver pipeline:
1. Using excellent `MetaInfoProp` tool by @super-dainiu to get all the meta information in the graph.
2. Linearize the graph with `linearize` module.
3. Construct chain with linearized nodes and meta informations.
4. Feed the chain into rotor/pofo sovler
5. Annotate graph with the solved sequence
6. codegen and run

later I will plot some results here.


colossalai/fx/codegen/activation_checkpoint_codegen.py
```diff
@@ -157,7 +157,7 @@ def _find_offload_regions(nodes: List[Node]):
     current_region = None
 
     for idx, node in enumerate(nodes):
-        if hasattr(node, 'activation_offload') and isinstance(getattr(node, 'activation_offload', False), tuple):
+        if hasattr(node, 'activation_offload') and isinstance(getattr(node, 'activation_offload', False), list):
```
FrankLeeeee, line 160: `isinstance(getattr(node, 'activation_offload', False), list)` seems weird to me, it seems that `getattr` will return a boolean value, but it is checking isinstance for the type list.

Cypher30: Yes the default should be None.
BTW, I change this from tuple to list to simplify the annotation.

FrankLeeeee (approved):

PR #1644, opened 2022-09-26
[feature] A new ZeRO implementation

1SAA:
# _A New ZeRO Implementation_

This PR is about a new kind of implementation of ZeRO.

## Backgrounds

In the current version, our ZeRO has a performance issue. The reason is that our asymmetric distribution of chunks makes one process hinder other processes when reading the content of the chunk located in the CPU memory. It really prolongs the time of the data transmission and undermines the efficiency of ZeRO.

## Implementation

In order to solve this problem, I refactored the class Chunk. The new chunk is distributed evenly to all processes. All processes can move the data from the CPU memory to the CUDA memory in the same time. Furthermore, I provide an option to enable the pin memory for chunks. Now all chunks can have a copy in the pinned CPU memory. The above optimizations prominently improved the efficiency of data movements between CPU and CUDA.

## Another Advantage

The new ZeRO supports the true hybrid parallelism. It creates different chunk groups for parameters which have different DP communication groups. This brings huge flexibility to our up coming automatic configuration of parallelism.


colossalai/gemini/chunk/chunk.py
```diff
@@ -306,46 +398,56 @@ def copy_tensor_to_chunk_slice(self, tensor: torch.Tensor, data_slice: torch.Ten
...
+            self.chunk_total.copy_(friend_chunk.chunk_total)
+            self.optim_sync_flag = True
+        elif friend_chunk.device_type == 'cuda' and self.device_type == 'cuda':
+            self.cuda_shard.copy_(friend_chunk.cuda_shard)
+            self.optim_sync_flag = True
+            self.cpu_vis_flag = False
         else:
-            assert self.cuda_shard is not None    # only check in CUDA
-            valid_tensor = self.cuda_shard[:self.valid_end]
+            assert friend_chunk.device_type == 'cpu'
```
ver217, line 435: Please add some comments to explain where the fp16 chunk will be updated when they are on CPU.

colossalai/zero/zero_optimizer.py
```diff
@@ -61,11 +58,20 @@ def __init__(self,
...
-        self.fp16_param_to_fp32_param: Dict[torch.Tensor, torch.Tensor] = {}
+        self.param_to_range: Dict[Parameter, Tuple[int, int]] = dict()
+        self.param_to_chunk32: Dict[Parameter, Chunk] = dict()
+        self.chunk16_set: Set[Chunk] = set()
+
         for p, fp32_p in zip(module.parameters(), module.fp32_params):
-            self.fp16_param_to_fp32_param[p] = fp32_p
+            chunk_16 = self.chunk_manager.get_chunk(p)
+            chunk_32 = self.chunk_manager.get_chunk(fp32_p)
+            chunk_32.init_pair(chunk_16)
```
ver217, line 70: Move this ZeroDDP

colossalai/zero/zero_optimizer.py
```diff
@@ -138,45 +155,14 @@ def step(self, *args, **kwargs):
...
-            total_norm = total_norm**(1 / norm_type)
-        return total_norm
-
     def clip_grad_norm(self, model: torch.nn.Module, max_norm: float, norm_type: float = 2.0):
-        if self.optim_state == OptimState.SCALED:
-            self._unscale_grads()
-        total_norm = self.compute_grad_norm(norm_type)
-        _clip_grad_norm(self.module.parameters(), max_norm, total_norm)
-        return total_norm
+        raise NotImplementedError
```
ver217, line 165: Is it difficult to implement grad clipping?

colossalai/zero/zero_optimizer.py
```diff
@@ -197,24 +183,31 @@ def _maybe_move_fp32_params(self):
...
-                    # stores grad now
-                    self.chunk_manager.move_chunk(fp16_param_chunk, get_current_device())
-                    self.module._set_chunk_grad_device(fp16_param_chunk, get_current_device())
-                    fp32_params_used_cuda_margin_mem += fp32_param_chunk.mem
-                    for p in fp16_param_chunk.get_tensors():
-                        state = self.optim.state[p]
+
+            for group in self.param_groups:
+                for fake_param in group['params']:
+                    chunk32 = self.param_to_chunk32[fake_param]
```
ver217, line 189: Shall we further filter those parameters with no grad? As we assume that it's a static graph, those parameters are not required to optimized.

tests/test_gemini/update/test_zeroddp_state_dict.py
```diff
@@ -0,0 +1,114 @@
...
+    chunk_manager = ChunkManager(config_dict)
+    gemini_manager = GeminiManager(placement_policy, chunk_manager)
+    model = ZeroDDP(model, gemini_manager, pin_memory=True)
+    model.train()
+
+    zero_dict = model.state_dict(only_rank_0=False)
+    torch_dict = torch_model.state_dict()
+
+    for key, value in torch_dict.items():
+        if key == 'model.lm_head.weight':
```
ver217, line 50: Why skip this parameter?

tests/test_gemini/update/test_zerooptim_state_dict.py
```diff
@@ -0,0 +1,81 @@
...
+        if i > 0:
+            break
+        optim.zero_grad()
+        logits = model(input_ids, attn_mask)
+        logits = logits.float()
+        loss = criterion(logits, input_ids)
+        optim.backward(loss)
+
+    optim_state_dict = optim.state_dict()
+    optim.load_state_dict(optim_state_dict)
```
ver217, line 63: I think this test doesn't check the correctness of state dict now, right?

tests/test_tensor/test_zero_optim.py
```diff
@@ -6,7 +6,7 @@
 from colossalai.utils.cuda import get_current_device
```
ver217, line 6: What is the difference between this file and `test_optim.py` in `test_gemini/update`?

tests/test_gemini/update/test_zerooptim_state_dict.py
1SAA: Just make sure the function 'state_dict' can be called.

tests/test_tensor/test_zero_optim.py
1SAA: It test tp, dp hybrid parallelism.

tests/test_gemini/update/test_zeroddp_state_dict.py
1SAA: Since it is used twice, it is omitted as 'model.lm_head.weight'.

colossalai/zero/zero_optimizer.py
1SAA: Do you mean unused parameters? In my opinion, a parameter should have its gradient.

colossalai/zero/zero_optimizer.py
1SAA: I wonder whether we should calculate the norm of the paramters in GPU.

ver217: You said calculating grad norm on CPU is slow, right?

colossalai/zero/zero_optimizer.py
ver217: When finetuning, some parameters may be frozen. They are used during forward pass, but they have no grad.

tests/test_gemini/update/test_zerooptim_state_dict.py
ver217: So shall we further ensure that the `state_dict` here is the same as original torchs'?

tests/test_tensor/test_zero_optim.py
ver217: Can you rename it to make its functionality clear?

tests/test_gemini/update/test_zerooptim_state_dict.py
1SAA: ok.

colossalai/zero/zero_optimizer.py
1SAA: right.

ver217: So why not calculating grad norm on CUDA?

1SAA: ok. let's adopt this feature in the next pr.

tests/test_tensor/test_zero_optim.py
1SAA: ok.

colossalai/zero/zero_optimizer.py
1SAA: currently, it doesn't support this feature. but users can set the learning rate of these parameters to 0 instead.

ver217 (approved):

PR #1652, opened 2022-09-26
[autoparallel] remove no strategy nodes

YuliangLiu0306:


colossalai/auto_parallel/solver/cost_graph.py
```diff
@@ -27,6 +29,15 @@ def __init__(self, leaf_strategies, simplify=True):
...
         self._build_cost_graph()
 
+    def _remove_invalid_node(self, node, attr_name):
+        remove_list = []
+        target_node_list = getattr(node, attr_name, None)
+        for target_node in target_node_list:
+            if target_node not in self.nodes:
+                remove_list.append(target_node)
+        for element in remove_list:
+            target_node_list.remove(element)
```
FrankLeeeee, line 39: If this node does not have `attr_name`, `target_node_list` will be None, and None cannot be iterated.

colossalai/auto_parallel/solver/cost_graph.py
```diff
@@ -83,7 +98,7 @@ def merge_node(self, src_node, dst_node):
         # build merge_map
         merge_map = {}
         for src_index, strategy in enumerate(src_node.strategies_vector):
-            min_cost = math.inf
+            min_cost = INFINITY_COST + 1
```
FrankLeeeee, line 101: Why plus one?

YuliangLiu0306: The highest reshard cost is INFINITY_COST, so it could make sure at least one strategy could be picked.

FrankLeeeee (approved):

PR #1659, opened 2022-09-27
[hotfix] fix singledispatchmethod incompatibility.

super-dainiu:
### Hotfix
`singledispatchmethod` in functool is incompatible with Python 3.7. According to [this](https://pypi.org/project/singledispatchmethod/), we can solve this problem.


super-dainiu:
I need to examine all compatibility issues.

PR #1660, opened 2022-09-27
Fix amp test bf16

ht-zhou:
fix test_naive_bf16.py


binmakeswell:
Hi @ht-zhou Is this PR still active?

PR #1662, opened 2022-09-27
[autoparallel] added compute resharding costs for node handler

FrankLeeeee:
This PR added the method `update_resharding_cost` to node handler and the node handler can support generating the sharding while registering sharding strategies. This PR cannot be tested at this moment, I have added a `TODO` note to add it back when other handlers are finished.


colossalai/auto_parallel/solver/op_handler/node_handler.py
```diff
@@ -28,13 +29,53 @@ def __init__(
...
+            # create data structrure to store costs
+            if op_data not in resharding_costs:
+                resharding_costs[op_data] = {}
+
+            # for each sharding spec generated by the predecessor's node handler
+            # compute the resharding cost to switch to the sharding spec generated
+            # by the current node handler
+            for prev_sharding_spec in prev_sharding_specs:
+                fwd_cost = shape_consistency_manager.shape_consistency(prev_sharding_spec, current_sharding_spec)
+                bwd_cost = shape_consistency_manager.shape_consistency(current_sharding_spec, prev_sharding_spec)
```
YuliangLiu0306, line 62: ShapeConsistencyManager will distinguish the mission type(training or inference) by a forward_only label, and returns a option related cost. Therefore, we don't have to compute fwd cost and bwd cost separately here.

FrankLeeeee: I see. I notice that ShapeConsistencyManager only returns a total cost. How to get the fwd and bwd cost then?

YuliangLiu0306: image

FrankLeeeee: I mean I want to store this cost as a TrainCycleItem so that solver can choose whether to consider the bwd cost.

YuliangLiu0306 (approved):

PR #1663, opened 2022-09-27
[autoparallel] add conv handler v2

YuliangLiu0306:


colossalai/auto_parallel/solver/strategy/conv_strategy_generator.py
```diff
@@ -0,0 +1,491 @@
...
+                communication_pattern=CollectiveCommPattern.IDENTITY_FWD_ALLREDUCE_BWD,
+                logical_process_axis=mesh_dim_0)
+            communication_action_mapping["other"] = other_comm_spec
+
+        if self.has_bias and self.is_param("bias"):
+            bias_comm_spec = self.get_communication_spec(
+                sharding_spec_mapping["bias"],
+                communication_pattern=CollectiveCommPattern.IDENTITY_FWD_ALLREDUCE_BWD,
+                logical_process_axis=mesh_dim_0)
+            communication_action_mapping["bias"] = bias_comm_spec
```
FrankLeeeee, line 179: Why bias will be all-reduced? bias is only involved in addition not multiplication.

YuliangLiu0306: It is a dp style sharding strategy, I think we should all reduce all the parameters grad during backward.

FrankLeeeee: Oh right, I think this is missing in the linear handler too. I will get to PR to fix that. Thx.

FrankLeeeee (approved):

PR #1673, opened 2022-09-30
[autoparallel] add following node generator

YuliangLiu0306:
### What is following strategies node
Some nodes like linear or conv generate strategies by the computation rules, and the input arguments have to do shape consistency to join the strategies. However, some nodes like unary elementwise kernel are not expected to change the predecessor ndoe strategy, so their strategy is just following the input's strategy. We call those node as "following strategies node"
### What does this PR do
1. To build following strategies node generator, node is added as an attribute into FollowingStrategyGenerator.
2. This PR adds handler for operator.getitem. There are mainly three use cases:
     - args_0._meta_data: torch.Tensor, args_1._meta_data: int
     - args_0._meta_data: torch.Tensor, args_1._meta_data: slice
     - args_0._meta_data: Tuple[torch.Tensor], args_1._meta_data: int
     - All above cases are covered in this PR.
3. Fix some previous bugs of other handlers.


FrankLeeeee:
This PR has a conflicted file, please fix the conflict first.

colossalai/auto_parallel/solver/op_handler/getitem_handler.py
```diff
@@ -0,0 +1,39 @@
...
+    def get_operation_data_mapping(self) -> Dict[str, OperationData]:
+        # use transposed shape for strategies
+        # the strategies will be transformed back to its original shape in self.post_process
+        physical_input_operand = OperationData(name=str(self.node.args[0]),
+                                               type=OperationDataType.ARG,
+                                               data=self.node.args[0]._meta_data)
+        physical_other_operand = OperationData(name="index", type=OperationDataType.ARG, data=self.node.args[1])
+        physical_output = OperationData(name=str(self.node), type=OperationDataType.OUTPUT, data=self.node._meta_data)
+
+        mapping = {"input": physical_input_operand, "other": physical_other_operand, "output": physical_output}
```
FrankLeeeee, line 37: the operand name should not be called `other`, it should be called `index` instead.

YuliangLiu0306: Sure, I will fix it.

colossalai/auto_parallel/solver/op_handler/getitem_handler.py
```diff
@@ -0,0 +1,39 @@
...
+@operator_registry.register(operator.getitem)
+class GetItemHandler(NodeHandler):
+    """
+    A GetItemHandler which deals with the sharding strategies for operator.getitem.
+    """
+
+    def get_strategy_generator(self) -> List[StrategyGenerator_V2]:
+        op_data_mapping = self.get_operation_data_mapping()
+        generators = []
+        if isinstance(self.node.args[0]._meta_data, torch.Tensor):
```
FrankLeeeee, line 21: `self.node.args[0]._meta_data` -> `op_data_mapping['other'].data

colossalai/auto_parallel/solver/op_handler/getitem_handler.py
FrankLeeeee: Is it because that the strategy generator requires it to be called `other`?

FrankLeeeee (approved):

YuliangLiu0306:
image

PR #1658, opened 2022-09-27
[autoparallel] add rotor C version

Cypher30:
# What's New?

In this PR I add C extension of rotor dynamic programming table calculation and remove repeated `MetaInfoProp` in rotor solver, right now you should execute the following code before running solver rotor

```python
if META_COMPATIBILITY:
    data_meta = MetaTensor(data, fake_device=next(gm.parameters()).device)
    MetaInfoProp(gm).run(data_meta)
```

As we might use rotor for simplified offload strategy, this modification is crucial as we might run solver over the same graph several time for auto capper heuristic, we don't need to run this MetaInfoProp over and over again. Actually with the new C version dynamic programming, the `MetaInfoProp` itself consumes more time than simply run the dynamic programming.

PR #1680, opened 2022-10-06
[autoparallel] fix insecure subprocess

Cypher30:
# What's New?

In the previous PR of C rotor version #1658, we built C extension with `subprocess.Popen` with `shell=True`, however it might cause some serious problems by code injection, therefore, I remove it this PR and modify it in the form of `shell=False`.


super-dainiu (approved):

colossalai/fx/passes/algorithms/ckpt_solver_rotor.py
```diff
@@ -353,10 +353,10 @@ def solver_rotor(gm: ColoGraphModule,
             logger.info("dynamic_programs_C_version hasn't been built! Building library...", ranks=[0])
             this_dir = os.path.dirname(os.path.abspath(__file__))
             result = subprocess.Popen(
-                f'python {os.path.join(this_dir, "build_c_ext.py")} build_ext --build-lib={this_dir}',
+                ["python", f"{os.path.join(this_dir, 'build_c_ext.py')}", "build_ext", f"--build-lib={this_dir}"],
```
FrankLeeeee, line 356: You should not use `python`. To be safe, you should use `sys.executable` to give the absolute path to the python bin.

Cypher30: okay

FrankLeeeee (approved):

PR #1688, opened 2022-10-11
[autoparallel] where_handler_v2

YuliangLiu0306:
### What does PR do?
1. simply, this PR add where_handler using refactored APIs.
2. In this PR, I overload post_process and register_strategy function to recover the logical strategy into physical strategy. The cost model rule is:
- using logical shape in compute_cost, memory_cost, and communication_cost computaion.
- using physical shape in resharding_cost computation.
Note: This rule may miscalculate some costs due to different kernel implement method, but it works fine for most of cases.


FrankLeeeee:
This PR has a conflict to be resolved.

FrankLeeeee (approved):

PR #1679, opened 2022-10-05
[fx/profiler] assigned UUID to each unrecorded tensor/ improved performance on GPT-2

super-dainiu:
### What's new?
1. Provide a function to calculate `fwd_in`, `fwd_tmp` and `fwd_out`. Hopefully, this will avoid a lot of misunderstandings.
2. Assigned UUID to each unrecorded tensor.
3. Improved performance on GPT-2

### Bugs in existing code
1. C version of the rotor solver is not correct.  cc @Cypher30 

C version results

Model | mem_limit | real_consumption | train step time | solver time
-- | -- | -- | -- | --
<function densenet121 at 0x7feff4622430> | mem_limit: None | real memory consumption: 15996.476 MB | train step time: 171.204 MS |  
<function densenet121 at 0x7feff4622430> | mem_limit: 5300.0 MB | real memory consumption: 5300.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 6680.0 MB | real memory consumption: 6680.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 8060.0 MB | real memory consumption: 8060.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 9440.0 MB | real memory consumption: 9440.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 10820.0 MB | real memory consumption: 10820.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 12200.0 MB | real memory consumption: 12200.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 13580.0 MB | real memory consumption: 13580.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 14960.0 MB | real memory consumption: 14960.000 MB | train step time: inf MS | solver time: 0.000 MS
<function densenet121 at 0x7feff4622430> | mem_limit: 16340.0 MB | real memory consumption: 15800.507 MB | train step time: 171.508 MS | solver time: 2038.141 MS
[...]

PR #1691, opened 2022-10-12
[autoparallel] fix C version rotor inconsistency

Cypher30:
# What's New?

Previously in PR #1679 , @super-dainiu reported the inconsistency between C version and python version of rotor solver, I dived into the C source code and found that the problem is that the C version compute the value in different order, with older version of profiler, the computation order will not damage the result, but currently some of the output of profiler has memory size of 0 (depends on newer version of graph analysis, some of the operation will discard the input), thus, the different order will cause solver failure. In this PR, I fix this problem and now the C version is consistent with python version.


colossalai/fx/passes/algorithms/ckpt_solver_rotor.py
```diff
@@ -406,9 +417,18 @@ def solver_rotor(gm: ColoGraphModule,
...
 
     # set __sequence__ attribute to GraphModule
-    setattr(gm, "__sequence__", sequence)
+    if SOLVER_FAILED:
+        setattr(gm, "__sequence__", None)
+    else:
+        setattr(gm, "__sequence__", sequence)
+
+    # set __opttable__ attribute to GraphModule
+    setattr(gm, "__opttable__", opt_table[0])
```
super-dainiu, line 432: why `opttable[0]`?

Cypher30: because I think don't need `what` table to be attached to the graph module, the returned `opt_table[0]` by solver is a table containing `opt_table[0] = opt, opt_table[1] = what`, and I think we only need `opt` to check the value.

super-dainiu: I see

super-dainiu (approved):

PR #1724, opened 2022-10-18
reformat all codes by autoflake and isort

1SAA:
* add autoflake and isort in pre-commit
* clean all codes


FrankLeeeee:
Don't change all files in this PR. You should only change the precommit config. File formatting should be done incrementally.

FrankLeeeee:
You don't need to use flake, instead you can use https://github.com/hadialqattan/pycln to remove unused imports.

PR #1729, opened 2022-10-18
[zero] add chunk init function for users

1SAA:


colossalai/gemini/chunk/search_utils.py
```diff
@@ -1,100 +1,106 @@
...
+from typing import Dict, List, Tuple
+
+import numpy as np
+import torch.nn as nn
+
+from colossalai.tensor import ColoParameter
+
+
+def in_ddp(param: nn.Parameter) -> bool:
+    return not getattr(param, '_ddp_to_ignore', False)
```
feifeibear, line 11: what if the param has no attr named as _ddp_to_ignore

colossalai/gemini/chunk/search_utils.py
```diff
@@ -1,100 +1,106 @@
...
+
+
+def _filter_exlarge_params(model: nn.Module, size_dict: Dict[int, List[int]]) -> None:
+    """Filter those parameters whose size is too large from others.
+    """
+    params_size = [p.numel() for p in model.parameters() if in_ddp(p)]
+    params_size_arr = np.array(params_size)
+
+    std = np.std(params_size_arr)
```
feifeibear, line 20: why not use torch.std? is it necessary to conver it to np?

colossalai/gemini/chunk/search_utils.py
```diff
@@ -1,100 +1,106 @@
...
+
+from colossalai.tensor import ColoParameter
+
+
+def in_ddp(param: nn.Parameter) -> bool:
+    return not getattr(param, '_ddp_to_ignore', False)
+
+
+def _filter_exlarge_params(model: nn.Module, size_dict: Dict[int, List[int]]) -> None:
+    """Filter those parameters whose size is too large from others.
```
feifeibear, line 15: add more comments. How to define too large than others!

colossalai/gemini/chunk/search_utils.py
```diff
@@ -1,100 +1,106 @@
...
+    std = np.std(params_size_arr)
+    mean = np.mean(params_size_arr)
+    upper_limit = mean + 3 * std
+
+    for key in size_dict:
+        org_list = size_dict[key]
+        size_dict[key] = list(filter(lambda x: x <= upper_limit, org_list))
+
+
+def _get_unused_byte(size_list: List[int], chunk_size: int) -> int:
```
feifeibear, line 29: are you sure the return value is in byte not element number?

colossalai/gemini/chunk/search_utils.py
1SAA: returns a False by default.

colossalai/gemini/chunk/search_utils.py
```diff
@@ -1,100 +1,106 @@
...
+    left = 0
+    for s in size_list:
+        if s > left:
+            acc += left
+            left = chunk_size
+        left -= s
+    return left + acc
+
+
+def clasify_params(model: nn.Module) -> Dict[int, List[ColoParameter]]:
```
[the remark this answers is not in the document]
feifeibear: please add comments or change a more straightforward name. Classify params w.r.t. their DP degree?

colossalai/gemini/chunk/search_utils.py
1SAA: I don't know. @ver217 wrote this code.

colossalai/gemini/chunk/search_utils.py
1SAA: 100% sure is not in byte. here byte is used to represent a minimum unit.

colossalai/gemini/chunk/utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+    config_dict, wasted_size = search_chunk_configuration(model, **kwargs_dict)
+
+    dist.barrier()
+    end = time()
+    span_s = end - begine
+    wasted_size /= 1024**2
+
+    if dist.get_rank() == 0:
+        print("searching chunk configuration is completed in {:.2f} s.\n".format(span_s),
```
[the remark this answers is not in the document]
1SAA: print_rank_0 is not robust enough to print multiple strings. it is only able to print a single string.

PR #1735, opened 2022-10-18
[autoparallel] add sequential order to communication actions

YuliangLiu0306:
### What does this PR do
1. This PR add CommAction class, which shows **the sequential order** of a communication action:
-  **BEFORE**: the communication action happens just before the computation operation.
-  **AFTER**: the communication action happens after the computation operation.
-  **HOOK**:  the communication action is used to do the grad all reduce.
-  **IMPLICIT**: the communication action happens during the kernel execution, such as SyncBatchNorm
2. This PR update **the conv handler and reshape handler with the above class**. **Other handlers will be updated in future PR**.
3. Build a unit test to check **the training cycle** of a simple convolution model.


colossalai/auto_parallel/tensor_shard/sharding_strategy.py
```diff
@@ -84,6 +85,23 @@ class MemoryCost:
...
+    AFTER = 1
+    HOOK = 2
+    IMPLICIT = 3
+
+
+@dataclass
+class CommAction:
+    comm_spec: CommSpec = None
+    comm_type: CommType = None
+    arg_index: int = -1
```
FrankLeeeee, line 102: What is `arg_index`?

colossalai/auto_parallel/tensor_shard/sharding_strategy.py
```diff
@@ -84,6 +85,23 @@ class MemoryCost:
...
+    CommType describes the sequential order of a communication action and a computation action.
+    """
+    BEFORE = 0
+    AFTER = 1
+    HOOK = 2
+    IMPLICIT = 3
+
+
+@dataclass
+class CommAction:
```
FrankLeeeee, line 99: Add a docstring for the dataclass.

colossalai/auto_parallel/tensor_shard/sharding_strategy.py
```diff
@@ -84,6 +85,23 @@ class MemoryCost:
     buffer: int = 0
 
 
+class CommType(Enum):
+    """
+    CommType describes the sequential order of a communication action and a computation action.
+    """
```
FrankLeeeee, line 91: Add a docstring to explain what each value means.

colossalai/auto_parallel/tensor_shard/sharding_strategy.py
YuliangLiu0306: Okay

colossalai/auto_parallel/tensor_shard/sharding_strategy.py
YuliangLiu0306: I will add a comment to explain it

FrankLeeeee: Why `arg_index` is needed? If the args of the node are changed, we should update the sharding strategy data accordingly instead of bookkeeping its order.

YuliangLiu0306:
image

FrankLeeeee (approved):

PR #1757, opened 2022-10-24
[autoparallel] refactor the runtime apply pass and add docstring to passes

YuliangLiu0306:
### What does this PR do
1. This PR refactor the runtime pass into two files(runtime_apply_pass.py and runtime_preparetion_pass.py), **runtime apply pass manages all the passes whose actions are used more than one time during training iters**, while **the runtime preparation pass manages all the passes whose actions are just used once during runtime preparation phase**.
2. Add docstring to passes.


colossalai/fx/passes/autoparallel/runtime_preparation_pass.py
```diff
@@ -0,0 +1,130 @@
...
+                setattr(target_module, name, buffer_sharded.detach().clone())
+
+    return gm
+
+
+def implicit_comm_action_apply_pass(gm: torch.fx.GraphModule):
+    """
+    replace the origin kernel into kernel with implicit communication inside.
+    """
+    pass
```
FrankLeeeee, line 120: Why pass?

colossalai/fx/passes/autoparallel/runtime_preparation_pass.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def runtime_preparation_pass(gm: torch.fx.GraphModule, solution: List[int], device_mesh: DeviceMesh):
+    gm, sharding_spec_convert_dict, origin_node_sharding_spec_dict, comm_actions_dict = _solution_annotatation_pass(
+        gm, solution)
+    # TODO: the pass below should be uncommented after the implementation of implicit_comm_action_apply_pass completed.
+    # gm = implicit_comm_action_apply_pass(gm)
+    gm = _module_params_sharding_pass(gm, device_mesh)
```
FrankLeeeee, line 128: The function names seem a bit weird to me, it is kind of saying that there are passes within passes as `runtime_preparation_pass` calls `_solution_annotatation_pass` and `_module_params_sharding_pass`. Would it better if we treat `_module_params_sharding_pass` and `_solution_annotatation_pass` as the helper functions instead of passes.

colossalai/fx/passes/autoparallel/runtime_apply_pass.py
```diff
@@ -0,0 +1,151 @@
...
+    OperationDataType,
+)
+from colossalai.device.device_mesh import DeviceMesh
+from colossalai.tensor.comm_spec import CommSpec
+from colossalai.tensor.shape_consistency import ShapeConsistencyManager
+
+shape_consistency_manager = ShapeConsistencyManager()
+
+
+def runtime_apply(node: Node, origin_dict: Dict, input_dict: Dict, node_index: int, user_node_index: int):
```
FrankLeeeee, line 20: I think we should keep the passes in the `auto_parallel` module instead of the `fx` module. The relationship should be that `auto_parallel` module depends on the `fx` module. The current relationship is that auto parallel solver relies on fx tracing, and then fx passes relies on the auto parallel sharding strategy, which leads to high coupling between these modules.

colossalai/fx/passes/autoparallel/runtime_preparation_pass.py
YuliangLiu0306: This pass will be implemented in later PR.

colossalai/fx/passes/autoparallel/runtime_preparation_pass.py
YuliangLiu0306: I see.

colossalai/fx/passes/autoparallel/runtime_apply_pass.py
YuliangLiu0306: Okay

FrankLeeeee (approved):

robotsp:
hi @YuliangLiu0306 @FrankLeeeee ,

I am using colossal-auto parallel module to train a model in an efficient way. I check the profiling data, in each step, the runtime apply are called for shape consistency which harms performance too much. Do we have any solutions to optimize the process, I think shape consistency should called only once in trace?

PR #1760, opened 2022-10-25
[autoparallel] add numerical test for node strategies

YuliangLiu0306:
### What does this PR do

This PR provides some apis to do numerical tests for node strategies. The function called numerical_test_for_node_strategy will check the **output of forward result, the gradient of the input activation and gradient of parameters if exist** with the original result of non-distributed version.

Note: only convolution handler tests updated with the above apis as an example, other handlers unit test will be updated in next PR.

### DIFF
--- a/colossalai/auto_parallel/passes/runtime_apply_pass.py
+++ b/colossalai/auto_parallel/passes/runtime_apply_pass.py
@@ -24,7 +24,6 @@ def runtime_apply(node: Node, origin_dict: Dict, input_dict: Dict, node_index: i
     """
     origin_sharding_spec = origin_dict[node_index]
     target_sharding_spec = input_dict[node_index][user_node_index]
-
     return shape_consistency_manager.apply_for_autoparallel_runtime(node, origin_sharding_spec, target_sharding_spec)
 
 
@@ -81,18 +80,22 @@ def _shape_consistency_apply(gm: torch.fx.GraphModule):
         if not hasattr(node, 'best_strategy') or node.op == 'output':
             continue
 
-        for user_node in node.strategies_vector.successor_nodes:
-            user_node_index = user_node.strategies_vector.predecessor_nodes.index(node)
+        for user_node_index, user_node in enumerate(node.strategies_vector.successor_nodes):
+            # user_node_index = user_node.strategies_vector.predecessor_nodes.index(node)
             with mod_graph.inserting_before(user_node):
                 shape_consistency_node = mod_graph.create_node('call_function',
                                                                runtime_apply,
                                                                args=(node, origin_dict_node, input_dict_node,
                                                                      node_to_index_dict[node], user_node_index))
-
-            origin_index_args = user_node.args.index(node)
             new_args = list(user_node.args)
-            new_args[origin_index_args] = shape_consistency_node
-            user_node.args = new_args
+            new_kwargs = dict(user_node.kwargs)
+            if node in new_args:
+                origin_index_args = new_args.index(node)
+                new_args[origin_index_args] = shape_consistency_node
+                user_node.args = new_args
+            elif str(node) in new_kwargs:
+                new_kwargs[str(node)] = shape_consistency_node
+                user_node.kwargs = new_kwargs
 
     return gm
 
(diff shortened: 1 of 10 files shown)


colossalai/auto_parallel/passes/runtime_apply_pass.py
```diff
@@ -81,18 +80,22 @@ def _shape_consistency_apply(gm: torch.fx.GraphModule):
         if not hasattr(node, 'best_strategy') or node.op == 'output':
             continue
 
-        for user_node in node.strategies_vector.successor_nodes:
-            user_node_index = user_node.strategies_vector.predecessor_nodes.index(node)
+        for user_node_index, user_node in enumerate(node.strategies_vector.successor_nodes):
+            # user_node_index = user_node.strategies_vector.predecessor_nodes.index(node)
```
FrankLeeeee, line 84: remove this line?

colossalai/auto_parallel/tensor_shard/sharding_strategy.py
```diff
@@ -115,6 +115,7 @@ class CommAction:
     comm_spec: CommSpec = None
     comm_type: CommType = None
     arg_index: int = -1
+    key_for_kwarg: any = None
```
FrankLeeeee, line 118: what is this used for?

colossalai/auto_parallel/passes/runtime_apply_pass.py
```diff
@@ -112,18 +115,29 @@ def _comm_spec_apply(gm: torch.fx.GraphModule):
...
-                node.args = new_args
+
+                if comm_action.key_for_kwarg is not None:
+                    new_kwargs = dict(node.kwargs)
+                    new_kwargs[comm_action.key_for_kwarg] = comm_spec_apply_node
+                    node.kwargs = new_kwargs
+                else:
+                    new_args = list(node.args)
+                    new_args[comm_action.arg_index] = comm_spec_apply_node
+                    node.args = new_args
```
FrankLeeeee, line 139: It is kind of difficult to understand what is going on here. Can you write some comments (not just here, but other places as well)?

tests/test_auto_parallel/test_tensor_shard/test_node_handler/utils.py
```diff
@@ -0,0 +1,137 @@
+import copy
+from re import T
```
FrankLeeeee, line 2: I think this import is not used.

colossalai/auto_parallel/tensor_shard/sharding_strategy.py
YuliangLiu0306: If a node is kwarg of other node, not a positional argument, we could use this to locate it in runtime pass.

colossalai/auto_parallel/passes/runtime_apply_pass.py
YuliangLiu0306: Sure

tests/test_auto_parallel/test_tensor_shard/test_node_handler/utils.py
YuliangLiu0306: I will remove these imports.

tests/test_auto_parallel/test_tensor_shard/test_node_handler/utils.py
```diff
@@ -0,0 +1,137 @@
...
+            assert_close((global_grad - grad_to_compare).sum(), zero_tensor)
+
+
+def to_global(distributed_tensor: torch.Tensor, sharding_spec: ShardingSpec):
+    shape_consistency_manager = ShapeConsistencyManager()
+    global_sharding_spec = ShardingSpec(sharding_spec.device_mesh, sharding_spec.entire_shape, {})
+    with torch.no_grad():
+        global_tensor = shape_consistency_manager.apply_for_autoparallel_runtime(distributed_tensor, sharding_spec,
+                                                                                 global_sharding_spec)
+    return global_tensor
```
FrankLeeeee, line 137: This method should be put in the `shariding_spec.py` so that it can be possibly used by other files in the future.

colossalai/auto_parallel/passes/runtime_apply_pass.py
YuliangLiu0306: Sure

tests/test_auto_parallel/test_tensor_shard/test_node_handler/utils.py
YuliangLiu0306: Okay, I will move it into sharding_spec.py.

tests/test_auto_parallel/test_tensor_shard/test_node_handler/utils.py
```diff
@@ -0,0 +1,137 @@
...
+        wrapper(kwarg_to_compare, name)
+        kwargs_to_compare[name] = kwarg_to_compare
+
+    return model_to_compare, args_to_compare, kwargs_to_compare
+
+
+def numerical_test_for_node_strategy(model: torch.nn.Module,
+                                     device_mesh: DeviceMesh,
+                                     node_index: int,
+                                     strategy_number: int,
```
FrankLeeeee, line 59: can this number be set to 1 first and be automatically updated after calling `build_strategies_and_cost`?

FrankLeeeee: so that we don't have to manually count how many strategies to test.

YuliangLiu0306: Yes, but I think strategy number should be a consistent for unit tests.

FrankLeeeee: Okay, one possible omission is that the tester gives a wrong number (e.g. lower than the actual number), then the test is not complete.

FrankLeeeee (approved):

PR #1763, opened 2022-10-26
[autoparallel] added matmul handler

FrankLeeeee:
# What is the problem?

We don't have a node handler to deal with `torch.matmul` operation. This makes it tough to parallelize modules like the self-attention module.

# What does this PR do?

This PR implemented the matmul node handler to cover the strategy generation for all compute cases. This node handler has been tested against more than 10 test cases to ensure correctness.

# Summary of Code Change

- `sharding_spec.py`: the `ShardingSpec` class's `__init__` is updated such that it will convert `list` to `torch.Size` when `entire_size` is a list of integers.
- `matmul_strategy_generator.py`: fixed some bugs in the strategy and added the `ignore_sharding_exception` decorator.
- `test_matmul_handler.py`: test for all dot/matrix-matrix/matrix-vector/batched-matrix multiplication.
- `matmul_handler.py`: this file contains the main change in this PR. These changes will be discussed below.

What is done is matmul_handler?
1. The types of matmul operation is enumerated by `MatMulType` as stated in [`torch.matmul` documentation](https://pytorch.org/docs/stable/generated/torch.matmul.html).
2. For different `MatMulType`, the node handler will use a different strategy generator.
3. The `BMM` case is the most complicated as it can possibly involve padding, broadcasting and view operations. To simplify the implementation, I used pipeline design to abstract them into the `BmmTransform` class. Each operation inherits this class and implements the `apply` and `recover` methods. These methods will be used to convert the logical and physical shape and sharding spec. The overall logical flow is given in the diagram below.
Screen Shot 2022-10-26 at 15 18 16

### DIFF
--- a/colossalai/auto_parallel/tensor_shard/node_handler/strategy/matmul_strategy_generator.py
+++ b/colossalai/auto_parallel/tensor_shard/node_handler/strategy/matmul_strategy_generator.py
@@ -60,12 +60,13 @@ class DotProductStrategyGenerator(MatMulStrategyGenerator):
     def update_compute_cost(self, strategy: ShardingStrategy) -> ShardingStrategy:
         sharded_input_shape = strategy.sharding_specs[self.op_data['input']].get_sharded_shape_per_device()
         fwd_compute_cost = sharded_input_shape[0]
-        bwd_compute_cost = sharded_input_shape * 2
+        bwd_compute_cost = fwd_compute_cost * 2
         compute_cost = TrainCycleItem(fwd=fwd_compute_cost,
                                       bwd=bwd_compute_cost,
                                       total=fwd_compute_cost + bwd_compute_cost)
         return compute_cost
 
+    @ignore_sharding_exception
     def no_split(self):
         name = f'R = R dot R'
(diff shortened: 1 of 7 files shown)


colossalai/auto_parallel/tensor_shard/node_handler/strategy/matmul_strategy_generator.py
```diff
@@ -759,8 +788,8 @@ def collate_strategies(self) -> List[ShardingStrategy]:
             strategy_list.append(self.split_batch_dim_lhs_space(0, 1))
             strategy_list.append(self.split_batch_dim_lhs_space(1, 0))
 
-            # split batch dim of two inputs and the j of the second tensor
-            # SbSj = Sb x SbSj
+            # # split batch dim of two inputs and the j of the second tensor
+            # # SbSj = Sb x SbSj
```
YuliangLiu0306, line 792: Remove the second '#'.

FrankLeeeee: removed.

YuliangLiu0306 (approved):

PR #1767, opened 2022-10-27
[autoparallel] add getattr handler

YuliangLiu0306:
This PR add strategy handler for **node with 'get_attr' op type.**


colossalai/auto_parallel/tensor_shard/node_handler/getatrr_hadnler.py
```diff
@@ -0,0 +1,28 @@
...
+    def get_strategy_generator(self) -> List[StrategyGenerator]:
+        op_data_mapping = self.get_operation_data_mapping()
+        generators = []
+        generators.append(GetattrGenerator(op_data_mapping, self.device_mesh))
+        return generators
+
+    def get_operation_data_mapping(self) -> Dict[str, OperationData]:
+        # use transposed shape for strategies
+        # the strategies will be transformed back to its original shape in self.post_process
+        physical_output = OperationData(name=str(self.node), type=OperationDataType.OUTPUT, data=self.node._meta_data)
```
FrankLeeeee, line 24: Is it possible that the output is not a tensor?

colossalai/auto_parallel/tensor_shard/node_handler/strategy/getattr_generator.py
```diff
@@ -0,0 +1,53 @@
...
+        strategy.memory_cost = memory_cost
+
+    def collate_strategies(self) -> List[ShardingStrategy]:
+        dim_partition_dict_mapping = {
+            "output": {},
+        }
+        communication_action_mapping = {}
+        sharding_spec_mapping = self.to_sharding_spec_mapping(dim_partition_dict_mapping)
+
+        name = 'Replica Attribute'
```
FrankLeeeee, line 47: It may not be replica? If the attribute has already been assigned a sharding spec like the weight of a module, it should inherit that spec.

colossalai/auto_parallel/tensor_shard/node_handler/getatrr_hadnler.py
YuliangLiu0306: I see, It could be any type if users define some attributes with random types to module.

colossalai/auto_parallel/tensor_shard/node_handler/strategy/getattr_generator.py
YuliangLiu0306: I see, I'll fix it

YuliangLiu0306:
The commit([add extra processes for Parameters](https://github.com/hpcaitech/ColossalAI/pull/1767/commits/6f985ee4299a9b2c9b54bc8e660daefc836db092)) pays some extra attention on nn.Parameters:
1. Firstly, a nn.Parameters to meta will become a normal tensor, therefore, the correct way to do it is nn.Parameter(param.to(''meta))
2. Due to all the parameters just sharded once before runtime, the resharding costs of Parameter will not be counted during updated_resharding_cost.

colossalai/auto_parallel/tensor_shard/node_handler/getatrr_hadnler.py
YuliangLiu0306: There are only two possible types for get_attr node:
1. torch.Tensor( torch.nn.Parameters)
2. torch.nn.Module
Temporarily, we just support first case in our tracer, so we don't have to worry about issue caused by output type.

colossalai/auto_parallel/tensor_shard/node_handler/strategy/getattr_generator.py
YuliangLiu0306: We will add sharding spec during runtime preparation.

FrankLeeeee (approved):

PR #1775, opened 2022-10-31
[tensorparallel] Updated TP layers

kurisusnowdeng:
- Polished tensor parallel implementation
- Achieved the same 1D throughput as the latest Megatron

### DIFF
--- a/colossalai/context/parallel_mode.py
+++ b/colossalai/context/parallel_mode.py
@@ -39,6 +39,8 @@ class ParallelMode(Enum):
     PARALLEL_3D_INPUT = '3d_input'
     PARALLEL_3D_WEIGHT = '3d_weight'
     PARALLEL_3D_OUTPUT = '3d_output'
+    PARALLEL_3D_INPUT_X_WEIGHT = "3d_input_x_weight"
+    PARALLEL_3D_OUTPUT_X_WEIGHT = "3d_output_x_weight"
 
     # 2.5D parallel
     PARALLEL_2P5D_ROW = '2p5d_row'
--- a/colossalai/context/process_group_initializer/initializer_3d.py
+++ b/colossalai/context/process_group_initializer/initializer_3d.py
@@ -176,6 +176,112 @@ class Initializer_3D_Output(ProcessGroupInitializer):
         return local_rank, group_world_size, process_group, cpu_group, ranks_in_group, mode
 
 
+class Initializer_3D_InputxWeight(ProcessGroupInitializer):
+    """3D tensor parallel initialization among input.
+
+    Args:
+        num_group (int): The number of all tensor groups.
+        depth (int): Depth of 3D parallelism.
+        rank (int): The rank of current process.
+        world_size (int): Size of whole communication world.
+        config (Config): Running configuration.
+        data_parallel_size (int): Size of data parallel.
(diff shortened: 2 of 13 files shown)


feifeibear:
Please add some unitests!

kurisusnowdeng:
> Please add some unitests!

We already have tests for tensor parallel layers

colossalai/context/parallel_mode.py
```diff
@@ -39,6 +39,8 @@ class ParallelMode(Enum):
     PARALLEL_3D_INPUT = '3d_input'
     PARALLEL_3D_WEIGHT = '3d_weight'
     PARALLEL_3D_OUTPUT = '3d_output'
+    PARALLEL_3D_INPUT_X_WEIGHT = "3d_input_x_weight"
+    PARALLEL_3D_OUTPUT_X_WEIGHT = "3d_output_X_weight"
```
feifeibear, line 43: why 3d_output_**X** uppercase?

colossalai/context/process_group_initializer/initializer_3d.py
```diff
@@ -176,6 +176,112 @@ def init_dist_group(self):
...
+            Tuple (local_rank, group_world_size, process_group, ranks_in_group, mode):
+                3D tensor parallelism's information among input in a tuple.
+        """
+        local_rank = None
+        ranks_in_group = None
+        process_group = None
+        cpu_group = None
+        group_world_size = None
+        mode = ParallelMode.PARALLEL_3D_INPUT_X_WEIGHT
+        env.input_x_weight_group_3d = mode
```
feifeibear, line 211: is env a global var? If so, it should be in uppercase.

colossalai/nn/layer/parallel_1d/_operation.py
```diff
@@ -43,3 +45,52 @@ def backward(ctx, grad_output):
...
+        total_input = total_input.view(total_input.shape[0] * total_input.shape[1], total_input.shape[2])
+
+        if ctx.async_grad_allreduce:
+            # Asynchronous all-reduce
+            handle = dist.all_reduce(grad_input, group=gpc.get_group(ctx.parallel_mode), async_op=True)
+            # Delay the start of weight gradient computation shortly (3us) to have
+            # all-reduce scheduled first and have GPU resources allocated
+            _ = torch.empty(1, device=grad_output.device) + 1
+
+        grad_weight = grad_output.t().matmul(total_input)
```
feifeibear, line 86: Can we use total_input here? Isn't it being allreduced?

colossalai/context/process_group_initializer/initializer_3d.py
kurisusnowdeng: it is a global variable manager

colossalai/nn/layer/parallel_1d/_operation.py
kurisusnowdeng: 1d weights do not need to be allreduced

feifeibear: I see

feifeibear (approved):

colossalai/context/parallel_mode.py
kurisusnowdeng: fixed

PR #1780, opened 2022-10-31
[fx] support module with bias addition

YuliangLiu0306:
### what does this PR do

1. Patch the torch.nn.Linear module, use torch.nn.functional.linear without bias and operator.add to substitute origin computation.
2. Patch the torch.nn.Convxd module, use torch.nn.functional.convxd without bias and operator.add to substitute origin computation.

**Note**: **the bias addition for convxd can't use PyTorch broadcast mechanism** because the last dimension of conv output is not channel dimension while bias only has the out channel dimension. Therefore, **an additional view node is added** to make sure patched module working correctly.


YuliangLiu0306:
The conv2d module after substituting looks like:
image

YuliangLiu0306:
The linear module after substituting looks like:
image

FrankLeeeee:
> The linear module after substituting looks like: image

Should the target be `F.linear`?

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/conv.py
```diff
@@ -0,0 +1,53 @@
+import copy
```
FrankLeeeee, line 1: `copy` is not used.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/linear.py
```diff
@@ -0,0 +1,31 @@
+import copy
```
FrankLeeeee, line 1: same here.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/linear.py
```diff
@@ -0,0 +1,31 @@
...
+import operator
+from weakref import ProxyType
```
FrankLeeeee, line 3: this is not used as well.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/conv.py
```diff
@@ -0,0 +1,53 @@
...
+import operator
+
+import torch
+
+from ..registry import bias_addition_module
+
+
+@bias_addition_module.register(torch.nn.Conv1d)
+def non_bias_nn_conv1d(tracer, target, args, kwargs):
+    return _non_bias_nn_conv(tracer, target, args, kwargs, torch.nn.functional.conv1d)
```
FrankLeeeee, line 11: I see that all these patched function inherits the same interface. From the design point of view, we should add constraints to make sure that when a new user adds a function, it follows the same interface. One way is that we can create an abstract class to ask the developer to implements the abstract method.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/conv.py
```diff
@@ -0,0 +1,53 @@
...
+def non_bias_nn_conv2d(tracer, target, args, kwargs):
+    return _non_bias_nn_conv(tracer, target, args, kwargs, torch.nn.functional.conv2d)
+
+
+@bias_addition_module.register(torch.nn.Conv3d)
+def non_bias_nn_conv3d(tracer, target, args, kwargs):
+    return _non_bias_nn_conv(tracer, target, args, kwargs, torch.nn.functional.conv3d)
+
+
+def _non_bias_nn_conv(tracer, target, args, kwargs, conv_function):
```
FrankLeeeee, line 24: Add some comments and docstring to explain the logic flow.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/linear.py
```diff
@@ -0,0 +1,31 @@
...
+
+import torch
+
+from ..registry import bias_addition_module
+
+
+@bias_addition_module.register(torch.nn.Linear)
+def non_bias_nn_linear(tracer, target, args, kwargs):
```
FrankLeeeee, line 11: same here.

colossalai/fx/tracer/tracer.py
```diff
@@ -77,18 +77,124 @@ def create_proxy(self, kind, target, args, kwargs, name=None, type_expr=None, pr
...
+
+    def _configure_tracer_type(self, tracer_type: TracerType):
+        if tracer_type == TracerType.DEFAULT:
+            self.proxy_cls = Proxy
+            self.tracer_type = TracerType.DEFAULT
+        elif tracer_type == TracerType.META:
+            self.proxy_cls = ColoProxy
+            self.tracer_type = TracerType.META
+        else:
+            raise ValueError(f"Unrecognised tracer type {tracer_type}")
```
FrankLeeeee, line 191: These methods seem to have no change, but only its position is changed?

colossalai/fx/tracer/tracer.py
```diff
@@ -77,18 +77,124 @@ def create_proxy(self, kind, target, args, kwargs, name=None, type_expr=None, pr
...
+        # to create node on computation graph
+        origin_arguments = (kind, target, args, kwargs, name, type_expr, proxy_factory_fn)
+        # dispatch the arguments generator depending on the kind and target in origin arguments.
+        args_metas, _ = extract_meta(*args, **kwargs)
+        if kind == "call_function":
+            if bias_addition_function.has(target):
+                return bias_addition_function.get(target)
+            elif bias_addition_function.has(target.__name__):
+                # use name for some builtin op like @ (matmul)
+                return bias_addition_function.get(target.__name__)
```
FrankLeeeee, line 100: Why directly return?

colossalai/fx/tracer/tracer.py
```diff
@@ -77,18 +77,124 @@ def create_proxy(self, kind, target, args, kwargs, name=None, type_expr=None, pr
...
+
+        elif kind == "call_method":
+            method = getattr(args_metas[0].__class__, target)
+            if bias_addition_function.has(method):
+                proxy_arguments_dict = bias_addition_function.get(method)
```
FrankLeeeee, line 105: don't quite get it why it directly returns when calling `call_function` but does not return for `call_method`.

tests/test_fx/test_tracer/test_bias_addition_module.py
```diff
@@ -0,0 +1,114 @@
...
+    #     %conv2d : [#users=1] = call_function[target=torch.conv2d](args = (%x, %conv_weight), kwargs = {})
+    #     %view : [#users=1] = call_method[target=view](args = (%conv_bias, [1, -1, 1, 1]), kwargs = {})
+    #     %add : [#users=1] = call_function[target=operator.add](args = (%conv2d, %view), kwargs = {})
+    #     %mul : [#users=1] = call_function[target=operator.mul](args = (%add, 2), kwargs = {})
+    #     return mul
+    graph = tracer.trace(root=model, meta_args={'x': torch.rand(4, 3, 64, 64).to('meta')})
+    # def forward(self, x : torch.Tensor):
+    #     conv_weight = self.conv.weight
+    #     conv_bias = self.conv.bias
+    #     conv2d = torch.conv2d(x, conv_weight);  x = conv_weight = None
```
FrankLeeeee, line 87: I thought this should be `F.conv2d`? However, this looks good to me as well

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/conv.py
YuliangLiu0306: Sure

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/conv.py
YuliangLiu0306: I see, I will create an abstract class.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/linear.py
YuliangLiu0306: ok

colossalai/fx/tracer/tracer.py
YuliangLiu0306: Proxy will be created at patched function internally.

colossalai/fx/tracer/tracer.py
YuliangLiu0306: Sorry, I will fix it, I just focus on the call_module cases last night

colossalai/fx/tracer/tracer.py
YuliangLiu0306: In my view, the position is not changed as well, the call_module function just follow the create_proxy function.

FrankLeeeee: the position change refers to the code position in the python e.g. a function definition is moved from one place to another place, this will show as part of the file commit but in fact no code is changed. Try to avoid such change in the future.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/conv.py
```diff
@@ -0,0 +1,53 @@
...
+from .bias_addition_module import BiasAdditionModule
+
+
+class BiasAdditionConv(BiasAdditionModule):
+
+    def create_non_bias_func_proxy(self):
+        return super().create_non_bias_func_proxy()
+
+    def create_bias_addition_proxy(self, non_bias_func_proxy, bias_proxy):
+        return super().create_bias_addition_proxy(non_bias_func_proxy, bias_proxy)
```
FrankLeeeee, line 14: If these two methods are standard, there is no need to annotate them with `abstract` method.

colossalai/fx/tracer/meta_patch/patched_bias_addition_module/conv.py
```diff
@@ -0,0 +1,53 @@
...
+    def generate(self):
+        non_bias_conv_func_proxy = self.create_non_bias_func_proxy()
+        output_dims = non_bias_conv_func_proxy.meta_data.dim()
+        bias_reshape_proxy = self.create_bias_reshape_proxy(output_dims)
+        bias_addition_proxy = self.create_bias_addition_proxy(non_bias_conv_func_proxy, bias_reshape_proxy)
+        return bias_addition_proxy
+
+
+@bias_addition_module.register(torch.nn.Conv1d)
+def non_bias_nn_conv1d(tracer, target, args, kwargs):
```
FrankLeeeee, line 39: I think I didn't express myself clearly previous. What I mean is that:

```
@bias_addition_module.register(torch.nn.Conv1d)
class BiasAdditionConv(BiasAdditionModule):
       ...
```

and in the proxy side, it will do 
```
handle = bias_addition_module.get(name)(tracer, target, args, kwargs)
return handle.generate()
```

In this way, a developer must follow the interface defined by `BiasAdditionModule`.

colossalai/fx/tracer/meta_patch/__init__.py
```diff
@@ -1,3 +1,5 @@
-from .registry import *
+from .patched_bias_addition_function import *
+from .patched_bias_addition_module import *
```
FrankLeeeee, line 2: Meanwhile, just wonder is it suitable to put it together with the patched function and module. The reason is that `patched_function` and `patched_module` are implemented to help the forward to execute the forward pass. Meanwhile, `patched_bias_addition_function` and `patched_bias_addition_module` are more like doing transform during graph tracing. I feel that there is a logical difference between them.

YuliangLiu0306: I see, I will refactor the file structure.

colossalai/fx/tracer/bias_addition_patch/patched_bias_addition_module/linear.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@bias_addition_module.register(torch.nn.Linear)
+class BiasAdditionLinear(BiasAdditionModule):
+
+    def create_non_bias_func_proxy(self):
+        return super().create_non_bias_func_proxy()
+
+    def create_bias_addition_proxy(self, non_bias_func_proxy, bias_proxy):
+        return super().create_bias_addition_proxy(non_bias_func_proxy, bias_proxy)
```
FrankLeeeee, line 15: can be removed.

YuliangLiu0306: Done

FrankLeeeee (approved):

PR #1784, opened 2022-11-01
[Pipeline]Add flatten when exec_seq is None

Wesley-Jzy:


colossalai/pipeline/pipelinable.py
```diff
@@ -123,23 +142,10 @@ def to_layer_list(self, exec_seq=None):
...
-                            if id(module) == id(child_in_container):
-                                children_name.append(name)
-                                break
-                else:
-                    self._layer_spec_list.append(layer_spec)
-                    for name, module in self._model.named_modules():
-                        if id(module) == id(child):
-                            children_name.append(name)
-                            break
+                self._layer_spec_list = self._layer_spec_list + self._get_flatten_model(child)
```
YuliangLiu0306, line 148: Why we have to trace the leaf submodule? For example, keep the GPTEmbedding is enough for us to do the partition, if users want to use leaf submodules of this Embedding module, such as col_nn.Embedding and dropout, they could use exec_seq to trace the different granularity submodules.

Wesley-Jzy: I added this before I realized that the fwd is the key problem. This PR can be closed now. But recording a iterator in layerspec.children is still weird since even a len() method will cause troubles for children.

YuliangLiu0306: I see.

PR #1789, opened 2022-11-02
[autoparallel] refactor and add rotorc.

super-dainiu:
### What's new?
I refactored the rotor c version. Now our `CheckpointSolverRotor().solve()` has different options, i.e. `force_python=True` and `force_python=False`. 
I will refactor test files later. And after everything is done, we can remove `colossalai.fx.passes.algorithms`.


colossalai/auto_parallel/checkpoint/ckpt_solver_rotor.c
```diff
@@ -0,0 +1,197 @@
+#define PY_SSIZE_T_CLEAN
```
super-dainiu, line 1: Not all c functions are required for checkpoint solver

colossalai/auto_parallel/checkpoint/ckpt_solver_rotor.py
```diff
@@ -180,17 +189,17 @@ def _extract_deps_size():
         return btmp
 
     @staticmethod
-    def _compute_table(chain: Chain, mem_slots: int) -> Tuple:
+    def _compute_table(chain: Chain, mmax: int) -> Tuple:
```
super-dainiu, line 192: mmax sounds more reasonable than mem_slots from my perspective

colossalai/auto_parallel/checkpoint/operation.py
```diff
@@ -163,79 +165,20 @@ class DiscardMemory(MemoryAccess):
...
-        self.name = name
-        self.args = args
-        self.str_args = ','.join(str(v) for v in self.args)
-
-    def __repr__(self):
-        return "{n}({args})".format(n=self.name, args=self.str_args)
-
-
-class Sequence:
+class Sequence(list):
```
super-dainiu, line 168: `Sequence` should be inherited from `list`.

FrankLeeeee (approved):

PR #1792, opened 2022-11-03
[kernelfusion] Added jit fused ops

kurisusnowdeng:
1. Added JIT fused ops for:

- bias_gelu
- bias_dropout_add
- masked_attention

2. Updated fused layernorm kernel for 1D


FrankLeeeee:
The `bias_gelu` and `bias_dropout_add` have already added in https://github.com/hpcaitech/ColossalAI/tree/main/colossalai/kernel/jit

colossalai/nn/layer/parallel_1d/layers.py
```diff
@@ -7,25 +7,44 @@
 
 import torch
 import torch.nn.functional as F
+from apex.normalization import FusedLayerNorm
```
FrankLeeeee, line 10: `apex` is not a necessary dependency for Colossal-AI, importing from apex here will raise exception.

colossalai/nn/fused_ops.py
```diff
@@ -0,0 +1,126 @@
+import math
```
FrankLeeeee, line 1: Many functions here overlap with https://github.com/hpcaitech/ColossalAI/tree/main/colossalai/kernel/jit.

feifeibear (approved):

PR #1798, opened 2022-11-07
[kernel] skip tests of flash_attn and triton when they are not available

feifeibear:


feifeibear:
When triton and flash attention are both available
image

colossalai/kernel/cuda_native/flash_attention.py
```diff
@@ -33,299 +36,396 @@ def triton_check():
...
-    o = tl.load(Out + off_m[:, None] * D_HEAD + off_n[None, :]).to(tl.float32)
-    do = tl.load(DO + off_m[:, None] * D_HEAD + off_n[None, :]).to(tl.float32)
-    denom = tl.load(L + off_m).to(tl.float32)
-    # compute
-    do = do / denom[:, None]
-    delta = tl.sum(o * do, axis=1)
-    # write-back
-    tl.store(NewDO + off_m[:, None] * D_HEAD + off_n[None, :], do)
-    tl.store(Delta + off_m, delta)
+if TRITON_AVALIABLE:
```
oahzxl, line 42: TRITON_AVALIABLE just check the cuda version. So it can be true if triton is not installed. Maybe we should also use global flag HAS_TRITON or check triton install in triton_check

feifeibear: I see, I will update the PR!

feifeibear: updated!

PR #1799, opened 2022-11-07
[fx] skip diffusers unitest if it is not installed

feifeibear:


tests/test_fx/test_tracer/test_hf_model/test_hf_diffuser.py
```diff
@@ -16,6 +21,7 @@
 TIME_STEP = 2
 
 
+@pytest.mark.skipif(not HAS_DIFFUSERS, reason="diffusers has not be installed")
```
super-dainiu, line 24: reason="``diffusers`` has not been installed"

super-dainiu:
Should we remove diffusers from test_requirements.txt also? The CI has difficulties installing this.

FrankLeeeee:
> Should we remove diffusers from test_requirements.txt also? The CI has difficulties installing this.

The standard operation procedure is that, when a developer wants to the do the test, he/she should do `pip install -r test-requirenements.txt` first before executing `pytest` in stead of making this test condititonal.

FrankLeeeee:
> Should we remove diffusers from test_requirements.txt also? The CI has difficulties installing this.

FrankLeeeee:
Sorry accidentally clicked on the `close with comment` button.

feifeibear:
Let's set dependencies such as torchrec, diffuser, and triton as optional in the CAI unit test. In my opinion, our unit tests do not need to be coupled with them.
1. These libraries' interface updates are not under our control. 
2. Unitests need these libraries are bonus optimizatin, and their absence does not actually affect the core functions.

FrankLeeeee:
> Let's set dependencies such as torchrec, diffuser, and triton as optional in the CAI unit test. In my opinion, our unit tests do not need to be coupled with them.
> 
> 1. These libraries' interface updates are not under our control.
> 2. Unitests need these libraries are bonus optimizatin, and their absence does not actually affect the core functions.

OK, this is fine. I can create another decorator to handle this case. Otherwise, every unit test file depending on these libraries will have redundant code 
```
try:
     import xxx
     HAS_XXX = True
else:
     HAS_XXX = False
```

feifeibear:
why use else instead of except?

PR #1808, opened 2022-11-07
[autoparallel] fix bugs caused by negative dim key

YuliangLiu0306:
This PR converts the key of dim_partition_dict with negative value to positive value, which could avoid some bugs due to negative value pointing the same dimension as other key with positive value.

This PR also adds ignore_exception decorator to node handlers.

### DIFF
--- a/colossalai/tensor/utils.py
+++ b/colossalai/tensor/utils.py
@@ -1,7 +1,8 @@
-import torch
+from typing import Dict, Iterator, List, Tuple, Union
 
-from typing import Iterator, Tuple, Union
+import torch
 import torch.nn as nn
+
 from colossalai.tensor.colo_tensor import ColoTensor
 
 
@@ -12,7 +13,7 @@ def all_gather_simulator(target_pair):
 
     We don't allow uncontiguous layout, such as all-gather(S012)->S02 is NOT allowed.
     Therefore, all gather operation just remove the last element in shard list,
-    e.g.: 
+    e.g.:
         all-gather(S01) -> S0
 
     Argument:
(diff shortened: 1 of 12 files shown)


colossalai/tensor/utils.py
```diff
@@ -164,3 +165,17 @@ def convert_parameter(module: torch.nn.Module, param_name: str):
...
+    '''
+    This method is used to convert the negative dim value to positive.
+    '''
+    dims_to_convert = []
+    for dim, mesh_list in dim_partition_dict.items():
+        if dim < 0:
+            dims_to_convert.append(dim)
+    for dim in dims_to_convert:
+        dim_partition_dict.pop(dim)
+        dim_partition_dict[dim_size + dim] = mesh_list
```
feifeibear, line 180: mesh_list is the last value in dict `dim_partition_dict`.
Do you mean that all the mesh_lists are the same?

YuliangLiu0306: 'dim_partition_dict' is a dictionary type datastructure whose key is dimension to be sharded and value is the mesh list decribing the sharding axises. Actually, negative dim value is okay if there is no further action on this datastructure. However, transposes or even squeeze may happen during logical shape recovering to physical shape, which means negative dim value may potentially introduce an invisible bug. 
For example, linear weight dim partition dict is {0: [0], -1:[1]}, if transpose logic for linear weight convert dim 0 into 1 and dim 1 into 0, the dim partition dict finally looks like {1: [0], -1: [1]}.

feifeibear (approved):

PR #1858, opened 2022-11-09
[fx] revisit lazy init

feifeibear:
Reimplement the Lazy Init Context. I can not pass the unittest of `test_complete_workflow`.
I believe that the `self._intercepted_nn_init_func_cache` is not correctly set.


feifeibear:
[](url)

PR #1859, opened 2022-11-09
[fx] test_complete_workflow on cuda

feifeibear:


tests/test_fx/test_complete_workflow.py
```diff
@@ -41,11 +43,12 @@ def run_workflow(world_size):
...
 
     # annotate
-    annotated_gm = transformer_mlp_pass(gm, process_group=ProcessGroup())
+    annotated_gm = transformer_mlp_pass(gm, process_group=ProcessGroup(tp_degree=world_size))
     annotated_gm.recompile()
 
     # materialization and sharding
-    ctx.lazy_init_parameters(annotated_gm)
+    ctx.lazy_init_parameters(annotated_gm, dev)
+    model = model.to(dev)
```
feifeibear, line 51: @FrankLeeeee @YuliangLiu0306 do you think this line is necessary? If I delete it, it will fail.

YuliangLiu0306: This line is not necessary if the lazy init module works correctly. I think we'd better to fix the lazy init context first, and use this(delete to device version) as the test case.

feifeibear: Thanks, yuliang!

PR #1860, opened 2022-11-09
[amp] add torch amp test

xcnick:
This pr add a torch amp unit test, compare torch amp and apex amp.

### DIFF
--- a/tests/test_amp/test_torch_fp16.py
+++ b/tests/test_amp/test_torch_fp16.py
@@ -20,7 +20,7 @@ def run_torch_amp():
     torch.backends.cudnn.deterministic = True
 
     # create layer
-    test_models = ['repeated_computed_layers', 'nested_model', 'resnet18']
+    test_models = ['resnet18', 'simple_net']
     for test_name in test_models:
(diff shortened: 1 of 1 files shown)


feifeibear (approved):
Perfect! Thanks Chi!

feifeibear:
@xcnick can you fix the CI error
https://github.com/hpcaitech/ColossalAI/actions/runs/3429779426/jobs/5715859214

xcnick:
@feifeibear This may be caused by an error in the upstream library apex, [this pr](https://github.com/NVIDIA/apex/pull/1282) is ready to fix this error, but it has not been merged so far.

I manually modified the Apex source code according to the pr in my local environment, and the unit test passed.

feifeibear:
> @feifeibear This may be caused by an error in the upstream library apex, [this pr](https://github.com/NVIDIA/apex/pull/1282) is ready to fix this error, but it has not been merged so far.
> 
> I manually modified the Apex source code according to the pr in my local environment, and the unit test passed.

That won't work for CAI. Since we must depend on a stable apex version. Can you think of some way to bypass this problem?

xcnick:
ok，I'll figure it out.

PR #1868, opened 2022-11-10
[autoparallel] add torch.nn.ReLU metainfo

Cypher30:
# What’s New?

In this PR I implement the `MetaInfo` For `torch.nn.ReLU`, and modify the memory test


colossalai/auto_parallel/meta_profiler/metainfo.py
```diff
@@ -91,11 +92,17 @@ def compute_metainfo(self):
...
+        meta_func = meta_register.get(self._target.__class__)
 
         # construct args for meta_func
         args = [self.compute_sharded_tensor(k, v) for k, v in self._strategy.sharding_specs.items()]
 
+        # construct kwargs
+        if self.target in INPLACE_MODULE:
+            kwargs = {'inplace': self.target.inplace}
+        else:
+            kwargs = {'inplace': False}
```
FrankLeeeee, line 105: What will happen is it is relu and inplace is False?

tests/test_auto_parallel/test_tensor_shard/test_metainfo/test_activation_metainfo.py
```diff
@@ -0,0 +1,61 @@
...
+    mem_test_for_node_strategy(rank=rank,
+                               model=model,
+                               device_mesh=device_mesh,
+                               node_index=node_index,
+                               strategy_number=strategy_number,
+                               input_args=[input],
+                               meta_arg_names=['input'])
+
+
+@run_on_environment_flag(name='AUTO_PARALLEL')
```
FrankLeeeee, line 51: If this test has no issue, you don't need this decorator. This decorator is used when we need to implement a test, but this test cannot be passed at the moment. Thus, we use this decorator to skip the test in CI but run this test locally for verification.

colossalai/auto_parallel/meta_profiler/metainfo.py
Cypher30: The ReLU with `inplace=False` will consume an extra space

tests/test_auto_parallel/test_tensor_shard/test_metainfo/test_activation_metainfo.py
Cypher30: I add this decorator because this test is more like a benchmark than a unit test. I think it’s ok to add this one.

colossalai/auto_parallel/meta_profiler/metainfo.py
FrankLeeeee: Will meta info differentiate these two cases (inplace True and False)?

Cypher30: Sure, `inplace=False` will create a new tensor in forward phase, while `inplace=True` will not

super-dainiu (approved):

PR #1880, opened 2022-11-10
[sc] add examples for auto checkpoint.

super-dainiu:
cc @Cypher30 please help me debug these files.

- [ ] debug
- [ ] add GPT examples
- [ ] polish README

PR #1889, opened 2022-11-10
[SC] add GPT example for auto checkpoint

Cypher30:
# What’s New?

In this PR, I fix the problem mention in the previous PR by @super-dainiu in #1880, add GPT2 test together with batch vs. throughput test, and refactor those tests into `.py` scripts.


examples/tutorial/auto_parallel/bench_utils.py
```diff
@@ -1,10 +1,13 @@
 import time
+from copy import deepcopy
```
FrankLeeeee, line 2: As a tutorial, it should be expected that comments are almost everywhere, at least detailed comments for each section of your code, so that the user can better understand what is going on. However, I see no comments in this file. Please do this for the jupyter notebook as well.

Cypher30: Sure I will add some later

examples/tutorial/auto_parallel/demo_resnet50.py
```diff
@@ -0,0 +1,106 @@
...
+        "--sample_points",
+        type=int,
+        default=15,
+        help=
+        "number of sample points for benchmark from start memory budget to maximum memory budget (free_memory), default 15"
+    )
+    parser.add_argument("--free_memory",
+                        type=int,
+                        default=11000 * 1024**2,
+                        help="maximum memory budget for benchmark, default maximum memory avialable of current device")
```
super-dainiu, line 96: Should be more straightforward, issit? Cuz our audience may not calculate 11000 * 1024**2 by themselves.

super-dainiu: Maybe in `MB` format

examples/tutorial/auto_parallel/bench_utils.py
super-dainiu: How shall we present the failure of training Resnet152 without checkpoint in `.py`?

Cypher30: I think maybe in README?

examples/tutorial/auto_parallel/demo_resnet50.py
Cypher30: I polish the help information here, I think it is clear enough

Cypher30: Oops I forget the command should be typed in bash, so I change it into MB format

binmakeswell (approved):

PR #1895, opened 2022-11-11
[autoparallel] Add alpha-beta estimation 

zhang677:
**What's new?**
Given a list of 1D devices, return $\alpha$ and $\beta$


tests/test_device/test_alpa_dp.py
```diff
@@ -0,0 +1,15 @@
+from colossalai.device import get_submesh_choices
+
+
+def test_alpa_dp():
```
YuliangLiu0306, line 4: I think you should add some assertions in unit test. I know the profiling result may have a large variance, but you could check whether the result in reasonable range, such as latency less than 100ms.

PR #1897, opened 2022-11-11
[fx] test GPT lazy init and recompile

feifeibear:


tests/test_fx/test_complete_workflow.py
```diff
@@ -32,17 +59,45 @@ def forward(self, x):
...
     for param in model.parameters():
         assert param.is_meta
 
     # tracing
     tracer = ColoTracer()
-    graph = tracer.trace(model)
+    graph = tracer.trace(model, meta_args=model_config[1])
     gm = torch.fx.GraphModule(model, graph, model.__class__.__name__)
 
     # annotate
```
feifeibear, line 103: @YuliangLiu0306 I believe this function `transformer_mlp_pass` is not robust enough.

FrankLeeeee: Yes, it is a very simple pass which can only deals with simple MLP model. It was implemented only for the purpose of going through the workflow.

feifeibear: After discussion with yuliang, he said `transformer_mlp_pass ` is a toy, so we will not use it in a real case.

feifeibear:
image

PR #1964, opened 2022-11-16
[zero] test gradient accumulation

1SAA:


tests/test_zero/low_level_zero/test_grad_acc.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # check updated param
+    for z1p, z2p in zip(zero1_model.parameters(), zero2_model.parameters()):
+        assert torch.equal(z1p.data, z2p.data)
+
+
+def exam_zero_1_grad_acc():
+    local_rank = torch.distributed.get_rank()
+    grad_scale = 32
+    seed_all(2008)
```
feifeibear, line 89: Please test the feature with prebuilt common test cases instead of your own simple net.
image

tests/test_zero/low_level_zero/test_zero1_2.py
```diff
@@ -106,31 +107,34 @@ def check_sharded_optim_against_torch_ddp():
...
-    torch_model = DDP(torch_model.cuda())
+    # torch_model = DDP(torch_model.cuda(), bucket_cap_mb=0)
+    torch_model = torch_model.cuda()
+
+    # for (n, p), z1p in zip(torch_model.named_parameters(), zero_model.parameters()):
+    #    half_close(p.data, z1p.data)
 
     # create optimizer
-    zero_optimizer = torch.optim.Adam(zero_model.parameters(), lr=0.001)
+    zero_optimizer = torch.optim.SGD(zero_model.parameters(), lr=1)
```
feifeibear, line 125: what if optimizer is Adam? will it still be OK? I belive ZeRO is exactly proposed for the 2nd order optimizers.

feifeibear: Is Adam ready to use in this PR?

tests/test_zero/low_level_zero/test_grad_acc.py
1SAA: I believe we should keep the simple net test here. there are two reasons.
first, it is easy to debug. second, we can set a smaller tolerance to test correctness.
we may need to add test for prebuilt models, but we should not delete simple net test.

tests/test_zero/low_level_zero/test_zero1_2.py
1SAA: Adam is used in this pr, see file `test_grad_acc.py`. but it is not ready for `[test_zero1_2.py](https://github.com/hpcaitech/ColossalAI/pull/1964/files/d99723ec6dc109b662a1b83691f1ee9603fb450a#diff-3cc07fc076702467742b211fc8b5d8283a7f5d2cf80c6b6fe42b98be3d6cee66)`

colossalai/zero/sharded_optim/low_level_optim.py
```diff
@@ -225,17 +235,21 @@ def _initialize_optimizer_states(self):
             fp32_partition_grad = torch.zeros_like(fp32_partition_param)
             fp32_partition_param.grad = fp32_partition_grad
 
+        # we do not need log information for optimizer, so comment them
         # update the parameter with zero gradients for initialization of optimizer states
-        self._optimizer.step()
+        # self._optimizer.step()
```
FrankLeeeee, line 240: This step is not for logging, but for optimizer state initialization.

colossalai/zero/sharded_optim/low_level_optim.py
```diff
@@ -58,7 +58,10 @@ def __init__(
             mp_parallel_mode=ParallelMode.MODEL,
 
     # cpu offload
-            cpu_offload=False):
+            cpu_offload=False,
+
+    # forced dtype
+            forced_dtype=None):
```
FrankLeeeee, line 64: Why do we need this？

colossalai/zero/sharded_optim/low_level_optim.py
1SAA: it influences the step of Adam.

colossalai/zero/sharded_optim/low_level_optim.py
```diff
@@ -225,17 +235,21 @@ def _initialize_optimizer_states(self):
...
+        # we do not need log information for optimizer, so comment them
         # update the parameter with zero gradients for initialization of optimizer states
-        self._optimizer.step()
+        # self._optimizer.step()
 
         # remove the grad of the paramter to save memory
-        for group_id, fp32_flat_tensor in self._fp32_flat_param_groups_of_current_rank.items():
-            fp32_flat_tensor.grad = None
+        # for group_id, fp32_flat_tensor in self._fp32_flat_param_groups_of_current_rank.items():
+        #     fp32_flat_tensor.grad = None
```
FrankLeeeee, line 244: Why comment out these lines？

tests/test_zero/low_level_zero/test_grad_acc.py
feifeibear: We must add tests for test on prebuilt models. This is our minimum guarantee for software quality. There are also some simple net in the prebuilt models. The test logic should not sensitive to the model structures, otherwise, it should not be written in the test_XXX.py.

colossalai/zero/sharded_optim/low_level_optim.py
FrankLeeeee: In what way？

colossalai/zero/sharded_optim/low_level_optim.py
1SAA: if we need to integrate to lightning, we need a way to force the optimizer in a current  dtype.

feifeibear (approved):

PR #1973, opened 2022-11-17
[autoparallel] Add alpha beta

zhang677:
**What's new?**
Given a list of 1D devices, return $\alpha$ and $\beta$

### DIFF
--- a/tests/test_device/test_alpha_beta.py
+++ b/tests/test_device/test_alpha_beta.py
@@ -3,7 +3,7 @@ import pytest
 from colossalai.device import profile_alpha_beta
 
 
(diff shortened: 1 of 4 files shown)


tests/test_device/test_alpha_beta.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from colossalai.device import profile_alpha_beta
+
+
+@pytest.mask.skip(reason="Skip because assertion fails for CI devices")
```
YuliangLiu0306, line 6: It should be pytest.mark, not mask

zhang677: orz

YuliangLiu0306 (approved):

PR #1984, opened 2022-11-18
[Gemini] open grad checkpoint when model building

feifeibear:


feifeibear:
The orange line indicates mem stats when using grad checkpoint
The blue line indicates  mem stats when using grad checkpoint

image

PR #1998, opened 2022-11-22
[hotfix] make Gemini work for conv DNN

feifeibear:
What this PR is for:
1. fix the bug that Gemini can not work for vgg, alexnet.

What did I do
1. the bug comes from inplace relu OP, so I do some special trick for them....

GOenS6LxAW

### DIFF
--- a/colossalai/nn/_ops/element_wise.py
+++ b/colossalai/nn/_ops/element_wise.py
@@ -24,7 +24,6 @@ def register_elementwise_op(op):
             return op(input_tensor, *args, **kwargs)
         else:
             output = op(input_tensor, *args, **kwargs)
-            assert isinstance(output, Tensor), f"op: {op}"
             # return output
             if isinstance(input_tensor, ColoTensor):
                 if isinstance(output, str):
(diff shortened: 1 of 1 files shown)


colossalai/nn/_ops/element_wise.py
```diff
@@ -15,16 +17,22 @@ def elementwise_op(input_tensor: GeneralTensor, *args, **kwargs):
...
-            if isinstance(output, str):
-                return output
-            if not isinstance(output, torch.Tensor):
-                raise NotImplementedError
-            return ColoTensor.from_torch_tensor(output,
-                                                spec=ColoTensorSpec(input_tensor.get_process_group(),
-                                                                    dist_attr=input_tensor.dist_spec))
+        if 'inplace' in kwargs:
+            # TODO(jiaruifang) inplace will cause bugs
+            print(op, 'input_tensor ', type(input_tensor))
```
1SAA, line 22: remove the print for debug.

ver217 (approved):

colossalai/nn/_ops/element_wise.py
feifeibear: sure! I have updated the PR

PR #2027, opened 2022-11-25
[Gemini] fix the inplace add bug

feifeibear:


colossalai/nn/parallel/data_parallel.py
```diff
@@ -272,7 +272,7 @@ def _setup_grads_ptr(self):
             p.grad = None
 
     def _post_backward(self):
-        assert self.chunk_manager.accessed_mem == 0
+        # assert self.chunk_manager.accessed_mem == 0
```
feifeibear, line 275: @1SAA please help me check this line.

PR #2037, opened 2022-11-29
[autoparallel] adapt solver with self attention

YuliangLiu0306:
### What does this PR do

1. update the node merge rule.
2. update the CostGrpah class and Solver class to adapt the latest interface.
3. update the reshape utils to deal with the corner case.
4. upload the self-attention test file to reproduce the search result.

### DIFF
--- a/colossalai/auto_parallel/tensor_shard/solver/cost_graph.py
+++ b/colossalai/auto_parallel/tensor_shard/solver/cost_graph.py
@@ -72,13 +72,13 @@ class CostGraph:
                 """
                 This method is used to check whether the data has a tensor inside or not.
                 """
-                label = False
+                has_tensor_flag = False
                 if isinstance(data, torch.Tensor):
                     return True
                 elif isinstance(data, (tuple, list)):
                     for d in data:
-                        label = label or _check_tensor_in_node(d)
-                return label
+                        has_tensor_flag = has_tensor_flag or _check_tensor_in_node(d)
+                return has_tensor_flag
 
             for node in strategies_vector.predecessor_nodes:
                 if _check_tensor_in_node(node._meta_data):
@@ -92,7 +92,9 @@ class CostGraph:
 
             if self.simplify and strategies_vector.check_merge():
                 for followed_node in strategies_vector.predecessor_nodes:
-                    # we only merge node pairs which src node has a tensor element inside
+                    # we only merge node pairs which src node has a tensor element inside.
+                    # This is necessay because the node without a tensor element inside will not
+                    # be assigned any strategy.
                     if _check_tensor_in_node(followed_node._meta_data):
                         self.merge_pair.append((followed_node, dst_node))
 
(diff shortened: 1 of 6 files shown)


colossalai/auto_parallel/tensor_shard/solver/cost_graph.py
```diff
@@ -63,14 +63,38 @@ def _build_cost_graph(self):
...
+            # parent_nodes = [node for node in strategies_vector.predecessor_nodes]
+            # children_nodes = [node for node in strategies_vector.successor_nodes]
+            parent_nodes = []
+            children_nodes = []
+
+            def _check_tensor_in_node(data):
+                """
+                This method is used to check whether the data has a tensor inside or not.
+                """
+                label = False
```
FrankLeeeee, line 75: `label` -> `has_tensor_flag` for clarity.

colossalai/auto_parallel/tensor_shard/solver/cost_graph.py
```diff
@@ -63,14 +63,38 @@ def _build_cost_graph(self):
...
+                if _check_tensor_in_node(node._meta_data):
+                    children_nodes.append(node)
+
             setattr(dst_node, 'parents', parent_nodes)
             setattr(dst_node, 'children', children_nodes)
 
             if self.simplify and strategies_vector.check_merge():
                 for followed_node in strategies_vector.predecessor_nodes:
-                    self.merge_pair.append((followed_node, dst_node))
+                    # we only merge node pairs which src node has a tensor element inside
```
FrankLeeeee, line 95: Can you give a specific case to explain why this is necessary in the comment?
```
# we only merge node pairs which src node has a tensor element inside
# this is necessary because ...
# consider this case:
# ...
```

colossalai/auto_parallel/tensor_shard/solver/solver.py
```diff
@@ -160,6 +160,7 @@ def _prepare_data_for_solver(self):
                     compute_cost = compute_cost_item.total
                     memory_cost = memory_cost_item.total
 
+                memory_cost = memory_cost.parameter + memory_cost.activation + memory_cost.buffer
```
FrankLeeeee, line 163: In line 157 and line 161, the variable `memory_cost` has been assigned, why do we overwrite it here?

colossalai/auto_parallel/tensor_shard/solver/cost_graph.py
YuliangLiu0306: Sure

colossalai/auto_parallel/tensor_shard/solver/cost_graph.py
YuliangLiu0306: OK

colossalai/auto_parallel/tensor_shard/solver/solver.py
YuliangLiu0306: I will add some comments to explain it. In short, the memory_cost in 157 and 161 is in type of MemoryCost, and what solver wants is int, so I extract the related values in line 163.

FrankLeeeee: Ok, I see. It is kind of dizzy to read without the comments haha

colossalai/auto_parallel/tensor_shard/solver/cost_graph.py
FrankLeeeee: @YuliangLiu0306 this comment is not handled.

FrankLeeeee (approved):

PR #2043, opened 2022-11-30
[bug] point no leaf module bug

feifeibear:


ver217 (approved):

tests/test_gemini/update/test_optim.py
```diff
@@ -40,8 +41,8 @@ def check_param(model: ZeroDDP, torch_model: torch.nn.Module):
 
 
 # 'gpt2', 'bert',
-TEST_MODELS = ['gpt2', 'bert']
-EXAMPLE_MODELS = ['simple_net']
```
1SAA, line 44: Why remove example models? If the two exam functions test the same thing, you do not need to test twice.

feifeibear: sorry, I revert it back.

tests/test_gemini/update/test_optim.py
```diff
@@ -57,8 +58,12 @@ def exam_model_step(placement_policy, model_name: str):
     torch_model, torch_optim = convert_to_apex_amp(torch_model, torch_optim, amp_config)
     torch_model = DDP(torch_model, device_ids=[dist.get_rank()])
 
-    with ColoInitContext(device=get_current_device()):
+    init_dev = get_current_device()
+    with ColoInitContext(device=init_dev):
         model = model_builder()
+
+    post_process_colo_init_ctx(model, device=init_dev)
```
1SAA, line 65: Could we add this function when exiting the ColoInitContext?

feifeibear: No.... ColoInitContext dose not know the root module.

tests/test_gemini/update/test_optim.py
feifeibear: DONE. add all possible models.

tests/test_gemini/update/test_optim.py
feifeibear: I add more design details in PR description.

PR #2053, opened 2022-11-30
[release] update to 0.1.11rc5

FrankLeeeee:
# What does this PR do?

Release a new version to update wheels for issues mentioned in #2048 .


version.txt
```diff
@@ -1 +1 @@
-0.1.11rc4
+0.1.11rc5
```
feifeibear, line 1: __init__.py好像也要改

FrankLeeeee: There is no need to update `colossalai.__init__.py` any more as mentioned in #2048 .

binmakeswell (approved):

PR #2058, opened 2022-12-01
[autoparallel] add binary elementwise metainfo for auto parallel

Cypher30:
# What's New?

In this PR, I add binary elementwise metainfo for auto parallel. I also fix some annotations.

NOTE: Currently this metainfo is only a rough estimation for binary elementwise operations, in some cases, memory estimation could not match with the real one. Moreover, the memory footprint of different operations could be different, I will dive into those problems later.


YuliangLiu0306 (approved):

colossalai/auto_parallel/meta_profiler/constants.py
```diff
@@ -1,5 +1,16 @@
+import operator
```
YuliangLiu0306, line 1: This file may be integrated with the constants.py in tensor_shard folder.

Cypher30: Sure, that will be better

colossalai/auto_parallel/meta_profiler/constants.py
```diff
@@ -1,5 +1,12 @@
...
+
 import torch
 import torch.nn as nn
 
+from ..tensor_shard.constants import *
```
Cypher30, line 6: @YuliangLiu0306 I will import `constants` in `tensor_shard` here first

PR #2062, opened 2022-12-02
[autoparallel] refactor metainfo forward memory calculation

Cypher30:
# What's New?

In this PR, I refactor the forward memory calculation of former patched operations. After this PR, in SPMD solver, we will use the conservative way to estimate the memory, i.e. always believing that shape consistency will create a new tensor for TP.


YuliangLiu0306 (approved):

Cypher30:
Build and Test failed because the latest hugging face GPT2 model could not be traced

PR #2066, opened 2022-12-02
[Pipeline Middleware] Adapt scheduler for Topo

Wesley-Jzy:


colossalai/pipeline/rpc/_pipeline_base.py
```diff
@@ -251,7 +249,6 @@ def _make_args_kwargs(self, microbatch, merge=False):
     # TODO(jiangziyue) Consider whether this function should be protected by Lock in DAG env.
     # TODO(jiangziyue) Define a Class for DAG.
     def set_input(self, microbatch_id: int, microbatch: Tuple[Any], forward_only: bool):
-        assert self.consumer_stage_ids is not None
```
YuliangLiu0306, line 254: Can you give me an example to show why this assertion is unreasonable?

colossalai/pipeline/rpc/_pipeline_base.py
```diff
@@ -344,16 +332,10 @@ def subscribe_producer(self, microbatch_id: int, forward_only: bool):
                 key = UniqueKey(microbatch_id, Phase.FORWARD)
                 if key in self.work_list:
                     return
-
-            producer_stage_ids = []
-            with self.producer_consumer_init_lock:
```
YuliangLiu0306, line 349: Why we need this lock previously, and why we could remove it safely now?

colossalai/pipeline/rpc/_pipeline_base.py
Wesley-Jzy: It's reasonable in previous design. But I remove the code dependent on self.consumer_stage_ids because it will introduce an extra def-use case when our design consider the situation of multithreads. In fact, it can be extracted from the topo attribute anytime we want to use it. So this dependence is not necessary.

Wesley-Jzy: And also, it's main cause of data race corruption.

colossalai/pipeline/rpc/_pipeline_base.py
Wesley-Jzy: Same as the last comments. In past, we calculate both of them in every worker thread and save it for possible rpc users(even in different rpc threads). Since stage relationship is fixed if and only if topo is given, I remove the def-use depends of  producer_stage_ids and consumer_stage_ids by re-calculating from topo anytime we need it.

YuliangLiu0306 (approved):

PR #2075, opened 2022-12-05
[Gemini] add albert in test models.

feifeibear:


feifeibear:
@superhao1995 The Albert runtime tracer results.

image

feifeibear:
The bert case
image

PR #2089, opened 2022-12-06
[autoparallel]add embedding handler

YuliangLiu0306:
### What does this PR do

1. This PR introduces embedding handler to generator valid strategies for embedding handler.
2. Construct unit test to check the numerical correctness.

### What need we do in future

1. The computation cost is **_estimated as a dense operation_**, instead of a sparse one. Therefore, we need to take care of it when we want to get the correct time cost estimation.
2. The weight of embedding has 2 dimensions, one is num_embeddings, another is embedding_dims. In this PR, **_only embedding_dims get shard,_** because num_embeddings case will require more actions in runtime preparation passes.


feifeibear (approved):

colossalai/auto_parallel/tensor_shard/node_handler/embedding_handler.py
```diff
@@ -0,0 +1,230 @@
...
+            output_sharding_spec = strategy_copy.get_sharding_spec_by_name(output_op_data.name)
+            try:
+                # replace the 0th dimension in the logical sharding with ith dimension in the physical sharding
+                update_partition_dim(sharding_spec=input_sharding_spec,
+                                     dim_mapping={0: i},
+                                     physical_shape=input_op_data.data.shape,
+                                     inplace=True)
+
+                if last_logical_output_dims in output_sharding_spec.dim_partition_dict:
+                    dim_mapping = {0: i, last_logical_output_dims: last_physical_output_dims}
```
YuliangLiu0306, line 65: I combine the last dims and batch dimension together, instead of doing these actions one by one. Consider this example: 
tensor shape: (1, 4, 16) 
Device mesh: (2, 2)
ShardingSpec before converting: S0S1
If we recover the last dimensions first, the new dim_partition_dict will be {0: [0], 2: [1]}, 0 is inherit from the old_partition_dict({0: [0], 1: [1]}) and 2 comes from the dim_mapping {1: 2}. Then, the new ShardingSpec initialization process will crash, because the size of 0 dimension is 1 which is not dividable by 2.

FrankLeeeee (approved):

PR #2098, opened 2022-12-07
[autoparallel] add bias addtion function class

YuliangLiu0306:
### What does this PR do

1. This PR implements BiasAdditionFunc class which is used to construct the restructure computation graph for call_func node with bias addition inside.
2. This PR uses the BiasAdditionFunc class as the base class to implement LinearBasedBiasFunc class which is used to construct the restructure computation graph for call_func node based on F.linear, such as torch.addmm, torch.addbmm.
3. This PR patchs the nodes with torch.addmm target to make sure the numerical correctness.
     For example:
        The formula for torch.addmm is output = alpha * input + beta * (m1 @ m2)
        The origin node is:
            %addmm: call_func[target=torch.addmm](args = (%input_1, m1, m2), kwargs = {beta=1, alpha=1})
        Restructured graph is:
            %transpose : [#users=1] = call_function[target=torch.transpose](args = (%m2, 0, 1), kwargs = {})
            %linear : [#users=1] = call_function[target=torch._C._nn.linear](args = (%m1, %transpose), kwargs = {})
            %mul : [#users=1] = call_function[target=operator.mul](args = (%input_1, 3), kwargs = {})
            %mul_1 : [#users=1] = call_function[target=operator.mul](args = (2, %linear), kwargs = {})
            %add : [#users=1] = call_function[target=operator.add](args = (%mul_1, %mul), kwargs = {})
4. Construct unit test to check the numerical correctness for the torch.addmm node.

### DIFF
--- a/colossalai/fx/tracer/bias_addition_patch/patched_bias_addition_function/__init__.py
+++ b/colossalai/fx/tracer/bias_addition_patch/patched_bias_addition_function/__init__.py
@@ -1,0 +1,2 @@
+from .addmm import *
+from .bias_addition_function import *
--- a/colossalai/fx/tracer/bias_addition_patch/patched_bias_addition_function/__init__.py
+++ b/colossalai/fx/tracer/bias_addition_patch/patched_bias_addition_function/__init__.py
@@ -1,2 +1,2 @@
-from .addmm import *
-from .bias_addition_function import *
+from .addmm import Addmm
+from .bias_addition_function import BiasAdditionFunc, LinearBasedBiasFunc, func_to_func_dict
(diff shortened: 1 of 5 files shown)


colossalai/fx/tracer/bias_addition_patch/patched_bias_addition_function/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .addmm import *
+from .bias_addition_function import *
```
feifeibear, line 2: I suggest avoiding using *.
point out the function names, and use __all__ = ['XXX'] to specific the function you would like to expose to users.

YuliangLiu0306: sure, I will fix it

feifeibear (approved):

PR #2119, opened 2022-12-12
[NFC] update chunk manager API

feifeibear:
Changes some var names.
The biggest change is Chunk Manager.
append_tensor -> register_tensor.

### DIFF
--- a/colossalai/gemini/chunk/chunk.py
+++ b/colossalai/gemini/chunk/chunk.py
@@ -294,7 +294,7 @@ class Chunk:
         self.chunk_temp = None
 
         self.__scatter()
-        # always gathered chunk does not have shard chunk
+        # gathered chunk never have shard attribute
         if self.keep_gathered:
             return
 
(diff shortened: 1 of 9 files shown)


colossalai/gemini/chunk/chunk.py
```diff
@@ -294,7 +294,7 @@ def close_chunk(self):
         self.chunk_temp = None
 
         self.__scatter()
-        # always gathered chunk does not have shard
+        # always gathered chunk does not have shard chunk
```
1SAA, line 297: why adding `chunk`?

feifeibear: 这句话中文是什么？你写一下吧
gathered chunk是一个名词？always + 名词并不符合语法？

1SAA: `consistently gathered chunk does not have shard attribute`

feifeibear: DONE

PR #2134, opened 2022-12-15
[autoparallel] integrate_gpt_related_tests

YuliangLiu0306:
### What does this PR do
1. This PR integrates all gpt related tests into a folder named 'test_gpt'. 
2. All the solver strategy tests are integrated into test_solver_with_gpt_modules.py.
3. All the runtime numerical tests  are integrated into test_runtime_with_gpt_modules.py.
4. All the gpt modules which are reused in both test files are integrated into gpt_modules.py.

### DIFF
--- a/tests/test_auto_parallel/test_tensor_shard/test_gpt/gpt_modules.py
+++ b/tests/test_auto_parallel/test_tensor_shard/test_gpt/gpt_modules.py
@@ -123,7 +123,6 @@ class GPT2Attention(nn.Module):
         present = (key, value)
 
         attn_output, attn_weights = self._attn(query, key, value, attention_mask, head_mask)
-        attn_output = attn_output
         attn_output = self._merge_heads(attn_output, self.num_heads, self.head_dim)
         attn_output = self.c_proj(attn_output)
         # attn_output = self.resid_dropout(attn_output)
@@ -155,9 +154,8 @@ class GPT2Block(nn.Module):
             attention_mask=attention_mask,
             head_mask=head_mask,
         )
-        attn_output = attn_outputs
         # residual connection
-        hidden_states = attn_output + residual
+        hidden_states = attn_outputs + residual
         residual = hidden_states
         hidden_states = self.ln_2(hidden_states)
         feed_forward_hidden_states = self.mlp(hidden_states)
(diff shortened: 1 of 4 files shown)


tests/test_auto_parallel/test_tensor_shard/test_gpt/gpt_modules.py
```diff
@@ -125,15 +123,11 @@ def forward(
         present = (key, value)
 
         attn_output, attn_weights = self._attn(query, key, value, attention_mask, head_mask)
-
+        attn_output = attn_output
```
feifeibear, line 126: why we need this line

YuliangLiu0306: This line could be removed.

YuliangLiu0306: Done.

tests/test_auto_parallel/test_tensor_shard/test_gpt/gpt_modules.py
```diff
@@ -2,32 +2,30 @@
...
+
+    def __init__(self, intermediate_size, config):
+        super().__init__()
+        embed_dim = config.hidden_size
+        self.c_fc = Conv1D(intermediate_size, embed_dim)
+        self.c_proj = Conv1D(embed_dim, intermediate_size)
+        self.act = ACT2FN[config.activation_function]
+        # We temporarily banned the Dropout layer because the rng state need
+        # to process to get the correct result.
+        # self.dropout = nn.Dropout(config.resid_pdrop)
```
feifeibear, line 20: you can use model.eval() during test. eval will ban dropout

YuliangLiu0306: okay

YuliangLiu0306: I tried model.eval(), it not only ban the dropout module, but change the behavior of LayerNorm module as well. I think comment that line manually is a safer way in unit test.

FrankLeeeee (approved):

tests/test_auto_parallel/test_tensor_shard/test_gpt/gpt_modules.py
FrankLeeeee: This raises another question for our consideration even though it is definitely not supposed to be handled in this PR. PyTorch users often do evaluation after training for a few epochs, I don't think we can support this mode change for now, but good to keep it in mind.

PR #2144, opened 2022-12-17
[autoparallel] memory estimation for shape consistency

Cypher30:
# What's New?

In this PR, I implement memory estimation for shape consistency inside the `ShapeConsistencyManager` class. It is a draft PR so we could have some discussion here to finalize this mechanism.


colossalai/tensor/shape_consistency.py
```diff
@@ -403,6 +405,157 @@ def get_all_one_step_transform_spec(self, source_spec: ShardingSpec, orig_cost_d
...
+            input_numel = np.prod(input_shape)
+            output_numel = input_numel * comm_spec.device_mesh.mesh_shape[comm_spec.logical_process_axis]
+            peak_mem = max(peak_mem, alloc_mem + output_numel * 2)
+            alloc_mem += output_numel
+            if discard_input:
+                alloc_mem -= input_numel
+
+        def split_analysis(comm_spec: CommSpec, discard_input: bool, alloc_mem: int, peak_mem: int):
+            """analyze split memory footprint
+            split will allocate memory for the output tensor if we apply shard on the first dimension of
```
YuliangLiu0306, line 448: If we apply shard on first dimension, no extra memory will be allocated?

colossalai/tensor/shape_consistency.py
```diff
@@ -403,6 +405,157 @@ def get_all_one_step_transform_spec(self, source_spec: ShardingSpec, orig_cost_d
...
+            TrainCycleItem: memory cost of such comm_action_sequence
+        """
+
+        def compute_shape(sharding_spec: ShardingSpec):
+            shape = sharding_spec.entire_shape
+            for dim, shard in sharding_spec.dim_partition_dict.items():
+                shape[dim] = shape[dim] // len(shard)
+            return shape
+
+        def gather_analysis(comm_spec: CommSpec, discard_input: bool, alloc_mem: int, peak_mem: int):
```
YuliangLiu0306, line 427: The type of alloc_mem and peak_mem should be float

colossalai/tensor/shape_consistency.py
```diff
@@ -403,6 +405,157 @@ def get_all_one_step_transform_spec(self, source_spec: ShardingSpec, orig_cost_d
...
+            Args:
+                comm_spec (CommSpec): input CommSpec
+                discard_input (bool): whether to discard the input tensor
+                alloc_mem (int): current allocated memory
+                peak_mem (int): current peak memory
+            """
+            input_shape = compute_shape(comm_spec.sharding_spec)
+            input_numel = np.prod(input_shape)
+            output_numel = input_numel * comm_spec.device_mesh.mesh_shape[comm_spec.logical_process_axis]
+            peak_mem = max(peak_mem, alloc_mem + output_numel * 2)
```
YuliangLiu0306, line 441: As we cannot access the tensor dtype here, should we just return output_numel in the function? If we add the alloc_memory and output_numel together, we need do extra job to recover the output_numel outside after we know the tensor dtype.

Cypher30: I think we will not change the tensor dtype in shape consistency, so the unit of `output_numel`, `input_numel`, `alloc_mem`, `peak_mem` are actually the same?

colossalai/tensor/shape_consistency.py
Cypher30: Yes, as we call `torch.narrow` in the `_split` action, it simply creates a reference to that part of the tensor, and as that part of the tensor is contiguous if we apply shard on the first dimension, the `.contiguous()` will not allocate any extra memory here.

colossalai/tensor/shape_consistency.py
Cypher30: okay~

colossalai/tensor/shape_consistency.py
YuliangLiu0306: So the alloc_mem is not counted in byte?

colossalai/tensor/shape_consistency.py
YuliangLiu0306: I mean the description in doc string is not correct. You should check it out.

colossalai/tensor/shape_consistency.py
Cypher30: Sure! It is more like `alloc_numel` and `peak_numel`, the variables names are a little bit confusing, will it be better if I rename them?

colossalai/tensor/shape_consistency.py
Cypher30: lol my bad

colossalai/tensor/shape_consistency.py
YuliangLiu0306: alloc_numel looks better to me.

Cypher30: So I think the `alloc_mem` and `peak_mem` could be `int`?

YuliangLiu0306: Sure, numel should be in type of int.

Cypher30: Resolved!

YuliangLiu0306 (approved):

colossalai/tensor/shape_consistency.py
```diff
@@ -403,6 +405,158 @@ def get_all_one_step_transform_spec(self, source_spec: ShardingSpec, orig_cost_d
...
+            if idx == 0:
+                fwd_action(comm_spec, False, fwd_alloc_numel, fwd_peak_numel)
+            else:
+                fwd_action(comm_spec, True, fwd_alloc_numel, fwd_peak_numel)
+
+        # analyze memory footprint for backward comm actions sequence
+        bwd_alloc_numel = 0
+        bwd_peak_numel = 0
+        for idx, bwd_action, comm_spec in enumerate(zip(reversed(bwd_actions), reversed(comm_action_sequence))):
+            bwd_action(comm_spec, True, bwd_alloc_numel, bwd_peak_numel)
```
Cypher30, line 552: @super-dainiu backward communication will discard grad input once it is used, but forward communication will not.

PR #2162, opened 2022-12-21
[autoparallel] Attach input, buffer and output tensor to MetaInfo class

Cypher30:
# What's New?

To combine our SPMD solver with the activation checkpoint solver, in this PR, I attach the input, buffer, and output tensor in our `MetaInfo` class. It will pave the way for constructing a new `MetainfoProp` pass that prepares data for the activation checkpoint solver.


colossalai/auto_parallel/meta_profiler/metainfo.py
```diff
@@ -99,14 +102,14 @@ def compute_metainfo(self):
             # module
             meta_func = meta_register.get(self._target.__class__)
 
-            # check whether the target in the module list that we don't need to save activation
-            self.save_fwd_in = self._target.__class__ not in NO_SAVE_ACTIVATION
+            # check whether the target in the list that we don't need to save activation
+            save_fwd_in = self._target.__class__ not in NO_SAVE_ACTIVATION
         except:
```
super-dainiu, line 107: This try-except seems dangerous.
You might use if-else instead, and raise exceptions with error messages

colossalai/auto_parallel/passes/runtime_apply_pass.py
```diff
@@ -126,6 +146,9 @@ def _shape_consistency_apply(gm: torch.fx.GraphModule):
                                                                    runtime_apply,
                                                                    args=(node, origin_dict_node, input_dict_node,
                                                                          node_to_index_dict[node], user_node_index))
+                    mem_cost = get_mem_cost(node, origin_dict_node, input_dict_node, node_to_index_dict[node],
```
Cypher30, line 149: Those lines are used to attach memory cost to shape consistency node

colossalai/auto_parallel/passes/runtime_apply_pass.py
```diff
@@ -45,6 +45,26 @@ def runtime_apply_for_iterable_object(node: Node, origin_dict: Dict, input_dict:
     return rst
 
 
+def get_mem_cost(node: Node, origin_dict: Dict, input_dict: Dict, node_index: int, user_node_index: int):
```
Cypher30, line 48: This function is used to get real memory cost from `ShapeConsistencyManager.mem_cost()` as the `ShapeConsistencyManager` itself doesn't have the access to tensor type.

colossalai/auto_parallel/passes/runtime_apply_pass.py
```diff
@@ -45,6 +47,52 @@ def runtime_apply_for_iterable_object(node: Node, origin_dict: Dict, input_dict:
     return rst
 
 
+def construct_meta_info(node: Node, user_node: Node) -> MetaInfo:
```
Cypher30, line 50: This part is to construct `MetaInfo` for the shape consistency node, I add a TODO here as there is no need to recompute this `comm_cost` as it is computed in the node handler.

colossalai/auto_parallel/meta_profiler/metainfo.py
```diff
@@ -33,10 +33,13 @@ def __init__(self, strategy: ShardingStrategy = None, target: Callable = None) -
         self.memory_cost: TrainCycleItem
 
         # list of input tensors
-        self.fwd_in: list[OperationData]
+        self.fwd_in: list[torch.Tensor]
```
super-dainiu, line 36: `from typing import List`

Cypher30: fixed

super-dainiu (approved):

Issue #2170, opened 2022-12-21
[BUG]:  TypeError: colo_addmm() got an unexpected keyword argument 'out' for 2D tensor parallelism


tongping:
### 🐛 Describe the bug

In nn/layer/parallel_2d/_operation.py，there are some code like this: 

class Matmul_AB_2D(torch.autograd.Function):
   def forward(...): 
       torch.addmm(C, A_list[cur], B_list[cur], out=C)

But this function will be intercepted and handled by the following function first. 

def colo_addmm(input_tensor: GeneralTensor,
               mat1: ColoTensor,
               mat2: ColoTensor,
               beta: Number = 1,
               alpha: Number = 1,
               *args) -> ColoTensor:

However, since colo_addmm() does not define and handle "out" parameter, then it should introduce the following error:
"colo_addmm() got an unexpected keyword argument 'out' ". 

Therefore, we could change as the attached file, and then pass 'out' paramter to torch.addmm function instead. 
![patch_addmm_0 1 13](https://user-images.githubusercontent.com/1903257/208985686-b1c257bb-4e53-4f92-b899-6d38d134f3d0.jpg)

PR #2171, opened 2022-12-21
[hotfix] add kwargs for colo_addmm

tongping:
…ossalAI/issues/2170. Basically, it will fix TypeError: colo_addmm() got an unexpected keyword argument 'out' for 2D tensor parallelism. 
The patch is related to the issue "https://github.com/hpcaitech/ColossalAI/issues/2170"

### DIFF
--- a/colossalai/nn/_ops/addmm.py
+++ b/colossalai/nn/_ops/addmm.py
@@ -55,7 +55,7 @@ def colo_addmm(input_tensor: GeneralTensor,
                mat2: ColoTensor,
                beta: Number = 1,
                alpha: Number = 1,
-               *args) -> ColoTensor:
+               **args) -> ColoTensor:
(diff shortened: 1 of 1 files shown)


colossalai/nn/_ops/addmm.py
```diff
@@ -55,7 +55,7 @@ def colo_addmm(input_tensor: GeneralTensor,
                mat2: ColoTensor,
                beta: Number = 1,
                alpha: Number = 1,
-               *args) -> ColoTensor:
+               **args) -> ColoTensor:
```
feifeibear, line 58: name is not appropriate. use **kwargs

tongping: Will do. Thanks for the suggestion.

feifeibear (approved):

PR #2179, opened 2022-12-22
[autoparallel] new metainfoprop based on metainfo class

Cypher30:
# What's New?

In the previous PR, we introduce `MetaInfo` class to process op meta data (memory, flops). In this PR, I introduce a new `MetaInfoProp` based on this mechanism so that we could combine our SPMD solver and activation checkpoint solver.

NOTE: 
1. The content in this PR is experimental, it might contain some bugs that need to be fixed in future PR after we do some tests on this mechanism.
2. For @super-dainiu , you might carefully look at the backward phase and the shape consistency node as you modify the `_construct_chain()` method for the activation checkpoint solver. Currently, all the attached tensors in PR #2162 follow the original `MetaInfoProp` behavior in `colossalai.fx.passes`. But the shape consistency node will discard the backward input after the first comm operation (see #2144 for detail). You also need to carefully modify the temp memory needed in `_construct_chain()`, as I use the `fwd_mem_tmp` and `bwd_mem_tmp` attr in `GraphInfo` class. Contact me if you meet any problems.


colossalai/auto_parallel/passes/meta_info_prop.py
```diff
@@ -0,0 +1,158 @@
...
+            # attach it to graph_info
+            graph_info.fwd_tmp = buffer_tensor
+
+            # set data_ptr for output_tensor
+            for tensor in output_tensor:
+                self._set_data_ptr(tensor)
+            # attach it to graph_info
+            graph_info.fwd_out = output_tensor
+
+        # fetch other memory informations
```
Cypher30, line 153: @super-dainiu I fetch temp memory informations here and store them in `GraphInfo.fwd_mem_tmp` and `GraphInfo.bwd_mem_tmp`

colossalai/auto_parallel/passes/runtime_preparation_pass.py
```diff
@@ -78,6 +78,10 @@ def _solution_annotatation(gm: torch.fx.GraphModule,
         origin_node_sharding_spec_dict[node_index] = strategies_vector[strategy_index].get_sharding_spec_by_name(
             str(node))
 
+        # attach the corresponding metainfo if node has the attribute `metainfo_vector`
+        if hasattr(node, 'metainfo_vector'):
```
Cypher30, line 82: @super-dainiu meta info for the best strategy is attached to the node here.

colossalai/auto_parallel/passes/meta_info_prop.py
```diff
@@ -0,0 +1,158 @@
...
+    def _set_data_ptr(self, x):
+        """
+        Set uuid to tensor
+        """
+        if isinstance(x, torch.Tensor):
+            if not x.data_ptr():
+                data_ptr = uuid.uuid4()
+                x.data_ptr = lambda: data_ptr
+
+    def _is_inpalce(self, node: Node):
```
super-dainiu, line 40: typo

colossalai/auto_parallel/passes/meta_info_prop.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @compatibility(is_backward_compatible=False)
+    def placeholder_handler(self, node: Node) -> None:
+        """
+        Handle the placeholder node.
+        """
+        graph_info = GraphInfo()
+        out = getattr(node, '_meta_data', None)
+        if out is not None:
+            out = [out]
```
super-dainiu, line 66: better use `normalize_tuple()`
what if `out` is already a tuple of tensors?

Cypher30: fixed

colossalai/auto_parallel/passes/meta_info_prop.py
Cypher30: fixed

super-dainiu (approved):

PR #2219, opened 2022-12-28
[example] update gpt benchmark

1SAA:


examples/language/gpt/train_gpt_demo.py
```diff
@@ -111,6 +112,22 @@ def get_tflops(model_numel, batch_size, seq_len, step_time):
     return model_numel * batch_size * seq_len * 8 / 1e12 / (step_time + 1e-12)
 
 
+def get_model_size(model: nn.Module):
+    total_numel = 0
+    for module in model.modules():
+        for p in module.parameters(recurse=False):
```
feifeibear, line 118: 为啥不model.parameters?你这么统计共享参数会被统计多次？

examples/language/gpt/train_gpt_demo.py
```diff
@@ -157,10 +174,10 @@ def gemini_zero_dpp(model: torch.nn.Module, pg: ProcessGroup, placememt_policy:
                           device=get_current_device(),
                           placement_policy=placememt_policy,
                           pin_memory=True,
-                          hidden_dim=4096,
+                          hidden_dim=8192,
                           search_range_mb=64)
         if placememt_policy == 'const':
-            model.gemini_manager._placement_policy.set_const_memory_boundary(10 * 1024)
+            model.gemini_manager._placement_policy.set_const_memory_boundary(2 * 1024)
```
feifeibear, line 180: set_const_memory_bound的参数能不能弄成一个ratio，你这是这个值很难理解，单位是byte还是MB？

examples/language/gpt/run.sh
```diff
@@ -2,12 +2,12 @@
...
-export TPDEGREE=4
-export GPUNUM=8
-export PLACEMENT='cpu'
+export TPDEGREE=1
+export GPUNUM=1
+export PLACEMENT='const'
 export USE_SHARD_INIT=False
 export BATCH_SIZE=32
-# export MODEL_TYPE="gpt2_24b"
+export MODEL_TYPE="gpt2_10b"
```
feifeibear, line 10: 这行代码别动，默认是medium

examples/language/gpt/train_gpt_demo.py
1SAA: The shared parameters will be calculated for multiple times. If you want to get a metric to measure the total calculation of the model, I bet you should count the shared parameters for multiple times. 
PS. if you check the model size calculated by DS zero3, they use the same way.

feifeibear: I see, you are right!

examples/language/gpt/train_gpt_demo.py
1SAA: I didn't figure out a better way yet. could we leave this problem alone in this pr.

feifeibear (approved):

PR #2226, opened 2022-12-29
[example] add benchmark.sh for gpt

feifeibear:


examples/language/gpt/train_gpt_demo.py
```diff
@@ -309,6 +309,8 @@ def main():
         if n >= WARMUP_STEPS:
             tflops_list.append(step_tflops)
 
+        logger.info(f"max memory {torch.cuda.memory_allocated() / 1024**2} MB", ranks=[0])
```
1SAA, line 312: why do we need this line?

feifeibear: Yes, I changed it to max_ memory_allocated

PR #2228, opened 2022-12-29
[DO NOT MERGE] evaluation works for palm train.py

feifeibear:
I convert this to draft, because GeminiDDP can not run inference. The states will be wrong.


FrankLeeeee:
Should this PR be closed?

PR #2232, opened 2022-12-29
[Pipeline Middleware] Reduce comm redundancy by getting accurate output

Wesley-Jzy:
local unit test 

image

### DIFF
--- a/colossalai/pipeline/rpc/_pipeline_base.py
+++ b/colossalai/pipeline/rpc/_pipeline_base.py
@@ -185,18 +185,7 @@ class WorkerBase(ABC):
             self.module_partition: nn.Module = partition_fn(*partition_args).to(device)
             self.partition_condition_lock.notify_all()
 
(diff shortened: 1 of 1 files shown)


colossalai/pipeline/rpc/_pipeline_base.py
```diff
@@ -230,6 +220,26 @@ def get_output_by_key(self, key: UniqueKey, ref_use=False) -> Any:
 
         return output
 
+    def sync_global_worker_rrefs(self, pp_rank_to_worker_rref: Dict[int, PyRRef]) -> None:
```
YuliangLiu0306, line 223: Why this function moves to here?

Wesley-Jzy: Because I want to put all private functions which start with '_' together, seperated from public functions.

YuliangLiu0306: I see

YuliangLiu0306 (approved):

PR #2261, opened 2023-01-03
[autoparallel] align the data_ptr with the old version of auto activation checkpoint pipeline

Cypher30:
# What's New?

In this PR, I modify the `MetaInfoProp` and some operations' meta registration, so that the `data_ptr` behavior could align with the previous auto activation checkpoint pipeline.

### DIFF
--- a/colossalai/auto_parallel/passes/meta_info_prop.py
+++ b/colossalai/auto_parallel/passes/meta_info_prop.py
@@ -8,9 +8,9 @@ from torch.fx import GraphModule
 from torch.fx.node import Node
 
 from colossalai.auto_parallel.meta_profiler import MetaInfo
+from colossalai.auto_parallel.passes.constants import OUTPUT_SAVED_MOD, OUTPUT_SAVED_OPS
 from colossalai.fx._compatibility import compatibility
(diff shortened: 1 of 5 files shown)


colossalai/auto_parallel/passes/meta_info_prop.py
```diff
@@ -102,56 +102,51 @@ def node_handler(self, node: Node) -> None:
...
-        input_tensor: List[torch.Tensor] = meta_info.fwd_in
-        buffer_tensor: List[torch.Tensor] = meta_info.fwd_buffer
-        output_tensor: List[torch.Tensor] = meta_info.fwd_out
+        input_tensors: List[torch.Tensor] = meta_info.fwd_in
+        buffer_tensors: List[torch.Tensor] = meta_info.fwd_buffer
+        output_tensors: List[torch.Tensor] = meta_info.fwd_out
 
-        if len(input_tensor) > 0:
+        if self._is_inplace(node):
+            # inplace operation will not create new tensor, and it only has one parent node
```
Cypher30, line 110: Operations like `flatten` will change the shape of the tensor, so in the in-place operation I modify the mechanism to set the `data_ptr`.

PR #2283, opened 2023-01-03
[amp] add gradient clipping for unit tests

1SAA:


colossalai/amp/naive_amp/naive_amp.py
```diff
@@ -40,7 +43,7 @@ def step(self):
         return self.optim.step()
 
     def clip_grad_norm(self, model: nn.Module, max_norm: float):
-        pass
+        raise RuntimeError("Misuse error: you should specify the clip_grad_norm in the amp config")
```
FrankLeeeee, line 46: Perhaps it is clearer to say "NaiveAMP optimizer has executed gradient clipping during optimizer.step() if you have supplied `clip_grad_norm` in the amp_config, executing the method clip_grad_norm is not allowed."

1SAA: @FrankLeeeee done.

FrankLeeeee (approved):

PR #2314, opened 2023-01-04
[builder] reconfig op_builder for pypi install

feifeibear:
## What's new
remove duplicated op_builder. Make it work both for setup and runtime building.


op_builder/builder.py
```diff
@@ -102,3 +104,9 @@ def builder(self, name) -> 'CUDAExtension':
                 'cxx': self.cxx_flags(),
                 'nvcc': self.nvcc_flags()
             })
+
+
+if __name__ == '__main__':
+    print(Path(__file__))
+    print(os.path.islink(Path(__file__)))
+    print(Path(__file__).parent.parent.absolute())
```
FrankLeeeee, line 112: Remove these lines?

feifeibear: DONE

FrankLeeeee:
I have checkout out your branch and tested locally. I found that symlink does not work as the `op_builder` folder does not exist after `pip install`.

Steps to reproduce:

1. Check out your branch and install without CUDA extensions
```bash
# do not run this in editable mode
NO_CUDA_EXT=1 pip install -v .
```

2. then get the colossalai file directory
```bash
python -c "import colossalai;print(colossalai.__file__)"
```

3. Check the files in the colossalai directory
Screenshot 2023-01-04 at 15 00 07

FrankLeeeee:
Please ignore my previous comment, it works now.

FrankLeeeee (approved):

PR #2348, opened 2023-01-05
Update requirement_colossalai.txt

haofanwang:
It is required to install 0.2.0 or install from the source. 

`from colossalai.nn.parallel.utils import convert_to_torch_module` do not work in _0.1.11rc5+torch1.12cu11.3_.


binmakeswell:
Thanks.

binmakeswell (approved):

PR #2356, opened 2023-01-06
[gemini] add get static torch model

1SAA:
* added `get_static_torch_model` function. users can get a static torch model used for saving checkpoints or numeric checks.
* updated the unit test of `get_static_torch_model`
* polished the dreambooth training script


Fazziekey (approved):

Fazziekey:
test for example is none, 
image

feifeibear (approved):

Issue #2361, opened 2023-01-06
[BUG]: Parameters missing in the state_dict output of ZeroDPP module


eric8607242:
### 🐛 Describe the bug

Hello,
I currently fine-tune the Huggingface GPT2 with ColossalAI. I follow the example with GeminiDPP and ZeroOptimizer. However, I found that there are some keys missing issue when I load the checkpoint stored by the ColossalAI to the GPT2 model.

The following codebase can reproduce the issue:
```
import colossalai
from colossalai.tensor import ProcessGroup, ShardSpec, ColoParameter
from colossalai.nn.parallel import GeminiDDP
from colossalai.utils import save_checkpoint
from colossalai.utils.model.colo_init_context import ColoInitContext

from transformers import GPT2LMHeadModel
import torch

if __name__ == "__main__":
    device = "cuda:0"
    tp_degree = 1
    shardinit = True
    placement_policy = "cpu"
    path_to_checkpoint = "./test.pth"

    colossalai.launch_from_torch(config={})

    default_pg = ProcessGroup(tp_degree=tp_degree)
    default_dist_spec = ShardSpec([-1], [tp_degree]) if shardinit else None
    with ColoInitContext(
        device,
        dtype=torch.half,
        default_dist_spec=default_dist_spec,
        default_pg=default_pg
    ):
        model = GPT2LMHeadModel.from_pretrained("gpt2")

    model = GeminiDDP(
        model, device=device,
        placement_policy=placement_policy,
        pin_memory=True, hidden_dim=768,
        search_range_mb=64
    )

    save_checkpoint(path_to_checkpoint, epoch=0, model=model)

    original_model = GPT2LMHeadModel.from_pretrained("gpt2")
    checkpoint = torch.load(path_to_checkpoint)
    original_model.load_state_dict(checkpoint["model"])
```

And the error message is shown as the follow:
```
RuntimeError: Error(s) in loading state_dict for GPT2LMHeadModel:
        Missing key(s) in state_dict: "lm_head.weight". 
```

### Environment

Os:
Ubuntu 22.04

GPU:
NVIDIA GeForce RTX 3090

Package list:
pytorch                      2.0.0_cuda11.6_cudnn8.3.2_0
cuda-toolkit              11.6.1  
colossalai                  0.2.0+torch2.0cu11.7

1SAA:
Hi @eric8607242 

you can solve this problem by just adding a argument in your `load_state_dict` function. write like this 
```
[...]

PR #2363, opened 2023-01-06
Fix state_dict key missing issue of the ZeroDDP

eric8607242:
#2361 
Hi, 
I try to dig into the root cause of this issue. 
After digging, I found the reason to raise the checkpoint key missing issue for GPT2 is that in the `ZeroDDP` the state_dict of the model is dumped by the method `self.named_parameters()` which returns the output of the Pytorch `named_parameters()`.

However, in the [original implementation of `named_parameters()` in Pytorch](https://github.com/pytorch/pytorch/blob/66745831d72583de49779d9117baf584e1f176bb/torch/nn/modules/module.py#L2061), the `named_parameters()` only return the weights with the same memory address once. 
This means that if there are two layers that share the same weight as the embedding layer and the final classifier layer in GPT2, the method `named_parameters()` would only return one of them.
But [the original implementation of `state_dict` in Pytorch](https://github.com/pytorch/pytorch/blob/66745831d72583de49779d9117baf584e1f176bb/torch/nn/modules/module.py#L1725) returns the whole state of the module instead.

Therefore, to address this issue, I create a dict named `name_to_fp32_params` to store the mapping between the whole module state and the fp32 params.
```python
# Set keep_vars=True to prevent the parameter to be detached
for name, p in module.state_dict(keep_vars=True).items():
    if p in params_to_fp32_params[name] = params_to_fp32_params[p]
        self.name_to_fp32_params[name] = params_to_fp32_params[p]
```
And in the method `_save_to_state_dict()`, instead, utilize the function `named_parameters()`, I use the `self.module.state_dict()` to get the whole model state.
```python
for name, p in self.module.state_dict().items():
    if p is not None and name in self.name_to_fp32_params:
        fp32_p = self.name_to_fp32_params[name]
    
        assert fp32_p in param_to_save_data, "Parameter '{}' is neglected in the chunk list".format(name)      
        record_parameter = param_to_save_data[fp32_p]
    else:
        record_parameter = p
    destination[prefix + name] = record_parameter
```
With the modification shown above, the `state_dict` can output the correct key value pair for the GPT2 without any impact on other models.


feifeibear:
Thanks @eric8607242 , very impressive.
Could you please solve the conflict? @1SAA please help review this PR.

eric8607242:
@feifeibear 
Hi,
Thanks for the quick response! 
The conflict fix done!

1SAA:
Hi @eric8607242 

Thanks for your great PR. 😊
But we just added a powerful function called `get_static_torch_model` which can generate a torch model from a GeminiDDP module. You can find the function [here](https://github.com/hpcaitech/ColossalAI/blob/48d33b1b1753f19361e7e54a68a7ac5999dc02e4/colossalai/nn/parallel/utils.py#L63). Could you try this function like 
```state_dict = get_static_torch_model(model).state_dict()```? 
Does this function meet your requirement? I think we could use this function to implement our new strict state_dict() funtion.

**Comment by eric8607242** [comment]
Hi @1SAA ,
Thanks for the great work and awesome function.

I think this is exactly what I want!

But, in my opinion, I think there should add a param type check before `colo_to_torch` assignment because the return param of the `named_parameters()` might contain the param with `None`, which may raise the `can not find parameter` error.
```python
for suffix_param_name, param in colo_module.named_parameters(recurse=False):
    if param is None:
        continue
```

Besides, the new function is coupled with the `ZeroDDP.state_dict`. 
Implementing the strict `state_dict` with the new function is a bit of a chicken-and-egg problem. 
And I concern this may confuse other users as I did.

1SAA:
Hi @eric8607242 

I think we can rename the current `state_dict` function as `_non_strict_state_dict` and write a new `state_dict` function based on `get_static_torch_model`. In my opinion, this way leads to minimum editions under the current circumstances.

eric8607242:
Hi @1SAA ,
I think this is a great idea.
Thanks for your suggestion!

I will try to follow this rule to modify the method as soon as possible.

eric8607242:
Hi @1SAA,
I rewrite the `state_dict` model with a new bool variable `strict` to specify which version of`state_dict` would be returned.

Look forward to your feedback.
Thanks!

1SAA:
Hi @eric8607242 

LGTM.

eric8607242:
Hi @1SAA ,
Thanks for review!

I modify the argument `GeminiDDP` module to `ZeroDDP` module in `get_static_torch_model` function to be compatible with the lower version (v0.1.9 and v0.1.10).

1SAA (approved):

PR #2364, opened 2023-01-06
[autochunk] add autochunk feature

oahzxl:
## AutoChunk
High memory consumption is a major bottleneck for models such as Alphafold, VIT and Diffusion. Sometimes it is even diffcult to infere a singe long sequence. To address this issue, AutoChunk is proposed, which automatically partitions the tensor along dimensions that are independent of the computation in the inference phase to reduce peak memory usage.

A simple example of Chunk is as follows, where we chunk the second dim of x, and reduce peak memory of Attention by 87.5%:
```
x = torch.randn([1, 1024, 1024, 256])
result = torch.empty_like(x)
for i in range(0, x.shape[1], 128):
    result[:, i:i + 128, :, :] = Attention(x[:, i:i+ 128, :, :])
```

In AutoChunk, we can adaptively search efficient chunks. The following seudo code shows the workflow of AutoChunk.
image

## Performance
autochunk_mem
autochunk_infer
We evaluate AutoChunk with Alphafold. Openfold w/o chunk is a baseline of Alphafold. Openfold w/ chunk is a expert-designed chunked Alphafold. AutoChunk is our auto generated chunked Alphafold.

For memory, we reduce it by 92.6% compared with Openfold w/o chunk and 34.4% compared with expert-designed chunked Openfold.
For inference speed, we speedup by 15% compared with chunked Openfold, and only lose 4% speed compared with Openfold w/o chunk.


colossalai/autochunk/select_chunk.py
```diff
@@ -0,0 +1,224 @@
+from .estiamte_memory import EstimateMemory
```
super-dainiu, line 1: estimate

oahzxl: Yes, thanks

oahzxl:
image

feifeibear (approved):
I did not revier. Merge this PR will not harm the other features. So let's merge it first.

PR #2374, opened 2023-01-06
[setup] support pre-build and jit-build of cuda kernels

### LINKED ISSUE #2337: [FEATURE]: Build kernel only when executed
### Describe the feature

In the current Colossal-AI implementation, we build Colossal-AI in two ways:
1. built when doing `CUDA_EXT=1 pip install colossalai`
2. build the CUDA kernel when importing Colossal-AI

However, building all kernels at least takes 5 min, which is too long for many users. Moreover, not all kernels are in fact required by a user. Therefore, only build the kernel needed for the current program would be the best to balance user experience and the completeness of the library.

FrankLeeeee:
# Overview

It will in the best interest of the user if they can install colossalai smoothly. As discussed in #2337 , we should support two way kernel build:
1. build kernel during installation
2. build kernel only when needed during runtim

# Issue Number

Fixed #2337 

# What does this PR do?

This PR has done several things:
1. removed rubbish files in `colossalai._C` which were pushed one month ago for no reason
2. enabled prebuild and runtime-build for all existing kernels
3. refactored the files which use these kernels
4. add a `README.md` for the op builder


colossalai/amp/naive_amp/_fp16_optimizer.py
```diff
@@ -30,6 +42,8 @@ def _multi_tensor_copy_this_to_that(this, that, overflow_buf=None):
     if overflow_buf:
         overflow_buf.fill_(0)
         # Scaling with factor `1.0` is equivalent to copy.
+        global multi_tensor_scale
```
feifeibear, line 45: where is multi_tensor_scale? do we need this line?

op_builder/layernorm.py
```diff
@@ -0,0 +1,34 @@
+import os
+
+from .builder import Builder
+from .utils import append_nvcc_threads, get_cuda_cc_flag
+
+
+class FusedOptimBuilder(Builder):
```
feifeibear, line 7: Why its name is FusedOptim?
Does it build layer_norm?

colossalai/amp/naive_amp/_fp16_optimizer.py
FrankLeeeee: Fixed the typo.

op_builder/layernorm.py
FrankLeeeee: Fixed the typo.

feifeibear (approved):

PR #2385, opened 2023-01-07
[doc] updated kernel-related optimisers' docstring

FrankLeeeee:
This PR updated the docstring for the kernel-related optimizers as the kernel can be built during either installation or runtime now.

### DIFF
--- a/colossalai/nn/optimizer/hybrid_adam.py
+++ b/colossalai/nn/optimizer/hybrid_adam.py
@@ -19,7 +19,7 @@ class HybridAdam(NVMeOptimizer):
       * Parameters on GPU and gradients on GPU is allowed.
       * Parameters on GPU and gradients on CPU is **not** allowed.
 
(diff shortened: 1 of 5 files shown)


colossalai/nn/optimizer/hybrid_adam.py
```diff
@@ -19,7 +19,7 @@ class HybridAdam(NVMeOptimizer):
       * Parameters on GPU and gradients on GPU is allowed.
       * Parameters on GPU and gradients on CPU is **not** allowed.
 
-    Requires ColossalAI to be installed via ``pip install .``
+    `HybriadAdam` requires CUDA extensions which cen be built during installation or runtime.
```
feifeibear, line 22: cen -> can

FrankLeeeee: done.

PR #2393, opened 2023-01-09
[autoparallel] integrate device mesh initialization into autoparallelize

YuliangLiu0306:
### What does this PR do
We implement AlphaBetaProfiler to help us detect cluster communication metrics and the best logical device mesh. In this PR, I integrate this feature into autoparallelize api.


FrankLeeeee:
Please remove the `.pt` files from your commit.

YuliangLiu0306:
> Please remove the `.pt` files from your commit.

The .pt file is saved solution which can be used to bypass the spmd solver. The reason to keep it is our solver is not efficient enough for large model, such as gpt2 with more than 12 layers.

FrankLeeeee:
> The .pt file is saved solution which can be used to bypass the spmd solver. The reason to keep it is our solver is not efficient enough for large model, such as gpt2 with more than 12 layers.

We should not store weights in the repository because they would make the repo size large. There are two ways to handle this:
1. store the weights somewhere on the web and download it in the python script
2. cache the weights somewhere on our CI server and expose it to the CI container.

YuliangLiu0306:
> > The .pt file is saved solution which can be used to bypass the spmd solver. The reason to keep it is our solver is not efficient enough for large model, such as gpt2 with more than 12 layers.
> 
> We should not store weights in the repository because they would make the repo size large. There are two ways to handle this:
> 
> 1. store the weights somewhere on the web and download it in the python script
> 2. cache the weights somewhere on our CI server and expose it to the CI container.

Okay, but those .pt files just contain a list of integers in length of graph size, and the largest one only cost 1kb.

FrankLeeeee:
Okay then.

FrankLeeeee (approved):

Issue #2407, opened 2023-01-09
[BUG]: backend for supporting distributed training


haofanwang:
### 🐛 Describe the bug

**Describe the bug**
I'm training with the [dreambooth example](https://github.com/hpcaitech/ColossalAI/tree/main/examples/images/dreambooth) with RDMA. But current backend (gloo) doesn't support it. It leads to following error.

```
RuntimeError: [../third_party/gloo/gloo/transport/tcp/pair.cc:799] connect 
[127.0.1.1]:29591: Connection refused
```

I check the corresponding function, and I guess [this line](https://github.com/hpcaitech/ColossalAI/blob/d3f5ce9efb35bf9e292aa041a3e98b737cbb68ee/colossalai/context/parallel_context.py#L378) is a typo?

old:
`cpu_group = dist.new_group(ranks, backend='gloo') if dist.get_backend() != 'gloo' else None
`
new:
`cpu_group = dist.new_group(ranks, backend='gloo') if dist.get_backend() == 'gloo' else None
`

haofanwang:
I find that the typo above existed in many other files, will add another PR soon.

FrankLeeeee:
Great, thanks!

haofanwang:
@FrankLeeeee Could you re-check this [PR](https://github.com/hpcaitech/ColossalAI/pull/2415)?

FrankLeeeee:
Okay sure.

kurisusnowdeng:
The problem is mostly caused by the firewall since GLOO uses random ports to initialize groups

binmakeswell:
We have updated a lot. This issue was closed due to inactivity. Thanks.

PR #2415, opened 2023-01-10
[hotfix] Fix typos for supporting distributed training

haofanwang:
This PR fixes the distributed training problem mentioned in https://github.com/hpcaitech/ColossalAI/issues/2407.


feifeibear (commented):
@haofanwang thanks for your warm-hearting contribution. However, I believe the original version should not be modified. The original intention is that if the backend is gloo, we don't need to use group_cpu but just use group directly.

haofanwang:
I see, but is it normal? As gloo backend is usually for distributed CPU training. @feifeibear 

The default backend is [nccl](https://github.com/hpcaitech/ColossalAI/blob/93f62dd15288379355d5f4f41c3959e681f5ac41/colossalai/initialize.py#L65), which makes claim "dist.get_backend() != 'gloo'" True, and group_cpu is always used in such a case. To be more specific, [gpc.init_global_dist](https://github.com/hpcaitech/ColossalAI/blob/93f62dd15288379355d5f4f41c3959e681f5ac41/colossalai/initialize.py#L100) and [gpc.init_parallel_groups()](https://github.com/hpcaitech/ColossalAI/blob/93f62dd15288379355d5f4f41c3959e681f5ac41/colossalai/initialize.py#L103) lead to [connection issue](https://github.com/hpcaitech/ColossalAI/issues/2407). However, with above changes, distributed GPU training works fine.

FrankLeeeee:
@feifeibear do we still need this CPU process group? Many users encounter environment issues when init this group.

kurisusnowdeng:
Hi @haofanwang . In our design, `get_group()` returns the default group (which can be nccl, gloo, etc). Meanwhile, as the default group usually is used by gpu communication, we need `get_cpu_group()` to return a gloo group for cpu communication. In your case, just specify the default backend and group as gloo, and then `get_cpu_group()` will redirect you to the default group. However, you are duplicating the gloo group. The change is sadly not how it is supposed to work.

feifeibear:
@kurisusnowdeng @haofanwang Can we close this PR? It looks like the problem has been solved.

Issue #2380, opened 2023-01-06
[BUG]: gpc.set_seed() doesn't work for multi-node training 


haofanwang:
### 🐛 Describe the bug

**Describe the bug**

I'm working on the [dreambooth example](train_dreambooth_colossalai.py).

gpc.set_seed(args.seed)  will cause following error.

`AssertionError: The seed for ParallelMode.DATA has been added
`

if I directly comment out it, everything works fine.

**To Reproduce**

I'm using below command to support further multi-machine training. But while I'm testing, I only run on a single 8*V100 machine.

```
python -m torch.distributed.run --nproc_per_node=$GPU_NUM --nnodes=$WORLD_SIZE \
  --node_rank=$RANK --master_addr=$MASTER_ADDR --master_port=$MASTER_PORT ./train_dreambooth_colossalai.py
```

**Expected behavior**
The seed should be rightly set.

haofanwang:
solved in the recent PR.

PR #2445, opened 2023-01-11
Fix bug in dreambooth example

haofanwang:
There exists a conflict between [gpc.set_seed(seed)](https://github.com/hpcaitech/ColossalAI/blob/169954f87e9d79af0990cd286d89a5308cb21532/colossalai/initialize.py#L113) and [gpc.set_seed(args.seed)](https://github.com/hpcaitech/ColossalAI/blob/main/examples/images/dreambooth/train_dreambooth_colossalai.py#L361) that leads to issue https://github.com/hpcaitech/ColossalAI/issues/2380.

This PR solves this problem.


FrankLeeeee:
Great, thanks for your contribution.

FrankLeeeee (approved):

PR #2464, opened 2023-01-12
[zero] low level optim supports ProcessGroup

feifeibear:
## What's new

This PR removes the dependency of LowLevelZeroOptimizer on gpc. gpc is a global variable. If use it, we can not use LowLevelZeroOptimizer together with ColoTensor TP.

This PR is the first step towards a more flexible zero system. In the future, we can use ColoTensor TP + ZeRO1/ZeRO2/GeminiDPP in a very flexible way.


1SAA:
You should add parallel group checks on the model.

1SAA:
You still should add unit tests for the compatibility between TP and ZeRO.

1SAA (approved):

PR #2468, opened 2023-01-12
[autoparallel] support origin activation ckpt on autoprallel system

YuliangLiu0306:
### What does this PR do

Previously, the communication nodes inserted by auto-parallel system will break the origin checkpoint structure, and the graphmodule generated by fx codegen will have a lot of unexpected checkpoint function.

In this PR, the activation annotation will be considered during the node insertion, and use the ColoGraphModule codegen feature to generate a new distributed model with activation checkpoint.

### DIFF
--- a/tests/test_auto_parallel/test_tensor_shard/test_gpt/test_checkpoint_trace.py
+++ b/tests/test_auto_parallel/test_tensor_shard/test_gpt/test_checkpoint_trace.py
@@ -1,0 +1,70 @@
+from functools import partial
+from typing import Optional, Tuple, Union
+
+import pytest
+import torch
+import torch.multiprocessing as mp
+import torch.nn as nn
+from torch.utils.checkpoint import checkpoint
+from transformers.pytorch_utils import Conv1D
+
+from colossalai.auto_parallel.tensor_shard.initialize import initialize_model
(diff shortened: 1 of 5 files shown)


tests/test_auto_parallel/test_tensor_shard/test_gpt/test_checkpoint_trace.py
```diff
@@ -0,0 +1,70 @@
...
+from colossalai.fx.graph_module import ColoGraphModule
+from colossalai.fx.tracer import ColoTracer
+from colossalai.initialize import launch
+from colossalai.logging import disable_existing_loggers
+from colossalai.tensor.shape_consistency import ShapeConsistencyManager
+from colossalai.testing import rerun_if_address_is_in_use
+from colossalai.testing.pytest_wrapper import run_on_environment_flag
+from colossalai.utils import free_port
+
+HIDDEN_SIZE = 768
```
feifeibear, line 22: can we use a smaller number to increase unitest speed?

YuliangLiu0306: Sure

PR #2485, opened 2023-01-16
[autochunk] support evoformer tracer

oahzxl:
support full evoformer tracer, which is a main module of alphafold. previously we just support a simplifed version of it.
1. support some evoformer's op in fx
2. support evoformer test
3. add a new dir for full version code
ps: now we have 3 code dirs in test_autochunk, which refers to simplified openfold, simplified fastfold, and full fastfold(added in this pr). plan to import them from github repo and remove those dirs in future.


tests/test_autochunk/test_autochunk_openfold_codegen.py
```diff
@@ -0,0 +1,160 @@
...
+import colossalai
+from colossalai.core import global_context as gpc
+from colossalai.fx._compatibility import is_compatible_with_meta
+from colossalai.fx.codegen.activation_checkpoint_codegen import CODEGEN_AVAILABLE
+from colossalai.fx.graph_module import ColoGraphModule
+from colossalai.fx.passes.meta_info_prop import MetaInfoProp
+from colossalai.fx.tracer.experimental import ColoTracer, symbolic_trace
+from colossalai.utils import free_port
+from tests.test_autochunk.evoformer.evoformer import evoformer_base
+from tests.test_autochunk.origin_openfold.evoformer import EvoformerBlock
```
feifeibear, line 17: can we import EvoformerBlock from a 3rd party software? I believe including an original openfold implementation in the test directory is not a good idea.

feifeibear: OK, next PR is fine.

feifeibear (approved):

tests/test_autochunk/test_autochunk_openfold_codegen.py
oahzxl: Already done now.

PR #2504, opened 2023-01-19
[autochunk] support extramsa

oahzxl:
Support extramsa, which is an another major module of Alphafold apart from evoformer.
1. support some new op in extramsa, including cat, sum and clone.
2. support using getitem as a slice function, and update some logic for it.
3. update some annotations.
4. update a test for extramsa.


oahzxl:
image

feifeibear (approved):

PR #2538, opened 2023-02-01
[autochunk] support multi outputs chunk search

oahzxl:
Support multi outputs chunk search. Previously we only support single output chunk search. It is more flexible and improve performance by a large margin. For transformer, we reduce memory by 40% than previous search strategy.
1. rewrite search strategy to support multi outputs chunk search
2. fix many, many bugs
3. update tests


feifeibear:
How is the speed?

feifeibear (approved):

oahzxl:
> How is the speed?

The speed loses about 3-4%. I will update a benchmark in next pr.

PR #3006, opened 2023-03-05
[docker] Add opencontainers image-spec to `Dockerfile`

### LINKED ISSUE #3005: [PROPOSAL]: add image specifications to `Dockerfile`
### Proposal

We can make a few changes to improve the overall quality of the docker image 🐳, such as adding annotations to the `Dockerfile` which later acts as metadata once the image is built. For reference annotations can be found [here](https://github.com/opencontainers/image-spec/blob/main/annotations.md)

### Self-service

- [X] I'd be willing to do some initial work on this proposal myself.

SauravMaheshkar:

## 🚨 Issue number

Fixes #3005

## 📝 What does this PR do?

This PR makes few changes to improve the overall quality of the docker image 🐳 . For reference more annotations can be found [here](https://github.com/opencontainers/image-spec/blob/main/annotations.md)

- [x] 🌝 Yes hell yeah, I do.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,10 +1,10 @@
 FROM hpcaitech/cuda-conda:11.3
 
 # metainformation
-LABEL org.opencontainers.image.version = "0.2.5"
+LABEL org.opencontainers.image.version = $(python -c "f = open("version.txt", "r"); print(f.read().strip())")
 LABEL org.opencontainers.image.source = "https://github.com/hpcaitech/ColossalAI"
 LABEL org.opencontainers.image.licenses = "Apache License 2.0"
-LABEL org.opencontainers.image.base.name="docker.io/library/hpcaitech/cuda-conda:11.3"
+LABEL org.opencontainers.image.base.name = "docker.io/library/hpcaitech/cuda-conda:11.3"
 
 # install torch
 RUN conda install pytorch==1.12.1 torchvision==0.13.1 torchaudio==0.12.1 cudatoolkit=11.3 -c pytorch
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,11 @@
 FROM hpcaitech/cuda-conda:11.3
 
+# metainformation
+LABEL org.opencontainers.image.version = "0.2.5"
+LABEL org.opencontainers.image.source = "https://github.com/hpcaitech/ColossalAI"
+LABEL org.opencontainers.image.licenses = "Apache License 2.0"
+LABEL org.opencontainers.image.base.name="docker.io/library/hpcaitech/cuda-conda:11.3"
+
 # install torch
 RUN conda install pytorch==1.12.1 torchvision==0.13.1 torchaudio==0.12.1 cudatoolkit=11.3 -c pytorch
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM hpcaitech/cuda-conda:11.3
 
+# metainformation
+LABEL org.opencontainers.image.version = "0.2.5"
```
FrankLeeeee, line 4: Can this be made auto-fetched instead of hard-coded? The version is always specified in the `version.txt`.

SauravMaheshkar: One possible solution could be to have inline python i.e.

```Dockerfile
LABEL org.opencontainers.image.version = $(python -c "import colossalai; print(colossalai.__version__)")
```

---

Another option could be to pass the version as an argument while building the docker image i.e.

```Dockerfile
ARG PKGVERSION=X.X.X
...
LABEL org.opencontainers.image.version = ${BASE_IMAGE}
```

and then build using

```bash
docker build --build-arg PKGVERSION= <...>
```

FrankLeeeee: I personally think the last method is the easiest for the following reaons:
1. For the first method, since `colossalai` is installed  after the `LABEL` declaration, we cannot import `colossalai` when fetching the version ahead of time
2. For the second method, I find it feasible but not really user friendly since many users directly run `docker build` and  do not take care about the arguments much. The second method requires the user to check the content of `version.txt` and then do `docker build`, which might be troublesome for users who are not familiar with this requirement.

As the last method is really automatic, I would recommend to proceed with it. Feel free to share your thoughts as well. 😃

FrankLeeeee: On a side note, we are making our development plan public this week. We'd really like to invite you to be part of the development if you are interested.

SauravMaheshkar: @FrankLeeeee that sounds interesting. How can I get involved ???

SauravMaheshkar: Also, if we plan to use the last method, I would really like to use the `get_version()` fn from `setup.py` but that's not really inline code. Can we just move the `version` label until after `colossalai` is installed ?

FrankLeeeee: Hey, @SauravMaheshkar , you just join our development channel via the link on this page.

https://github.com/hpcaitech/ColossalAI/wiki/2.-Developer-Guideline

SauravMaheshkar: > Hey, @SauravMaheshkar , you can join our development channel via the link on this page.
> 
> https://github.com/hpcaitech/ColossalAI/wiki/2.-Developer-Guideline

Joined 👍🏽 @FrankLeeeee

FrankLeeeee: Great.

FrankLeeeee: I have tried your commit, seem it is not working as expected. When running `docker inspect`, I get the following output.

Screenshot 2023-03-13 at 10 16 15

As you can see `org.opencontainers.image.version` does not actually resolve the bash command.

SauravMaheshkar: @FrankLeeeee tbh this makes sense. I checked my reference and even that codebase dropped this after a while. My suggestion would be to drop the `org.opencontainers.image.version` label for now, and rather version the images using docker tags.

FrankLeeeee: Ok, that's fine for me.

FrankLeeeee (approved):

PR #3058, opened 2023-03-08
The Hugginface binary 13B file has been fixed... it properly downloads the file and starts the training process.

tomekrut:
In the OPT 13B model there was an exception handler based on the presumption that Hggingface had a problem with this file binary. It is no longer the case as it properly downloads the model.


1SAA (approved):

binmakeswell:
Hi @tomekrut Thank you very much for your contribution!

PR #3178, opened 2023-03-20
[hotfix] Improved flash attention usage

kurisusnowdeng:
Previous flash attention APIs are in a complete mess, and very hard to use in practice. The new API `ColoAttention` here only imports kernels from `xformers` rather than the multiple sources we previously used. It is able to cover the most common cases in Transformer models, such as:

- Causal attention (e.g. GPT, OPT)
- Bidirectional attention with padded mask (e.g. BERT)
- Bidirectional attention with no mask (e.g. VIT)
- Relative position embedding (e.g. Transformer-XL/BERT)
- Alibi (e.g. BLOOM)
- Cross attention

Test results are shown below:
image

### DIFF
--- a/colossalai/kernel/cuda_native/flash_attention.py
+++ b/colossalai/kernel/cuda_native/flash_attention.py
@@ -1,12 +1,6 @@
 """
-The triton-based flash attention implementation is copied from the OpenAI/triton repository
-
-You can find the repository in Triton https://github.com/openai/triton
-You can find the source file in https://github.com/openai/triton/blob/main/python/tutorials/06-fused-attention.py
-
-Reference:
-1. Dao et al., https://arxiv.org/pdf/2205.14135v2.pdf
-2. Rabe and Staats https://arxiv.org/pdf/2112.05682v2.pdf
+A general attention module using the flash attention kernels from xformers:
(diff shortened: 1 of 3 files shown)


colossalai/kernel/cuda_native/flash_attention.py
```diff
@@ -1,12 +1,6 @@
...
-The triton-based flash attention implementation is copied from the OpenAI/triton repository
-
-You can find the repository in Triton https://github.com/openai/triton
-You can find the source file in https://github.com/openai/triton/blob/main/python/tutorials/06-fused-attention.py
-
-Reference:
-1. Dao et al., https://arxiv.org/pdf/2205.14135v2.pdf
-2. Rabe and Staats https://arxiv.org/pdf/2112.05682v2.pdf
+A general attention module using the flash attention kernels from xformers:
+https://github.com/facebookresearch/xformers/tree/main/xformers/ops/fmha
```
ver217, line 3: Should we add copyright and update license?

ver217 (approved):

colossalai/kernel/cuda_native/flash_attention.py
kurisusnowdeng: License added

PR #3623, opened 2023-04-22
[chat] fix value computing bug

### LINKED ISSUE #3622: [BUG]: [chat] Using wrong part of values to compute value
### 🐛 Describe the bug

[https://github.com/hpcaitech/ColossalAI/blob/b0ce5a10326912961f0bc07cbbd250bab7b9c399/applications/Chat/coati/models/base/critic.py#L45-L50](https://github.com/hpcaitech/ColossalAI/blob/b0ce5a10326912961f0bc07cbbd250bab7b9c399/applications/Chat/coati/models/base/critic.py#L45-L50)

We should use the generation part of the sequenses to compute value. So the indexing should be `[-num_actions:]` instead of `[:-num_actions]` and using `action_mask`.

### Environment

_No response_

zhang-yi-chi:

## 🚨 Issue number

fixed #3622 

## 📝 What does this PR do?

[https://github.com/hpcaitech/ColossalAI/blob/b0ce5a10326912961f0bc07cbbd250bab7b9c399/applications/Chat/coati/models/base/critic.py#L45-L50](https://github.com/hpcaitech/ColossalAI/blob/b0ce5a10326912961f0bc07cbbd250bab7b9c399/applications/Chat/coati/models/base/critic.py#L45-L50)

We should use the generation part of the sequenses to compute value. So the indexing should be `[-num_actions:]` instead of `[:-num_actions]` and using `action_mask`.

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.

### DIFF
--- a/applications/Chat/coati/experience_maker/naive.py
+++ b/applications/Chat/coati/experience_maker/naive.py
@@ -24,7 +24,7 @@ class NaiveExperienceMaker(ExperienceMaker):
         action_log_probs = self.actor(sequences, num_actions, attention_mask)
         base_action_log_probs = self.initial_model(sequences, num_actions, attention_mask)
         value = self.critic(sequences, action_mask, attention_mask)
-        r = self.reward_model(sequences, attention_mask)
+        r = self.reward_model(sequences, attention_mask, action_mask=action_mask)
         reward = compute_reward(r, self.kl_coef, action_log_probs, base_action_log_probs, action_mask=action_mask)
 
         advantage = reward - value
--- a/applications/Chat/coati/models/base/critic.py
+++ b/applications/Chat/coati/models/base/critic.py
@@ -44,7 +44,7 @@ class Critic(LoRAModule):
 
         if action_mask is not None and self.use_action_mask:
             num_actions = action_mask.size(1)
-            values = values[:, -num_actions:]
+            values = values[:, -num_actions - 1:-1]
             value = masked_mean(values, action_mask, dim=1)
             return value
 
--- a/applications/Chat/coati/models/base/critic.py
+++ b/applications/Chat/coati/models/base/critic.py
@@ -44,9 +44,8 @@ class Critic(LoRAModule):
 
         if action_mask is not None and self.use_action_mask:
             num_actions = action_mask.size(1)
-            prompt_mask = attention_mask[:, :-num_actions]
-            values = values[:, :-num_actions]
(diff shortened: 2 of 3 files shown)


applications/Chat/coati/experience_maker/naive.py
```diff
@@ -24,7 +24,7 @@ def make_experience(self, input_ids: torch.Tensor, **generate_kwargs) -> Experie
         action_log_probs = self.actor(sequences, num_actions, attention_mask)
         base_action_log_probs = self.initial_model(sequences, num_actions, attention_mask)
         value = self.critic(sequences, action_mask, attention_mask)
-        r = self.reward_model(sequences, attention_mask)
+        r = self.reward_model(sequences, attention_mask, action_mask=action_mask)
```
ht-zhou, line 27: To compute reward，we should use the whole sequence as training rm

applications/Chat/coati/models/base/critic.py
```diff
@@ -44,9 +44,8 @@ def forward(self,
 
         if action_mask is not None and self.use_action_mask:
```
ht-zhou, line 45: Using action mask here means to only use state (prompt) to compute value, as described in RL patterns.

ht-zhou:
Also we are updating our PPO pattern recently, the modeling of MDP will be changed soon.

applications/Chat/coati/models/base/critic.py
zhang-yi-chi: Sorry for my misunderstanding. So here uses one step MDP if I understand this correctly? Maybe we should change to multi step MDP and compute values for every answer input.

applications/Chat/coati/experience_maker/naive.py
zhang-yi-chi: Yes you are right, here we should follow the rm training. 

This remind me of another issue: either log-sig or log-exp computes (chosen_reward - reject_reward) which actually remove the prompt's reward if we ignore the average computation.  Does this means the prompt part of the sequence are under Inadequate training?

zhang-yi-chi:
> Also we are updating our PPO pattern recently, the modeling of MDP will be changed soon.

Looking forward to seeing the changes. Closed this pull request.

ht-zhou:
Anyway, thanks for your contribution!

Issue #3678, opened 2023-05-04
[PROPOSAL]: Unlimiting optimizer types in ZeRO3


yhna940:
### Proposal

Currently ZeRO3 limits the types of [optimizers](https://github.com/hpcaitech/ColossalAI/blob/1a60dc07a8a6ef6980ffb913b8cbed795de92bff/colossalai/zero/gemini/gemini_optimizer.py#L77).

Why?

 If possible, why not have the `zero_allow_untested_optimizer` option like DeepSpeed ​​does?

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

PR #3699, opened 2023-05-06
[gemini] Remove Type Restriction for Zero3 Optimizers

yhna940:

## 🚨 Issue number

> `Discussions #3449 `
> `PROPOSAL #3678 `

## 📝 What does this PR do?

I have been using Gemini Zero in my projects and found it to be an incredibly useful and powerful tool. However, from a user perspective, I noticed that the choice of optimizers is quite limited, which can be a bit inconvenient.

The reason for this inconvenience is that the gradient generated is FP16, and the master weight is FP32.

Inspired by the `low-level Zero optimizer`, I have made changes to cast the gradient to the master data type. 

https://github.com/hpcaitech/ColossalAI/blob/2629f9717dbb1f0dc0ecad3f3bda18a4b44e3785/colossalai/zero/low_level/low_level_optim.py#L468

I have successfully tested the changes with the standard PyTorch `Adam` optimizer, and the forward , backward and step work without any issues.

I would be grateful if you could provide feedback on my proposed changes. 

Please let me know if you have any questions or require further clarification.

Thank you for your time and consideration.

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.

### DIFF
--- a/colossalai/zero/gemini/gemini_optimizer.py
+++ b/colossalai/zero/gemini/gemini_optimizer.py
@@ -55,6 +55,7 @@ class ZeroOptimizer(ColossalaiOptimizer):
         norm_type (float, optional): The type of norm used for gradient clipping. Currently, only L2-norm (norm_type=2.0)
             is supported in ZeroOptimizer. Defaults to 2.0.
         verbose (bool, optional): Whether to print verbose information, including grad overflow info. Defaults to False.
+        allow_untested_optimizer (bool, optional): Whether to allow using untested optimizer. Defaults to True.
     """
 
     def __init__(self,
@@ -71,11 +72,10 @@ class ZeroOptimizer(ColossalaiOptimizer):
                  clipping_norm: float = 0.0,
                  norm_type: float = 2.0,
                  verbose: bool = False,
+                 allow_untested_optimizer: bool = True,
                  **defaults: Any):
         super().__init__(optim)
         assert isinstance(module, ZeroDDP)
-        assert type(optim) in _AVAIL_OPTIM_LIST, "You should use an optimizer in the available list:\n" \
-            f"{_AVAIL_OPTIM_LIST}"
         self.module = module
         self.gemini_manager = module.gemini_manager
         self.chunk_manager: ChunkManager = self.gemini_manager.chunk_manager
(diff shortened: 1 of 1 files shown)


colossalai/zero/gemini/gemini_optimizer.py
```diff
@@ -86,6 +86,12 @@ def __init__(self,
         self.clipping_flag = clipping_norm > 0.0
         self.max_norm = clipping_norm
         self.verbose = verbose
+        self.untested_optim_flag = True if type(optim) not in _AVAIL_OPTIM_LIST else False
```
ver217, line 89: I think this line can be simplified as `self.untested_optim_flag = type(optim) not in _AVAIL_OPTIM_LIST`

colossalai/zero/gemini/gemini_optimizer.py
```diff
@@ -137,9 +143,8 @@ def _set_grad_ptr(self):
                 begin, end = self.param_to_range[fake_param]
                 chunk16 = chunk32.paired_chunk
 
-                fake_param.data = chunk16.payload[begin:end]
-                fake_param.grad = fake_param.data
                 fake_param.data = chunk32.payload[begin:end]
+                fake_param.grad = chunk16.payload[begin:end].to(chunk32.dtype)
```
ver217, line 147: Why cast grad to fp32 here? This will use extra memory, as hybrid adam will cast grad to fp32 one by one during optimizing.

colossalai/zero/gemini/gemini_optimizer.py
yhna940: Thanks for the correction. I immediately reflected.

colossalai/zero/gemini/gemini_optimizer.py
yhna940:
```suggestion
                grad_payload = chunk16.payload[begin:end]
                if self.untested_optim_flag:
                    grad_payload = grad_payload.to(chunk32.dtype)         
                fake_param.grad = grad_payload
```

Introduced casting to work with normal torch optimizers (the purpose of this PR). Normal PyTorch optimizers do not accept FP16 gradients and FP32 parameters. I cast the type to update FP32 via FP16 gradients as the [low level zero optimizer does](https://github.com/hpcaitech/ColossalAI/blob/f83ea813f5bb6ecc5597c7f1bf97870d46de1c49/colossalai/zero/low_level/low_level_optim.py#LL468C21-L468C21). Since Colossal's optimizer supports hybrid tensors, why not just cast gradients for untested optimziers? Please suggest me if there is a better way. thank you.

PR #4048, opened 2023-06-20
[WIP] support tensor bias in flash attention

WXY9901:
The official implementation of flash attention includes a triton version that is able to support tensor bias such as Alibi in BLOOM models. However, it is not compatible with dropout while there may be some issue of robustness as well.

This PR is trying to modify the CUDA version of flash attention in order to support tensor bias.

Still in progress.


kurisusnowdeng:
@WXY9901 Remember to add the copyright of the original repo in our codes, as well as the its license.

PR #4153, opened 2023-07-04
[zero] allow passing process group to zero12

### LINKED ISSUE #4151: [zero] pass process_group for low level zero
The `gpc` and the `ColoProcessGroup` would be removed after.
Therefore, leaving an interface that allows users to pass a process_group is needed to support the new ProcessGroupMesh.

For compatible with current codes, if no procss_group is passing, the default one would be use.

Gy-Lu:

## 🚨 Issue number

>

fixed #4151 

## 📝 What does this PR do?

> Summarize your work here.

This PR support passing dp and tp process_group to low_level_zero, which further supports the new processGroupMesh.

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.

### DIFF
--- a/colossalai/zero/low_level/low_level_optim.py
+++ b/colossalai/zero/low_level/low_level_optim.py
@@ -75,8 +75,8 @@ class LowLevelZeroOptimizer(OptimizerWrapper):
             overlap_communication: bool = False,
             partition_grad: bool = False,    # stage 2 flag
             cpu_offload: bool = False,    # cpu offload
-            dp_process_group: ProcessGroup = None,    # the dp pg for comm
-            tp_process_group: ProcessGroup = None,    # if using tp
+            dp_process_group: Optional[ProcessGroup] = None,    # the dp pg for comm
+            tp_process_group: Optional[ProcessGroup] = None,    # if using tp
             forced_dtype: Optional[torch.dtype] = None):
 
         # TODO:
--- a/colossalai/zero/low_level/low_level_optim.py
+++ b/colossalai/zero/low_level/low_level_optim.py
@@ -54,8 +54,8 @@ class LowLevelZeroFP16MixedPrecisionMixin(FP16MixedPrecisionMixin):
         return False
 
 
-class TPLowLevelZeroOptimizer(OptimizerWrapper):
-    """Optimizer used for ZeRO-1 and ZeRO-2 with Tensor Parallelism.
+class LowLevelZeroOptimizer(OptimizerWrapper):
+    """Optimizer used for ZeRO-1 and ZeRO-2.
     """
 
     def __init__(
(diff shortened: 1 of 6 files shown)


colossalai/zero/low_level/low_level_optim.py
```diff
@@ -506,3 +469,47 @@ def no_sync(self):
             yield
         finally:
             self.require_grad_sync = old_require_grad_sync
+
+
+class LowLevelZeroOptimizer(TPLowLevelZeroOptimizer):
```
FrankLeeeee, line 474: If this is like a complete inheritance, why not just make it the same class?

Gy-Lu: In fact, there is a tiny difference.
`LowLevelZeroOptimizer` would not receive the kwarg `tp_process_group`, while `TPLowLevelZeroOptimizer` would.
The reason is quite simple: passing `tp_process_group` for naive ZeRO is quite weird.

Gy-Lu: Let me make it more clear:
For naive ZeRO, it's weird that you can pass a `tp_process_group` to it because in common sense, tp has nothing to do with ZeRO.
However, passing a `tp_process_group` to a class named "TP" and "ZeRO" seems to be much more reasonable.

I would write a README for the usage.

FrankLeeeee: I mean you can make tp_process_group default to None in the `LowLevelZeroOptimizer`?

Gy-Lu: You are right.
The two classes may introduce unnecessary workloads for users as well.
I'll change it.

colossalai/zero/low_level/low_level_optim.py
```diff
@@ -96,30 +95,12 @@ def __init__(
...
-            self._world_size = gpc.get_world_size(dp_parallel_mode)
-            self._dp_global_ranks = gpc.get_ranks_in_group(dp_parallel_mode)
-            self._dp_torch_group = gpc.get_group(dp_parallel_mode)
-            self._mp_torch_group = None
-            if gpc.is_initialized(mp_parallel_mode) and gpc.get_world_size(mp_parallel_mode) > 1:
-                self._mp_torch_group = gpc.get_group(mp_parallel_mode)
-        else:
-            raise NotImplementedError
+        # if process_group is none, will use the default one
+        self.dp_pg = dp_process_group
```
ver217, line 99: If it's None, we should use global group to avoid errors.

Gy-Lu: If `self.dp_pg` is None, `torch.distributed` would use the default group.

ver217: OK, then the type hint should be `Optional[ProcessGroup]`

Gy-Lu: Done.

PR #4429, opened 2023-08-14
[moe] support low level zero optim

oahzxl:

## 🚨 Issue number

>

## 📝 What does this PR do?

> Summarize your work here.
moe param should not be stored in working_groups because they have different parallel strategy. so we need to store them separately in param_groups instead of working_groups

- [ ] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.


colossalai/zero/low_level/low_level_optim.py
```diff
@@ -418,6 +434,11 @@ def step(self, closure=None):
         # update the parameters
         self.optim.step()
 
+        # release the moe grad
+        if len(self.param_groups) > len(self._working_param_groups):
+            for param in self.param_groups[-1]['params']:
+                param.grad = None
+
         # release the grad
```
ver217, line 442: Can you integrate moe grad handle into this optimizer? As we won't have any grad handler anymore.

oahzxl: this grad handle is only for pytorch model test. will not be used in our model

ver217 (changes requested):
By the way, can you create a moe param interface like dtensor? https://github.com/hpcaitech/ColossalAI/blob/main/colossalai/tensor/d_tensor/api.py

colossalai/zero/low_level/low_level_optim.py
ver217: Then can we move this to `tests/`? Engine will be deprecated soon.

oahzxl: ok

oahzxl:
> By the way, can you create a moe param interface like dtensor? https://github.com/hpcaitech/ColossalAI/blob/main/colossalai/tensor/d_tensor/api.py

done

ver217 (approved):

PR #4593, opened 2023-09-03
Fix:Added BatchEncoding support to to_device method under trainer.utils

### LINKED ISSUE #4489: [BUG]: to_device don't support BatchEncoding
### 🐛 Describe the bug

```python
from transformers.tokenization_utils_base import BatchEncoding

def to_device(x: Any, device: torch.device) -> Any:
    
    def _to(t: Any):
        if isinstance(t, BatchEncoding):
            for k in t.keys():
                t[k] = t[k].to(device)
            return t
        elif isinstance(t, torch.Tensor):
            return t.to(device)
        return t

    return tree_map(_to, x)
```
Use code above to fix bug that [to_device doesn't support `BatchEncoding`](https://github.com/hpcaitech/ColossalAI/blob/main/applications/Chat/coati/trainer/utils.py#L44).  Because the dataloader may return `BatchEncoding`.

### Environment

_No response_

AdarshAcharya5:
## 📝 What does this PR do?
Added support for batch_encoding for to_device method based on Issue #4489 
Fixes #4489

### DIFF
--- a/applications/Chat/coati/trainer/utils.py
+++ b/applications/Chat/coati/trainer/utils.py
@@ -41,10 +41,16 @@ def is_rank_0() -> bool:
     return not dist.is_initialized() or dist.get_rank() == 0
 
 
-def to_device(x: Any, device: torch.device) -> Any:
+from transformers.tokenization_utils_base import BatchEncoding
 
+def to_device(x: Any, device: torch.device) -> Any:
+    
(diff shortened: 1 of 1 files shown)


TongLi3701:
Hi, thanks for your fix.

Could you please address the conflict first. Thanks

PR #4637, opened 2023-09-06
[moe] support openmoe train

oahzxl:
* support openmoe train for zero2 and zero2+ep
* rewrite checkpoint for ckpt load and save
* use chunk and checkpoint to reduce lm_head's high activation memory due to large vocab size
* update model details to adapt to train

TODO: align result and accuracy

### DIFF
--- a/colossalai/context/moe_context.py
+++ b/colossalai/context/moe_context.py
@@ -21,6 +21,7 @@ class MoeContext(metaclass=SingletonMeta):
         self.max_ep_size = None
         self.min_dp_size = None
         self.aux_loss = None
+        self.parallel = None
         self.use_kernel_optim = True
 
         self.has_setup = False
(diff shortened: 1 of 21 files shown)


colossalai/nn/layer/moe/checkpoint.py
```diff
@@ -7,7 +8,7 @@
 from torch.optim import Optimizer
 
 from colossalai.checkpoint_io import CheckpointIO
-from colossalai.tensor.moe_tensor.api import get_ep_group
+from colossalai.tensor.moe_tensor.api import get_dp_rank, get_ep_group, get_ep_rank, get_ep_size, is_moe_tensor
 
 
 class MoeCheckpintIO(CheckpointIO):
```
ver217, line 14: Maybe you should use `GeneralCheckpointIO`.

oahzxl: done

colossalai/context/moe_context.py
```diff
@@ -21,6 +21,7 @@ def __init__(self):
         self.max_ep_size = None
         self.min_dp_size = None
```
ver217, line 22: This class does not have `__enter__` or `__exit__` method. It seems that it's not a context manager. By the way, why is it asingleton?

oahzxl: it is just a gloabl value to store some settings and is designed by haichen

ver217: Can you move it to `colossalai/moe`? And `colossalai/nn/layer/moe` can move there too.

oahzxl: ok. i will do it in the following pr

ver217 (approved):

PR #4748, opened 2023-09-18
[moe] support hybrid parallel

oahzxl:
support hybrid parallel with pp + dp + ep
* add custom shard policy for openmoe
* add moe hybrid plugin which modifies some init settings of original hybrid plguin
* updata parallel process group mesh in moe
* use layernorm kernel and jit fuse in shardformer

### DIFF
--- a/colossalai/booster/plugin/moe_hybrid_parallel_plugin.py
+++ b/colossalai/booster/plugin/moe_hybrid_parallel_plugin.py
@@ -1,226 +1,38 @@
 import random
-from contextlib import nullcontext
-from functools import partial
-from typing import Any, Callable, Iterator, List, Optional, OrderedDict, Tuple, Union
+from typing import Any, Callable, Iterator, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
 import torch.distributed as dist
-from torch.distributed import ProcessGroup
-from torch.nn import Module, SyncBatchNorm
-from torch.nn.parallel import DistributedDataParallel as DDP
+from torch.nn import Module
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
-from torch.utils._pytree import tree_map
 from torch.utils.data import DataLoader
 from torch.utils.data.distributed import DistributedSampler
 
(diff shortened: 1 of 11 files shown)


colossalai/booster/plugin/moe_hybrid_parallel_plugin.py
```diff
@@ -0,0 +1,339 @@
...
+from colossalai.pipeline.stage_manager import PipelineStageManager
+from colossalai.shardformer import ShardConfig
+from colossalai.shardformer.policies.base_policy import Policy
+
+from .pp_plugin_base import PipelinePluginBase
+
+PP_AXIS, DP_AXIS, TP_AXIS = 0, 1, 2
+
+
+class MoeHybridParallelPlugin(PipelinePluginBase):
```
ver217, line 33: What's the difference between this plugin and hybrid parallel plugin?

oahzxl: their process grouph axis is different

ver217: Then can it inherit HybridParallelPlugin and overwrite a few methods like `__init__()`?

examples/language/openmoe/test_ci.sh
```diff
@@ -1,5 +0,0 @@
-set -xe
-pip install -r requirements.txt
-
-python infer.py --model "test"
-torchrun --standalone --nproc_per_node 2 train.py --model_name "test" --batch_size 1 --num_epoch 20
```
ver217, line 5: When can you run ci?

colossalai/booster/plugin/moe_hybrid_parallel_plugin.py
oahzxl: ok

examples/language/openmoe/test_ci.sh
oahzxl: ci fails due to some kernel initialization. will be updated soon

CWHer (approved):

Issue #4833, opened 2023-09-28
[FEATURE]: shardformer to support interleaved pipeline parallel


ppt0011:
### Describe the feature

shardformer currently only supports non-interleaved 1f1b pipeline parallel

PR #4448, opened 2023-08-16
[shardformer] support interleaved pipeline

Gy-Lu:
## 📌 Checklist before creating the PR

- [x] I have created an issue for this PR for traceability
- [x] The title follows the standard format: `[doc/gemini/tensor/...]: A concise description`
- [x] I have added relevant tags if possible for us to better distinguish different PRs

## 🚨 Issue number

> Link this PR to your issue with words like fixed to automatically close the linked issue upon merge
>
> e.g. `fixed #1234`, `closed #1234`, `resolved #1234`

closed #4447 

## 📝 What does this PR do?

> Summarize your work here.
> if you have any plots/diagrams/screenshots/tables, please attach them here.

1. This PR mainly supports interleaved schedule with the new coordinate.
2. Weaken the p2p ops. Things like examing if the first/last stage should be done in schedule.
3. Support a ring group, e.g. the prev_rank of rank0 is the last rank, and the next_rank of the last rank is rank0.
4. Remove useless codes.

## 💥 Checklist before requesting a review

- [x] I have linked my PR to an issue ([instruction](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue))
- [x] My issue clearly describes the problem/feature/proposal, with diagrams/charts/table/code if possible
- [x] I have performed a self-review of my code
- [x] I have added thorough tests.
- [x] I have added docstrings for all the functions/methods I implemented

## ⭐️ Do you enjoy contributing to Colossal-AI?

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.

Tell us more if you don't enjoy contributing to Colossal-AI.

PR #4834, opened 2023-09-28
[shardformer] support interleaved pipeline parallel for Bert finetune example

ppt0011:

## 🚨 Issue number
#4833 
>

## 📝 What does this PR do?

> Summarize your work here.

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.


colossalai/shardformer/policies/bert.py
```diff
@@ -275,11 +287,28 @@ def get_held_layers(self) -> List[Module]:
...
-        start_idx, end_idx = self.get_stage_index(layers_per_stage, stage_manager.stage)
-        held_layers.extend(module.encoder.layer[start_idx:end_idx])
+        if num_model_chunks > 1:
+            stage_index = Policy.get_stage_index(
+                layers_per_stage, stage_manager.stage, stage_manager.num_stages, num_model_chunks
+            )
+        else:
+            stage_index = Policy.get_stage_index(layers_per_stage, stage_manager.stage)
+
+        # interleaved stage index for one device comes in pairs, e.g.[[0,3],[6,9]]
```
ppt0011, line 303: this illustrates the main idea for interleaved pipeline parallel. take bert for example, the original non-interlaved bert layer distribution among the two pipeline parallel devices are: device0=[0,6], device1=[6,12]; for interleaved, we now have: device0=[[0,3],[6,9]], deivce1=[[3,6],[9,12]]

colossalai/booster/plugin/hybrid_parallel_plugin.py
```diff
@@ -376,22 +380,39 @@ def __init__(
         self.stage_manager = None
         self.schedule = None
         self.custom_policy = custom_policy
+        self.num_microbatches = num_microbatches
```
ver217, line 383: It seems that this attribute is not used.

colossalai/pipeline/schedule/interleaved_pp.py
```diff
@@ -184,7 +206,15 @@ def forward_step(
 
         # for the first stage, input_obj is None
         # for the non-first stage, input_obj is the output of the previous stage and it's must be a dict
-        output_obj = model_forward(model_chunk[model_chunk_id], micro_batch, input_obj)
+
+        if isinstance(model_chunk, ModuleList):
+            output_obj = model_forward(model_chunk[model_chunk_id], micro_batch, input_obj)
+        else:
```
ver217, line 212: When will it be normal module?

colossalai/pipeline/stage_manager.py
```diff
@@ -18,7 +18,9 @@ class PipelineStageManager:
         stage (int): The current stage.
     """
 
-    def __init__(self, pg_mesh: ProcessGroupMesh, pipeline_axis: int, is_virtual: bool = False) -> None:
+    def __init__(
+        self, pg_mesh: ProcessGroupMesh, pipeline_axis: int, is_virtual: bool = False, num_model_chunks=1
```
ver217, line 22: It seems that `is_virtual` is not used.

colossalai/shardformer/policies/base_policy.py
```diff
@@ -214,13 +221,30 @@ def distribute_layers(num_layers: int, num_stages: int) -> List[int]:
...
-        start_idx = num_layers_per_stage_accumulated[stage]
-        end_idx = num_layers_per_stage_accumulated[stage + 1]
-
-        return [start_idx, end_idx]
+    def get_stage_index(
+        layers_per_stage: List[int], stage: int, num_stages=None, num_model_chunks=1
+    ) -> Union[List[int], List[List[int]]]:
+        # num_stages info is only needed for interleaved pipeline stage assignment
+        if num_stages is None:
+            """
```
ver217, line 229: Use `# xxxx`

colossalai/pipeline/stage_manager.py
```diff
@@ -32,6 +34,9 @@ def __init__(self, pg_mesh: ProcessGroupMesh, pipeline_axis: int, is_virtual: bo
         # the next rank of the last rank is rank0
         next_coord = coord[: self.pipeline_axis] + (coord[self.pipeline_axis] + 1,) + coord[self.pipeline_axis + 1 :]
         self.next_rank = self.pg_mesh.ravel(next_coord, self.pg_mesh.shape, mode="wrap")
+        # number of layer chunks in each stage for interleaved pipeline, with each device has discontinuous layers
+        self.num_model_chunks = num_model_chunks
```
ver217, line 38: Can you keep these attributes in schedule?

colossalai/shardformer/policies/bert.py
```diff
@@ -251,13 +251,34 @@ def set_pipeline_forward(self, model_cls: nn.Module, new_forward: Callable, poli
...
+                    "forward": partial(
+                        new_forward,
+                        stage_manager=stage_manager,
+                        stage_index=stage_index,
+                        shard_config=self.shard_config,
+                    )
+                }
+            # for interleaved, stage index for each forward is chosen in scheduler
+            else:
+                self.scheduler.layers = stage_index
```
ver217, line 278: Can it be derived from stage and num model chunks?

colossalai/pipeline/stage_manager.py
ppt0011: It seems to me that stage manager should have model chunk id information. As interleave added, it is only the last stage if the device is the last device _AND model chunk is the last chunk_. Notably, we now make a difference between last device and last stage https://github.com/hpcaitech/ColossalAI/blob/ad33868e8b8d9359b1b7eca20da27583fa2d6f32/colossalai/pipeline/stage_manager.py#L73

colossalai/shardformer/policies/bert.py
ppt0011: yes, that was actually my initial implementation. I changed it to avoid the need to modify the forward function of all other shardformer models.

if we want to sync the implementation style for interleaved and non-interleaved layer assignment for the foward function, I think the better way is to stop passing layer information as part of the partial function, since we no longer know it in advance. I can raise another PR for non-interleaved 1f1b schedule to pass layer information in the scheduler instead if that works, something like https://github.com/hpcaitech/ColossalAI/blob/ad33868e8b8d9359b1b7eca20da27583fa2d6f32/colossalai/pipeline/schedule/interleaved_pp.py#L216

colossalai/booster/plugin/hybrid_parallel_plugin.py
ppt0011: I followed 1f1b here https://github.com/hpcaitech/ColossalAI/blob/8aed02b957082e676cf424068583c78240b27f9d/colossalai/pipeline/schedule/one_f_one_b.py#L42 user can pass in either microbatch_size  or num_microbatches.

colossalai/pipeline/schedule/interleaved_pp.py
ppt0011: It is actually a normal module for the bert finetune example, because we pass in the entire model for each stage (not a list of modules that the current stage is responsible for), just that the layers which the current stage is not responsible for have their weights set to none. 

It is a module list for the previously written test_interleaved.py script.

colossalai/pipeline/stage_manager.py
ppt0011: it works together with a previous commit https://github.com/hpcaitech/ColossalAI/pull/4448/files#
specifically https://github.com/hpcaitech/ColossalAI/blob/8aed02b957082e676cf424068583c78240b27f9d/colossalai/pipeline/stage_manager.py#L44

colossalai/shardformer/policies/base_policy.py
ppt0011: re-formatted

colossalai/booster/plugin/hybrid_parallel_plugin.py
ver217: This is stored in schedule. Why store here again?

colossalai/pipeline/stage_manager.py
ver217: Current code does not use it. Just delete it.

colossalai/pipeline/schedule/interleaved_pp.py
```diff
@@ -367,10 +400,10 @@ def forward_backward_step(
         if not forward_only:
             for i in range(num_microbatches_remaining, num_microbatches):
                 model_chunk_id = self.get_model_chunk_id(i, forward=False)
+                self.stage_manager.model_chunk_id = model_chunk_id
```
ver217, line 403: Add a method to set this attr

colossalai/shardformer/policies/bert.py
ver217, line 278: Add a method to set this attr

colossalai/booster/plugin/hybrid_parallel_plugin.py
ppt0011: I see. Removed from plugin attr

colossalai/pipeline/stage_manager.py
ppt0011: retained is_virtual as discussed

colossalai/shardformer/policies/bert.py
ppt0011: moved layers attr to stage manager and added a set method there

ver217 (approved):

PR #4877, opened 2023-10-10
[moe] support hybrid zero strategy. 

oahzxl:
* Support hybrid zero strategy. 
  * For non-moe param, zero will be activated for the global dp group. 
  * For moe param, we will use expert parallel and zero, where ep group * zero group = global dp group. So zero will be activated only for moe zero group here.
* add unit test for hybrid zero
* enable moe unit test
* fix previous unit test

### DIFF
--- a/colossalai/booster/plugin/hybrid_parallel_plugin.py
+++ b/colossalai/booster/plugin/hybrid_parallel_plugin.py
@@ -209,14 +209,14 @@ class HybridParallelZeroOptimizer(LowLevelZeroOptimizer):
             dp_process_group: Optional[ProcessGroup] = None,    # the dp pg for comm
             tp_process_group: Optional[ProcessGroup] = None,    # if using tp
             forced_dtype: Optional[torch.dtype] = None,
-            outer_dp_process_group: Optional[ProcessGroup] = None):
+            inner_dp_process_group: Optional[ProcessGroup] = None):
         self.param_info = param_info
         if use_pipeline:
             init_pipeline_optimizer(optimizer, model)
         super().__init__(optimizer, initial_scale, min_scale, growth_factor, backoff_factor, growth_interval,
                          hysteresis, max_scale, clip_grad_norm, verbose, reduce_bucket_size, communication_dtype,
                          overlap_communication, partition_grad, cpu_offload, dp_process_group, tp_process_group,
-                         forced_dtype, outer_dp_process_group)
+                         forced_dtype, inner_dp_process_group)
 
 
 class HybridParallelPlugin(PipelinePluginBase):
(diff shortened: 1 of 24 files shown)


colossalai/zero/low_level/low_level_optim.py
```diff
@@ -279,27 +298,73 @@ def _run_reduction(self):
...
+                    else:
+                        # record moe and non moe param
+                        moe_list = []
+                        for param in self._bucket_store._param_list:
+                            moe_list.append(is_moe_tensor(param))
+
+                        # divide them into different groups
+                        moe_grad_list = []
+                        non_moe_grad_list = []
+                        for rank, grad_list in self._bucket_store._grad_in_bucket.items():
```
ver217, line 335: `rank` is not used, why not use `xxx.values()`? By the way, you can also `zip(self._bucket_store._param_list, self._bucket_store._grad_in_bucket.values()`

oahzxl: ok, fixed.

colossalai/booster/plugin/hybrid_parallel_plugin.py
```diff
@@ -208,14 +208,15 @@ def __init__(
             cpu_offload: bool = False,    # cpu offload
             dp_process_group: Optional[ProcessGroup] = None,    # the dp pg for comm
             tp_process_group: Optional[ProcessGroup] = None,    # if using tp
-            forced_dtype: Optional[torch.dtype] = None):
+            forced_dtype: Optional[torch.dtype] = None,
+            extra_dp_process_group: Optional[ProcessGroup] = None):
```
ver217, line 212: If this is only used with MOE, how about moving this to moe hybriad parallel plugin?

colossalai/moe/manager.py
```diff
@@ -140,6 +140,10 @@ def get_info(self, num_experts: int, use_tp: bool = False) -> Tuple[int, MoePara
 
         if not (ep_size in self.parallel_info_dict):
             self.parallel_info_dict[ep_size] = get_moe_info(ep_size, dp_size, pp_size, ep_inside=self.use_ep_inside)
+            if self.use_ep_inside:
+                print(f"MoE Parallel: pp {pp_size}, dp {dp_size}, ep {ep_size}")
```
ver217, line 144: Please ensure this is only printed on one process

colossalai/booster/plugin/hybrid_parallel_plugin.py
oahzxl: done

colossalai/moe/manager.py
oahzxl: done

ver217 (approved):

PR #4914, opened 2023-10-15
[moe] support load balance

oahzxl:
* support load balance for moe
* add a unit test

### DIFF
--- a/colossalai/moe/layers.py
+++ b/colossalai/moe/layers.py
@@ -2,15 +2,17 @@ import math
 from typing import Optional, Tuple
 
 import torch
+import torch.distributed as dist
 import torch.nn as nn
 import torch.nn.functional as F
 
 from colossalai.moe._operation import AllGather, AllToAll, MoeCombine, MoeDispatch, ReduceScatter
 from colossalai.moe.experts import BaseMLPExperts, get_expert_class
+from colossalai.moe.load_balance import LoadBalancer
 from colossalai.moe.manager import MOE_MANAGER
 from colossalai.moe.routers import MoeRouter, get_router_cls
 from colossalai.moe.utils import get_noise_generator
-from colossalai.tensor.moe_tensor.api import get_ep_group, get_ep_size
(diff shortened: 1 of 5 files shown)


colossalai/moe/layers.py
```diff
@@ -257,3 +285,19 @@ def get_chunk_slice(idx: int, gap: int) -> Tuple[slice]:
...
+
+    def _apply_recursive(module: nn.Module):
+        for _, sub_module in module.named_children():
+            if isinstance(sub_module, SparseMLP):
+                if sub_module.enable_load_balance == True:
+                    sub_module.load_balancer.balance_load(optim)
+            _apply_recursive(sub_module)
+
+    _apply_recursive(model)
+    dist.barrier()
```
ver217, line 303: Is it safe to synchornize on all GPUs? How about pipeline parallelism?

colossalai/moe/load_balance.py
```diff
@@ -0,0 +1,429 @@
...
+        """
+        swap_loss = 0
+        for swap in swap_list:
+            for group_id, index_id in zip([group_i, group_j], [index_i, index_j]):
+                # the group has been swapped
+                if group_id in [swap[0], swap[2]]:
+                    # the index has been swapped
+                    # we want to avoid the situation that the same index is swapped twice
+                    if index_id in [swap[1], swap[3]]:
+                        swap_loss += 1e5
```
ver217, line 94: What's the effect of `1e5`?

colossalai/moe/layers.py
oahzxl: you're right. i remove it since it's unnecessary

colossalai/moe/load_balance.py
oahzxl: If it tries to swap an expert twice, it will lead to bad performance. so i give it a high cost to make it the ranked at bottom when choosing.

ver217 (approved):

PR #4923, opened 2023-10-16
[moe] support bf16 and update benchmark

oahzxl:
* support bf16 to avoid overflow
* update benchmark
  * support mock dataset
  * use tokenizor in dataset
  * add load balance option

### DIFF
--- a/examples/language/openmoe/benchmark/benchmark_cai.py
+++ b/examples/language/openmoe/benchmark/benchmark_cai.py
@@ -1,3 +1,4 @@
+import json
 import os
 
 import torch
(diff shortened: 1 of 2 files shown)


examples/language/openmoe/benchmark/benchmark_cai.py
```diff
@@ -264,6 +301,9 @@ def main():
             optimizer.step()
             optimizer.zero_grad()
             performance_evaluator.on_step_end(exmaple_data["input_ids"])
+            if (step == args.warmup // 2) and args.load_balance:
```
ver217, line 304: Is this the standard method to use load balance?

oahzxl: it is just for benchmark. it will be updated to train later. we will check the load every n steps and do update if load is not balanced

ver217 (approved):

PR #5432, opened 2024-03-07
[Inference] Finish Online Serving Test, add streaming output api, continuous batching test and example

CjhHa1:

## 🚨 Issue number

>

## 📝 What does this PR do?

> Summarize your work here.
1. complete completion service api, add streaming output
2. finish test, locust test up to 8000 users
3. add examples and tests

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.


colossalai/inference/server/api_server.py
```diff
@@ -52,13 +52,14 @@ async def generate(request: Request) -> Response:
     stream = request_dict.pop("stream", None)
 
     request_id = id_generator()
+    print("generate_id", request_id)
```
ver217, line 55: remove it

examples/inference/client/run_locust.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# launch server
+model_path="/home/caidi/llama_model/"
```
ver217, line 4: Don't use hard-coded path. Use CLI argument.

colossalai/inference/server/completion_service.py
```diff
@@ -1,10 +1,18 @@
 import asyncio
+from typing import AsyncIterator
 
 from colossalai.inference.core.async_engine import AsyncInferenceEngine
 
 from .utils import id_generator
 
 
+async def completion_stream_generator(
+    result_generator: AsyncIterator,
```
ver217, line 10: Is this function essential? It seems that it just create a new async generator from an existing async generator

CjhHa1: It is needed as we should return a async generator when using streaming response, while not, we return a single  content.(which is also `generator` in this case).

examples/inference/client/run_locust.sh
CjhHa1: fixed

colossalai/inference/server/api_server.py
CjhHa1: done

colossalai/inference/core/async_engine.py
```diff
@@ -63,7 +63,7 @@ class RequestTracker:
     """Synchronous abstraction for tracking requests."""
 
     def __init__(self) -> None:
-        self._request_streams: Dict[str, AsyncStream] = {}
+        self._request_streams: Dict[int, AsyncStream] = {}
         self._finished_requests: asyncio.Queue[int] = asyncio.Queue()
```
FrankLeeeee, line 67: Would this mean that each request will have its own stream which creates a new `asyncio.Queue()`？

examples/inference/client/run_locust.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+#argument1: model_path
```
FrankLeeeee, line 3: You should give a default value.

```bash
model_path=${1:-"lmsys/vicuna-7b-v1.3"}
```

colossalai/inference/core/async_engine.py
CjhHa1: Actually right

FrankLeeeee: I think it might be quite inefficient? Do you think there is better implementation?

examples/inference/client/run_locust.sh
CjhHa1: fixed

colossalai/inference/core/async_engine.py
CjhHa1: it is true, but change this would cause a different response, I will check that

tests/test_infer/test_continous_batching.py
```diff
@@ -0,0 +1,92 @@
...
+    max_input_len: int = 64
+
+    inputs_token_ids = generate_inputs(10 * max_batch_size, min_length=10, max_length=max_input_len)
+
+    do_sample = True
+    top_p = 0.5
+    top_k = 50
+
+    if use_engine:
+        inference_config = InferenceConfig(max_output_len=max_output_len, prompt_template=prompt_template)
```
yuanheng-zhao, line 48: Explicitly set `max_batch_size` even if the config has a default value. And you might want to make the following vars parameterized to this test:
```python
    max_batch_size: int = 8
    max_output_len: int = 512
    max_input_len: int = 64
```

tests/test_infer/test_continous_batching.py
```diff
@@ -0,0 +1,92 @@
...
+        inference_engine = InferenceEngine(model, tokenizer, inference_config, verbose=True)
+        assert inference_engine.generation_config.max_new_tokens == max_output_len
+        inference_engine.add_request(prompts_token_ids=inputs_token_ids)
+        assert inference_engine.request_handler._has_waiting()
+        generation_config = GenerationConfig(do_sample=do_sample, top_p=top_p, top_k=top_k)
+        outputs = inference_engine.generate(generation_config=generation_config)
```
yuanheng-zhao, line 54: Wonder is there a better way to test continuous batching? In this test we could verify the engine is able to handle a pool of requests more than its maximum batch size, but could hardly demonstrate the engine processes those requests in a continuous-batching way, instead of processing batches of size `max_batch_size` one by one.

tests/test_infer/test_continous_batching.py
```diff
@@ -0,0 +1,92 @@
...
+            pad_token_id=tokenizer.pad_token_id,
+            max_new_tokens=max_output_len,
+        )
+        outputs = model.generate(inputs, generation_config=generation_config)
+        outputs = tokenizer.batch_decode(outputs, skip_special_tokens=True)
+    assert len(outputs) == 10 * max_batch_size
+
+
+@parameterize("prompt_template", [None, "llama"])
+def check_continous_batching(prompt_template):
```
yuanheng-zhao, line 76: Several typos in this file: continuous. Remember to revise the name of this file as well.

colossalai/inference/core/request_handler.py
```diff
@@ -197,6 +197,7 @@ def schedule(self):
                             break
 
                     num_seqs_to_add = min(len(lst), self.max_batch_size - self.running_list.total_seq_num)
+                    # for now the recycle logic is not working
```
yuanheng-zhao, line 200: It's not working only for async engine?

CjhHa1: not working for all conditions.

tests/test_infer/test_continous_batching.py
CjhHa1: fixed

tests/test_infer/test_continous_batching.py
CjhHa1: the inputs are randomly lengthed, with a actual weight, the model outputs lengths are also different

tests/test_infer/test_continous_batching.py
CjhHa1: fixed

yuanheng-zhao (approved):

PR #5446, opened 2024-03-12
[Auto Parallel]: Speed up intra-op plan generation by 44%

### LINKED ISSUE #5436: [PROPOSAL]: Speed improvement of Intra-Op plan generation in ColossalAuto
### Proposal

Generating an Inter-Op plan with ColossalAuto takes usually a 1-2 minutes when running `examples/tutorial/auto_parallel/auto_parallel_with_resnet.py`. Profiling with cProfile reveals that a large portion of this time is consumed by calling `copy.deepcopy`, especially in the method `DimSpec.build_difference_2d_dict()`. Since many `DimSpec` objects are created[^1], that function is also called hundreds of thousands of times. Upon closer examination of the logic in this function, it becomes apparent that the result of this method is in fact independent of the `DimSpec` object, and its content is not mutated throughout its lifetime. Hence, it suffices to only create this dict once and share it among all instances of `DimSpec`. Due to the large quantity of `DimSpec` instances created throughout the plan generation, this change can introduce a speed-up of up to 50%[^2].

[^1]: many of which are just empty placeholders btw
[^2]: when running `examples/tutorial/auto_parallel/auto_parallel_with_resnet.py`

### Self-service

- [X] I'd be willing to do some initial work on this proposal myself.

stephankoe:

- [X] I have created an issue for this PR for traceability

## 🚨 Issue number

resolves #5436 

## 📝 What does this PR do?

As stated in issue #5436, the generation of a `DimSpec` object is comparatively costly because a dictionary is created every time and a deep copy of two strings is being made. The large volume of `DimSpec` objects created leads to several seconds spent in total by creating `DimSpec` objects when generating intra-op plans with ColossalAuto.

This pull requests addresses this inefficiency by converting this dict to a class attribute of the `DimSpec` class, so it's shared among all its instances. This dict is initialized lazily the first time the property `difference_dict` is used. This is possible because the contents of `difference_dict` are not modified by other portions of the code.

Additionally, the methods `build_difference_2d_dict` and `convert_str_to_shard_list` are made class/static methods and private because they don't need access to the instance properties.

### Effect

These changes reduce the end-to-end wall-clock time to build the strategy constructor by **44%** while running the script `examples/tutorial/auto_parallel/auto_parallel_with_resnet.py`. The time required to compute a solution on  a laptop with Intel Core i7 7700HQ was on average 56.79 s before the change. After applying above-mentioned changes, the wall-clock time reduced to about 31.93 s.

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.


colossalai/tensor/sharding_spec.py
```diff
@@ -43,24 +43,43 @@ def __repr__(self):
...
+        Return:
+            difference(int): the difference between two _DimSpec.
+
+        Example:
+            dim_spec = _DimSpec([0])
+            other_dim_spec = _DimSpec([0, 1])
+            print(dim_spec.difference(other_dim_spec))
+
+        Output:
+            5
```
Edenzzzz, line 76: This seems to be an old typo. Remove or put "int"?

colossalai/tensor/d_tensor/sharding_spec.py
```diff
@@ -39,24 +39,43 @@ def __repr__(self):
...
+        Return:
+            difference(int): the difference between two _DimSpec.
+
+        Example:
+            dim_spec = _DimSpec([0])
+            other_dim_spec = _DimSpec([0, 1])
+            print(dim_spec.difference(other_dim_spec))
+
+        Output:
+            5
```
Edenzzzz, line 72: Old typo

Edenzzzz:
Thanks for the PR! DimSpec is used extensively outside in TP, so overall looks good to me.

Edenzzzz (changes requested):

colossalai/tensor/sharding_spec.py
stephankoe: So, should it changed as follows?

```
Output: 
    (int) 5
```

Edenzzzz: Oh I missed it, the return signature is above. So this is actually fine

Edenzzzz:
cc @ver217  looks good?

colossalai/tensor/sharding_spec.py
stephankoe: I see. I'll nevertheless push a change to this docstring, because it uses an old class name.

ver217 (approved):

Edenzzzz (approved):

PR #5767, opened 2024-05-31
[moe/zero] refactor low level optimizer

botbw:
# 📝 What does this PR do?

## TLDR: Refactor low-level ZeRO optimizer to better deal with param groups with different parallel strategies.

Basically, the original `LowLevelZeroOptimizer` was decoupled into a few classes (`LowLevelZeroOptimizer`, `LowLevelOptStrategyBase`, `LowLevelOptStrategy`, `MoeZeroStrategy`), where each `param_group` and `process_group` are bonded into one `LowLevelOptStrategyBase`.

`LowLevelOptStrategyBase` utilizes previous `XXXStore`s and manages the grad communication and synchronization in low level, and the children classes handle the grad manipulation  (between `working_params` and `master_params`) in tensor level.

In this way, different param_groups might use different zero and comm policies and do not interfere with each other, and they can have their own comm logic.

Other things we can do: refactor `XXXStore`s, previous `XXXStore`s manage all `param_groups` by `group_id`, which is a bit complex. however currently this logic is moved to `LowLevelOptStrategyBase` and `group_id == 0` is used by each store (search `DEFAULT_STORE_GROUP_ID`).

WARNING: Reviewing this is going to be painful.

### DIFF
--- a/tests/test_moe/test_moe_zero_fwd_bwd.py
+++ b/tests/test_moe/test_moe_zero_fwd_bwd.py
@@ -1,0 +1,107 @@
+import pytest
+import torch
+import torch.distributed as dist
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+import colossalai
+from colossalai.moe.manager import MOE_MANAGER
+from colossalai.tensor.moe_tensor.api import is_moe_tensor
+from colossalai.testing import rerun_if_address_is_in_use, spawn
+from colossalai.testing.random import seed_all
+from colossalai.zero.low_level.low_level_optim import LowLevelZeroOptimizer
+from colossalai.zero.low_level.low_level_strategy import LowLevelOptStrategy, MoeZeroStrategy
+from tests.test_moe.moe_utils import MoeModel, delete_moe_info, loose_close, sync_local_from_ep
+
+
+def run_zero_test(local_rank):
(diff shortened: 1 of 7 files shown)


tests/test_moe/test_moe_zero_fwd_bwd.py
botbw, line 1: Note: This test cannot run since we no longer have `MoeModel` on the `feature/moe` branch.

tests/test_moe/test_moe_zero_optim.py
botbw, line 1: Note: This test cannot run since we no longer have `MoeModel` on the `feature/moe` branch.

tests/test_moe/test_moe_load_balance.py
```diff
@@ -176,6 +176,7 @@ def run_dist(rank, world_size, port):
     run_hybrid_zero_optim_test(rank, world_size, stage=2)
 
 
+@pytest.mark.skip(reason="moe need to be refactored")
```
botbw, line 179: Note: This test cannot run since we no longer have `MoeModel` on the `feature/moe` branch.

Hz188 (approved):

Hz188 (approved):

PR #5785, opened 2024-06-06
[Inference]Lazy Init Support

LRY89757:
This PR support the Lazy Init of Llama2, Llama 3 and Baichuan(partially).
> NOTE：for baichuan model, the weight of lm_head loaded by lazy init has some difference with transformers model, which results the output diff with transformers output. 

Lazy Init supports both local path and remote path.

Lazy Init could accelerate the process of model loading.

For Llama 70B, fp32, tensor parallel=4, It only take 3.3 min to load the model. Without Lazy init, program will freeze due to limited CPU Memory.


colossalai/inference/executor/rpc_worker.py
```diff
@@ -179,19 +181,18 @@ def _init_model(self, model_or_path: Union[nn.Module, str], model_policy: Policy
         """
 
         if isinstance(model_or_path, str):
-            # is_local = os.path.isdir(model_or_path)
+            is_local = os.path.isdir(model_or_path)
             try:
-                hf_config = AutoConfig.from_pretrained(model_or_path, trust_remote_code=True)
+                hf_config = AutoConfig.from_pretrained(model_or_path, trust_remote_code=True, torch_dtype=self.dtype)
                 arch = getattr(hf_config, "architectures")[0]
```
yuanheng-zhao, line 187: What if the input model name or path is not supported yet? Please refer to the usage in `engine.py`

colossalai/inference/executor/rpc_worker.py
```diff
@@ -179,19 +181,18 @@ def _init_model(self, model_or_path: Union[nn.Module, str], model_policy: Policy
...
-                # NOTE(lry89757) Currently we load the model using transformers-api,
-                # but we will use lazy tensor and checkpoint io to accelerate
-                # the model load process in the future.
-                model = _SUPPORTED_MODELS[arch].from_pretrained(model_or_path, trust_remote_code=True)
-                # if is_local:
-                #     model = _SUPPORTED_MODELS[arch](hf_config)
-                # else:
-                #     # load the real checkpoint
-                #     model = _SUPPORTED_MODELS[arch].from_pretrained(model_or_path, trust_remote_code=True)
+                if is_local:
```
yuanheng-zhao, line 188: For this method (`_init_model`), please also keep up to date with that (`init_model`) in `engine.py`. For example, the handling of condition of no policy is provided.

colossalai/inference/modeling/models/nopadding_llama.py
```diff
@@ -298,6 +298,10 @@ def __init__(
         self.gate_up_weight = nn.Parameter(
             torch.stack([mlp_gproj_w.transpose(0, 1), mlp_uproj_w.transpose(0, 1)], dim=0)
         )
+        self.gate_up_dict = {
+            "gate_proj.weight": None,
+            "up_proj.weight": None,
+        }  # used and delattr in load/shard of gate/up weight
```
yuanheng-zhao, line 304: Wonder why we initialize the dictionary here? It seems you only use it when loading from state dict

colossalai/inference/modeling/models/nopadding_llama.py
```diff
@@ -335,44 +339,47 @@ def _load_from_state_dict(
     ):
         # NOTE This is a hack to ensure we could load the right weight from LlamaMLP checkpoint due to the use of torch.stack(gate_weight, up_weight)
 
-        for hook in self._load_state_dict_pre_hooks.values():
-            hook(state_dict, prefix, local_metadata, strict, missing_keys, unexpected_keys, error_msgs)
-
-        persistent_buffers = {k: v for k, v in self._buffers.items() if k not in self._non_persistent_buffers_set}
-        local_name_params = itertools.chain(self._parameters.items(), persistent_buffers.items())
-        local_state = {k: v for k, v in local_name_params if v is not None}
+        if hasattr(self, "gate_up_dict"):
```
yuanheng-zhao, line 342: 1. With `self.gate_up_dict` already set in `__init__`, the if block is noneffective and redundant. 
2. prevent hard-coding

colossalai/inference/modeling/models/nopadding_llama.py
```diff
@@ -335,44 +339,47 @@ def _load_from_state_dict(
...
+            key = "gate_up_weight"
+            device_mesh = self.helper_layout.device_mesh
+            sharding_spec = self.helper_layout.sharding_spec
+            for weight_name in self.gate_up_dict:
+                prefix_weight_name = prefix + weight_name
+                if prefix_weight_name in state_dict.keys():
+                    w = distribute_tensor(state_dict[prefix_weight_name], device_mesh, sharding_spec)
+                    self.gate_up_dict[weight_name] = w.T
+
+            if None not in self.gate_up_dict.values():
```
yuanheng-zhao, line 359: Do we need to handle the other condition?

colossalai/inference/modeling/models/nopadding_llama.py
```diff
@@ -335,44 +339,47 @@ def _load_from_state_dict(
...
-        key = "gate_up_weight"
-        k1 = "gate_proj.weight"
-        k2 = "up_proj.weight"
+            persistent_buffers = {k: v for k, v in self._buffers.items() if k not in self._non_persistent_buffers_set}
+            local_name_params = itertools.chain(self._parameters.items(), persistent_buffers.items())
+            local_state = {k: v for k, v in local_name_params if v is not None}
 
-        gate_w = state_dict[prefix + k1]
-        up_w = state_dict[prefix + k2]
+            key = "gate_up_weight"
```
yuanheng-zhao, line 350: You could remove key assignment and directly use it later in `local_state.get`, though it's still sort of hard-coding...

colossalai/inference/modeling/models/nopadding_llama.py
```diff
@@ -439,7 +446,15 @@ def __init__(
             self.helper_layout = (
                 attn_qproj_w.dist_layout
             )  # NOTE this is a hack for the right load/shard of qkv_weight(used in _load_from_state_dict)
+            self.qkv_dict = {
+                "q_proj.weight": None,
+                "k_proj.weight": None,
+                "v_proj.weight": None,
```
yuanheng-zhao, line 452: Same as above. If the structure is only used in a single-time-calling function, we don't need to set it as object attributes

colossalai/inference/modeling/models/nopadding_llama.py
```diff
@@ -646,49 +661,85 @@ def forward(
...
+            if prefix + "q_proj.weight" in state_dict.keys():
+                _load(origin_weight_name="q_proj.weight", local_weight_name="q_proj_weight")
+
+            if prefix + "k_proj.weight" in state_dict.keys():
+                _load(origin_weight_name="k_proj.weight", local_weight_name="k_proj_weight")
+
+            if prefix + "v_proj.weight" in state_dict.keys():
+                _load(origin_weight_name="v_proj.weight", local_weight_name="v_proj_weight")
+
+            strict = False  # to avoid unexpected_keys
```
yuanheng-zhao, line 741: Redundant assignment

colossalai/inference/executor/rpc_worker.py
LRY89757: We will not call the remote worker if the engine doesn't support this model.
The logic you mentioned is in `rpc_engine` instead of `rpc_worker`.
If in rpc_worker, the error message will show many times(there are many rpc_workers. @yuanheng-zhao

colossalai/inference/modeling/models/nopadding_llama.py
LRY89757: Because of the problems we discussed before, q,k,v will be stored in diff checkpoint files, so hardcode is inevitable :(.

colossalai/inference/modeling/models/nopadding_llama.py
LRY89757: No

colossalai/inference/modeling/models/nopadding_llama.py
LRY89757: Same as above.

colossalai/inference/modeling/models/nopadding_llama.py
LRY89757: q,k,v will be stored in diff checkpoint files, we need to maintain a global dict to store the q,k,v. So hardcode is inevitable here :(.

colossalai/inference/executor/rpc_worker.py
LRY89757: Mentioned Above

colossalai/inference/core/engine.py
```diff
@@ -122,16 +123,27 @@ def init_model(
...
-                hf_config = AutoConfig.from_pretrained(model_or_path, trust_remote_code=True)
+                hf_config = AutoConfig.from_pretrained(model_or_path, trust_remote_code=True, torch_dtype=self.dtype)
                 arch = getattr(hf_config, "architectures")[0]
                 if arch in _supported_models.keys():
-                    # NOTE(lry89757) Currently we load the model using transformers-api,
-                    # but we will use lazy tensor and checkpoint io to accelerate
-                    # the model load process in the future.
-                    model = _supported_models[arch].from_pretrained(model_or_path, trust_remote_code=True)
+                    if arch is "BaichuanForCausalLM":
+                        self.logger.info(
```
yuanheng-zhao, line 135: 1. Logging level supposed to be warning
2. You could put these two statements into a single one

colossalai/inference/executor/rpc_worker.py
```diff
@@ -178,20 +179,24 @@ def _init_model(self, model_or_path: Union[nn.Module, str], model_policy: Policy
...
-                # if is_local:
-                #     model = _SUPPORTED_MODELS[arch](hf_config)
-                # else:
-                #     # load the real checkpoint
-                #     model = _SUPPORTED_MODELS[arch].from_pretrained(model_or_path, trust_remote_code=True)
+                if arch is "BaichuanForCausalLM":
+                    self.logger.info(
+                        "Attention ! We use lazy init by default, which could be faster for model loading."
+                    )
+                    self.logger.info("For baichuan model, the output maybe have a slight difference with transformers")
```
yuanheng-zhao, line 193: Same as above.

PR #5838, opened 2024-06-19
[Feat] Refactor Engine and Diffusion Model(PixArtAlpha/StableDiffusion3) Support

LRY89757:

## 📝 What does this PR do?

Support Stable Diffusion 3 for our ColossalAI InferenceEngine

- [x] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.


colossalai/inference/config.py
```diff
@@ -393,3 +393,64 @@ class ModelShardInferenceConfig:
...
+    prompt_2: Optional[Union[str, List[str]]] = None
+    prompt_3: Optional[Union[str, List[str]]] = None
+    height: Optional[int] = None
+    width: Optional[int] = None
+    num_inference_steps: int = 28
+    timesteps: List[int] = None
+    guidance_scale: float = 7.0
+    negative_prompt: Optional[
+        Union[str, List[str]]
+    ] = ""  # NOTE(@lry89757) in pixart default to "", in sd3 default to None, we set "" here, but the best solution is change the check_inputs of each forward fuction
```
yuanheng-zhao, line 414: For type hint defined as `Optional[...]` in python, we usually use `None` as its default value.

yuanheng-zhao: trivial: And you might want to place relatively long comment to a different line (not inline) so that the formatter won't reformat and place it to a longer line with statements

colossalai/inference/config.py
```diff
@@ -393,3 +393,64 @@ class ModelShardInferenceConfig:
...
+        for field_info in fields(self):
+            name = field_info.name
+            value = getattr(self, name)
+
+            if field_info.default is not MISSING:
+                default = field_info.default
+            else:
+                default = None
+
+            if field_info.default_factory is not MISSING:
```
yuanheng-zhao, line 443: In common cases we should avoid the import and usage of `dataclasses.MISSING`. Refer to https://docs.python.org/3/library/dataclasses.html

colossalai/inference/config.py
```diff
@@ -393,3 +393,64 @@ class ModelShardInferenceConfig:
...
+
+@dataclass
+class GenerationParams:
+    """
+    Param for diffusion
+    """
+
+    # prompt: Union[str, List[str]] = None # NOTE no sure if we should add this param
+    prompt_2: Optional[Union[str, List[str]]] = None
+    prompt_3: Optional[Union[str, List[str]]] = None
```
yuanheng-zhao, line 406: I'm a bit confused with these attributes. The purpose of `GenerationParams` is to provide default generation configs for diffusion models, right? Then is it appropriate to have default prompts for the second and the third tokenizers, or are there extra explanations/descriptions for these two prompts?

colossalai/inference/config.py
```diff
@@ -393,3 +393,64 @@ class ModelShardInferenceConfig:
     use_cuda_kernel: bool = False
     use_spec_dec: bool = False
     use_flash_attn: bool = False
+
+
+@dataclass
+class GenerationParams:
```
yuanheng-zhao, line 399: trivial: A class name identifying its purpose and usage for diffusion models is preferred.

colossalai/inference/struct.py
```diff
@@ -46,6 +47,17 @@ def is_waiting(status: "RequestStatus") -> bool:
         return status == RequestStatus.WAITING
 
 
+@dataclass
+class Sequence_:
```
yuanheng-zhao, line 51: Same as above. The naming is causing confusion; please use a meaningful and clear name. You could even create a baes sequence class with commonly used attrs such as request id and prompt, and modify, for example, the `DiffusionSequence` and `LLMSequence`,  to inherit from the base class.

colossalai/inference/core/engine.py
```diff
@@ -3,8 +3,10 @@
 from typing import Dict, List, Optional, Tuple, Type, Union
 
 import numpy as np
+import PIL.Image
```
yuanheng-zhao, line 6: **About the design of Engine**: 

I recommend to split the two engine instances into two files, saying that, LLMEngine and DiffusionEngine, which inherit from a base abstract class. In this way, you are able to:
1. retain some abstraction level and split the logics of executing into different places
2. instantiate certain engine during initialization
3. **prevent dispatching step method** (and other methods applying conditions) every time of inference

yuanheng-zhao: This will apply to modifications along this file

colossalai/inference/core/request_handler.py
```diff
@@ -98,7 +98,46 @@ def move_prefill_to_decoding(self, seq_ids: List[int]) -> None:
...
+
+    def schedule(self):
+        ret = None
+        if self._has_waiting:
+            ret = self.waiting_list[0]
+            self.waiting_list = self.waiting_list[1:]
+        return ret
+
+
+class RequestHandler(NaiveRequestHandler):
```
yuanheng-zhao, line 140: In your inheritance and modification, calling `RequestHandler._has_running` will cause an TypeError, though it's never invoked and the error never exposed.

colossalai/inference/utils.py
```diff
@@ -158,3 +159,22 @@ def can_use_flash_attn2(dtype: torch.dtype) -> bool:
...
+
+
+def get_model_type(model_or_path: Union[nn.Module, str, DiffusionPipeline]):
+    if isinstance(model_or_path, DiffusionPipeline):
+        return ModelType.DIFFUSION_MODEL
+    elif isinstance(model_or_path, nn.Module):
+        return ModelType.LLM
+    elif isinstance(model_or_path, str):
+        # TODO(lry89757) we need to support string input for diffusion model
+        return ModelType.LLM
```
yuanheng-zhao, line 180: Please add and handle else condition.

colossalai/inference/utils.py
```diff
@@ -158,3 +159,22 @@ def can_use_flash_attn2(dtype: torch.dtype) -> bool:
...
+from enum import Enum
+
+
+class ModelType(Enum):
+    DIFFUSION_MODEL = "Diffusion Model"
+    LLM = "Large Language Model (LLM)"
+    UNKNOWN = "Unknown Model Type"
+
+
+def get_model_type(model_or_path: Union[nn.Module, str, DiffusionPipeline]):
```
yuanheng-zhao, line 173: The str input won't work in your case handling.

colossalai/inference/config.py
LRY89757: Thx!

colossalai/inference/config.py
LRY89757: Thanks!

colossalai/inference/struct.py
LRY89757: thanks!

colossalai/inference/core/engine.py
LRY89757: thanks!

colossalai/inference/utils.py
LRY89757: thanks!

colossalai/inference/utils.py
LRY89757: thanks!

colossalai/inference/utils.py
```diff
@@ -158,3 +161,24 @@ def can_use_flash_attn2(dtype: torch.dtype) -> bool:
...
+    if isinstance(model_or_path, DiffusionPipeline):
+        return ModelType.DIFFUSION_MODEL
+    elif isinstance(model_or_path, nn.Module):
+        return ModelType.LLM
+    elif isinstance(model_or_path, str):
+        try:
+            hf_config = AutoConfig.from_pretrained(model_or_path, trust_remote_code=True)
+            return ModelType.LLM
+        except:
+            pass
```
yuanheng-zhao, line 182: Please expose the exception as well

requirements/requirements.txt
```diff
@@ -23,3 +23,4 @@ rpyc==6.0.0
 fastapi
 uvicorn==0.29.0
 galore_torch
+diffusers==0.29.0
```
yuanheng-zhao, line 26: To be discussed: whether to add diffusers into requirements or not.

yuanheng-zhao (approved):
LGTM. Left some comments and partially approved.

colossalai/inference/utils.py
```diff
@@ -158,3 +161,24 @@ def can_use_flash_attn2(dtype: torch.dtype) -> bool:
...
+    # TODO(lry89757) we need to support string input for diffusion model
```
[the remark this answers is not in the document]
LRY89757: Solution Found, we could use like:
```python
try:
 from diffusers import DiffusionPipeline
 cfg = DiffusionPipeline.load_config(model_or_path)
 return ModelType.DIFFUSION_MODEL
except:
 pass
```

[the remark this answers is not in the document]
LRY89757: It is a very lightweight way.

colossalai/inference/utils.py
LRY89757: Thx, this is resolved

PR #5891, opened 2024-07-05
[Hoxfix] Fix CUDA_DEVICE_MAX_CONNECTIONS for comm overlap 

Edenzzzz:

## 📝 What does this PR do?
Setting CUDA_DEVICE_MAX_CONNECTIONS=1 can force the kernel launch order to be the same as on the host, which means communication will be launched before computation instead of the other way. 
Previously, the variable was probably set after the distributed environment (NCCL etc.) is set up, so it had no effect. 
image

I've tested that minor CUDA operations such as torch.ones(1) before launching colossalai will not override this variable. 
This ensures async_grad_allreduce in TP works and increases throughput by around 5% with TP 2 PP 2 on H800.
image

- [ ] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.

### DIFF
--- a/colossalai/legacy/nn/layer/parallel_1d/_operation.py
+++ b/colossalai/legacy/nn/layer/parallel_1d/_operation.py
@@ -81,7 +81,6 @@ class LinearWithAsyncCommunication(torch.autograd.Function):
             handle = dist.all_reduce(grad_input, group=gpc.get_group(ctx.parallel_mode), async_op=True)
             # Delay the start of weight gradient computation shortly (3us) to have
             # all-reduce scheduled first and have GPU resources allocated
-            _ = torch.empty(1, device=grad_output.device) + 1
 
         grad_weight = grad_output.t().matmul(total_input)
         grad_bias = grad_output.sum(dim=0) if use_bias else None
--- a/colossalai/initialize.py
+++ b/colossalai/initialize.py
@@ -3,6 +3,12 @@
 
 import os
 
+# set CUDA_DEVICE_MAX_CONNECTIONS=1 to ensure that when overlapping communication and computation,
+# the order of of kernel launches on GPUs are the same as on the CPU so that comm is launched first.
+# see https://github.com/NVIDIA/Megatron-LM/issues/533
+# https://forums.developer.nvidia.com/t/how-many-streams-maximum-number-of-streams/6571/16
+os.environ["CUDA_DEVICE_MAX_CONNECTIONS"] = "1"
+
 import torch.distributed as dist
 
 from colossalai.accelerator import get_accelerator
--- a/colossalai/shardformer/shard/shardformer.py
+++ b/colossalai/shardformer/shard/shardformer.py
@@ -1,4 +1,3 @@
-import os
 from typing import Dict, List, Tuple
 
 import torch.distributed as dist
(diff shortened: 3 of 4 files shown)


colossalai/legacy/nn/layer/parallel_1d/_operation.py
```diff
@@ -81,7 +81,6 @@ def backward(ctx, grad_output):
             handle = dist.all_reduce(grad_input, group=gpc.get_group(ctx.parallel_mode), async_op=True)
             # Delay the start of weight gradient computation shortly (3us) to have
             # all-reduce scheduled first and have GPU resources allocated
-            _ = torch.empty(1, device=grad_output.device) + 1
```
ver217, line 84: Comment this line. This may be useful for further refactoring

ver217 (approved):

Edenzzzz:
# NOTE
In rare cases, this can cause false dependencies between comm & compute.
Such as with this script
```
import torch
import torch.distributed as dist
import torch.multiprocessing as mp
import os
def init_process_group(rank=0, world_size=1):
    dist.init_process_group(backend='nccl', rank=rank, world_size=world_size)
class ExampleModule:
    def __init__(self, kv_buffer_idx, buffer_list):
        self.kv_buffer_idx = kv_buffer_idx
        self.buffer_list = buffer_list
        self.buffer_list_refer = [torch.zeros_like(buffer) for buffer in buffer_list]
        self.s = torch.cuda.Stream()
        self.handle = None
    def forward(self, query, kv, batch_size):

        self.buffer_list[self.kv_buffer_idx].copy_(kv)
        
        
        s = self.s
        with torch.cuda.stream(s):
            self.handle = dist.all_gather(self.buffer_list, kv, async_op=True)
            if self.handle is not None:
                self.handle.wait()
            full_kv = torch.cat(self.buffer_list, dim=1)
        key, value = torch.split(full_kv, full_kv.shape[1] // 2, dim=1)
        hidden_states = self.compute_attention(query, key, value, batch_size)
        return hidden_states
    
    def compute_attention(self, query, key, value, batch_size):
        # hidden_states = F.scaled_dot_product_attention(query, key, value)
        hidden_states = query@key.transpose(-1, -2) @ value
        return hidden_states
    
def run_test(rank, world_size):
    os.environ['MASTER_ADDR'] = 'localhost'  # or the IP of the master node
    os.environ['MASTER_PORT'] = '8125'       # make sure this port is free
    init_process_group(rank, world_size)
    torch.manual_seed(rank)
    bs, seqlen, head_num, head_dim = 1024, 128, 16, 256
    buffer_list = [torch.zeros(bs, seqlen // world_size * 2, head_num, head_dim).to(f"cuda:{rank}") for _ in range(world_size)]
    module = ExampleModule(kv_buffer_idx=rank, buffer_list=buffer_list)
    query = torch.randn(bs, seqlen, head_num, head_dim).to(f"cuda:{rank}")
    kv = torch.randn(bs, seqlen // world_size * 2, head_num, head_dim).to(f"cuda:{rank}")
    batch_size = bs
    with torch.profiler.profile(
            record_shapes=True,
            with_stack=True,
            with_modules=True,
            activities=[
                torch.profiler.ProfilerActivity.CPU,
                torch.profiler.ProfilerActivity.CUDA,
            ],
        ) as prof:
        for i in range(10):
            output = module.forward(query, kv, batch_size)
            query.copy_(output)
    prof.export_chrome_trace(f"comp_comm_overlap_rank{rank}.json")
    assert output.shape == (bs, seqlen, head_num, head_dim), f"Unexpected output shape: {output.shape}"
    print(f"Rank {rank}, output shape: {output.shape}")
# CUDA_VISIBLE_DEVICES_set_n_least_memory_usage 2 && CUDA_DEVICE_MAX_CONNECTIONS=1 python test_comp_comm_overlap.py
# CUDA_VISIBLE_DEVICES_set_n_least_memory_usage 2 && python test_comp_comm_overlap.py
# CUDA_VISIBLE_DEVICES=4,5 python test_comp_comm_overlap.py
if __name__ == "__main__":
    world_size = 2
    mp.spawn(run_test, args=(world_size,), nprocs=world_size, join=True)
```

PR #6045, opened 2024-09-03
disable all_to_all_fp8 in intranode 

BurkeHulk:

## 🚨 Issue number

>

## 📝 What does this PR do?

> Summarize your work here.

- [ ] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.


colossalai/quantization/fp8.py
```diff
@@ -22,6 +23,17 @@ def wait(self):
...
 
+def process_group_is_intranode(pg):
+    if pg is None:
+        from torch.distributed.distributed_c10d import _get_default_group
+
+        pg = _get_default_group()
+    local_world_size = int(os.environ["LOCAL_WORLD_SIZE"])
+    group_ranks = list(dist.distributed_c10d._pg_group_ranks[pg].values())
+    group_ranks_node_ids = [rank // local_world_size for rank in group_ranks]
+    return min(group_ranks_node_ids) == max(group_ranks_node_ids)
```
ver217, line 34: There is a simpler method that we can compare device count and world size

colossalai/quantization/fp8.py
ver217, line 1: And for other communication op, we don't actually use fp8 even when user set `fp8_communication=True` to get a better overall performance.

colossalai/quantization/fp8.py
BurkeHulk: Is it possible that the all-to-all process group only includes a subset of ranks? 
E.g. Given 2 nodes, 8 gpus each node. all-to-all is performed within each node.

ver217: Yes, you're right. But the env variable "LOCAL_WORLD_SIZE" is not always reliable. E.g. user may launch with slurm/mpi. Maybe we can use device count to get local world size

BurkeHulk: I see. But I also found that  local_world_size can be smaller than device_count if program is launched with a subset of gpus. E.g. There are 8 gpus per node but `--nproc_per_node=4`. Unless `CUDA_VISIBLE_DEVICES` is set.

ver217: Oh, if so we may try difference env variables.
- https://docs.hpc.shef.ac.uk/en/latest/referenceinfo/scheduler/SLURM/SLURM-environment-variables.html#gsc.tab=0- 
- https://docs.open-mpi.org/en/v5.0.x/tuning-apps/environment-var.html

BurkeHulk: Get. Thanks.

ver217 (approved):

PR #6071, opened 2024-09-25
[Ring Attention] fix the 2d ring attn when using multiple machine

wangbluo:
## 🚨 Issue number

fixed #6017

## 📝 What does this PR do?

The double_ring_groups need to consider the tp groups as the tp axis is the first axis.
And the ranks in double_ring_groups need to transformered into global ranks.
refer: https://arxiv.org/pdf/2406.18485
image
image

### Problem Transformation:
Given a consecutive array from 0 to world_size - 1 and a number num_rings, divide the array into num_rings subarrays by evenly dividing world_size by num_rings.
### Inner Ring Group:
For each subarray, that is a single num_ring, take one number every tp_size elements. The extracted numbers are grouped together every inner_ring_size to form a new element. Continue this process until reaching world_size. The collection of all these new elements forms the inner ring group array.
### Inter Ring Group:
From each subarray, take the numbers at the same index to form a new element. These new elements are combined into a new array, which is the inter ring group.

### Test code:

```
            num_ring_size = world_size//num_rings
            ranks = []
            end = num_ring_size
            for i in range(num_rings):
                start = i * num_ring_size
                end = (i+1) * num_ring_size
                for idx in range(start, end):
                    inner_rank = []
                    for k in range(inner_ring_size):
                        current_num = idx + k * tp_size
                        if current_num >= end:  
                            break
                        inner_rank.append(current_num)
                    if len(inner_rank) == inner_ring_size and inner_rank not in ranks:
                        ranks.append(inner_rank)
            print("inner ranks:",ranks) 
               
            inter_ranks = []
            for i in range(num_ring_size):
                inter_rank = [i + j * num_ring_size for j in range(num_rings)]
                inter_ranks.append(inter_rank)
            print("inter ranks:",inter_ranks)   
```

**example1:**
world size = 8, tp 2, sp 4, inner_ring_size = 2

results:
image

**example2:**
world size = 32, tp 8, sp 4, inner_ring_size = 2, 
2 nodes and 16 gpus in one node

results:
image

**example3:**
world_size = 32  tp_size = 4  sp_size = 8,inner_ring_size = 4

results:
image

**example 4:**
world_size = 24  tp_size = 2  sp_size = 6，inner_ring_size = 3

results:
image


colossalai/shardformer/layer/attn.py
```diff
@@ -471,19 +471,42 @@ def get_double_ring_groups(sp_group, inner_ring_size=None):
...
-
-        # Create inter ring groups
-        for i in range(num_rings):
-            ranks = list(range(i, sp_size, num_rings))
-            group = dist.new_group(ranks)
-            if sp_rank in ranks:
-                inter_ring_group = group
+        world_size = dist.get_world_size()
+        rank = dist.get_rank()
+        groups = int(world_size / sp_size)
```
Edenzzzz, line 476: If you just do `world_size / sp_size`, it will loop over all tp * pp * dp ranks?

colossalai/shardformer/layer/attn.py
```diff
@@ -471,19 +471,42 @@ def get_double_ring_groups(sp_group, inner_ring_size=None):
...
+
+        if tp_size > 1:
```
Edenzzzz, line 478: Perhaps something like this?
```
world_size = 16
sp_size = 4
num_rings = 2
inner_sp_size = sp_size // num_rings
tp_size = 2
groups = world_size // (num_rings * tp_size) # loop through groups of size (inner_ring * tp) 
for i in range(groups):
    # inside each group, duplicate tp group inner_sp_size times
    for j in range(inner_sp_size):
        print(f"inner ring ranks: {list(range(j + i * inner_sp_size * tp_size, (i + 1) * inner_sp_size * tp_size, tp_size))}")
print("---------------------------------")

step_size = tp_size * inner_sp_size
for j in range(inner_sp_size):
    for k in range(tp_size):
        print(f"inner ring ranks: {list(range(j * tp_size + k, world_size, step_size))}")
        
```

colossalai/shardformer/layer/attn.py
```diff
@@ -471,19 +471,42 @@ def get_double_ring_groups(sp_group, inner_ring_size=None):
...
+            for group_id in range(groups):
```
Edenzzzz, line 479: Not totally sure, but it should look like this?
```
world_size = 16
sp_size = 4
outer_ring_size = 2
inner_ring_size = sp_size // outer_ring_size
tp_size = 2
total_inner_rings = world_size // (outer_ring_size * tp_size) # loop through groups of size (inner_ring * tp) 
total_inner_size = tp_size * inner_ring_size
for j in range(total_inner_rings):
    # inside each group, duplicate tp group inner_sp_size times
    for k in range(tp_size):
        print(f"inner ring ranks: {list(range(k + j * total_inner_size, (j + 1) * total_inner_size, tp_size))}")
print("---------------------------------")


sp_tp_size = total_inner_size * outer_ring_size
n_groups = world_size // sp_tp_size # dp * pp
print(f"n_groups: {n_groups}")
for i in range(0, n_groups):
    start = i * sp_tp_size
    end = (i + 1) * sp_tp_size
    for j in range(outer_ring_size):    
        for k in range(tp_size):
            print(f"inter ring ranks: {list(range(start + k + j * tp_size, end, total_inner_size))}")
```
image
image

colossalai/shardformer/layer/attn.py
wangbluo: I don't think so, as the tp axis is the first, and sp axis is the next, see as the HybirdPlugin class. groups means the sp groups number. If sp_size = 8, which means there is 8 gpus in one sp group, and if you use two machines to do the trainning(world size = 16),  the groups = 2. So it should be pp * dp ranks.

colossalai/shardformer/layer/attn.py
wangbluo: Yes, I use 4 gpus in two nodes separately in the example ,see as https://arxiv.org/pdf/2406.18485. 
However, I believe there is an issue with the current algorithm, as such a rank group cannot proceed with the next step of communication. Additionally, in the CI tests, the author only tested it with 4 GPUs, so this PR may need to be closed for now.

wangbluo: If use this config, which is :
inner group:[0, 1, ][2, 3],[4, 5],[ 6, 7],
inter group: [0,4][1,5][2,6][3,7]，
the communication can be proceed and can be training, seems like the current algorithm need the inner group number equals to the inter group number.
However it's wrong, the paper does not impose such requirements, the inner group can be set to 4, and the inter group to 2.

Edenzzzz: I'm not sure I understand--the current group creation is wrong in that it doesn't consider TP and doesn't duplicate groups over dp * pp?
I fixed the code a bit, please test/modify on this and hope it works.
```
world_size = 16
sp_size = 8
outer_ring_size = 2
inner_ring_size = sp_size // outer_ring_size
tp_size = 1
total_inner_size = tp_size * inner_ring_size
total_inner_rings = world_size // total_inner_size # loop through groups of size (inner_ring * tp) 

for j in range(total_inner_rings):
    # inside each group, duplicate tp group inner_sp_size times
    for k in range(tp_size):
        print(f"inner ring ranks: {list(range(k + j * total_inner_size, (j + 1) * total_inner_size, tp_size))}")
print("---------------------------------")


sp_tp_size = total_inner_size * outer_ring_size
n_groups = world_size // sp_tp_size  # dp * pp
print(f"num dp * pp groups: {n_groups}")
for i in range(0, n_groups):
    start = i * sp_tp_size
    end = (i + 1) * sp_tp_size
    for j in range(inner_ring_size):    
        for k in range(tp_size):
            print(f"inter ring ranks: {list(range(start + k + j * tp_size, end, total_inner_size))}")
```
image

colossalai/shardformer/layer/attn.py
Edenzzzz: Yeah but sp_group's size doesn't include tp

colossalai/shardformer/layer/attn.py
wangbluo: Yes but this group config have communication error with subsequent codes, that's the problem.

wangbluo: okay, let me check one more time.

wangbluo: Still NCCL error:
image
image
I agree with your rank group but unluckily this doesn't work, still figure it out.

wangbluo: if sp = 8, world size = 16,
inner group：[0, 1, 2, 3] [4, 5, 6, 7]，
inter group：[0,4] [1,5] [2,6] [3,7]，

if tp = 2, sp = 4, world size = 16,
inner group: [0,2][1,3][4,6][5,7]，
inter group: [0,4][1,5][2,6][3,7]

Edenzzzz: Thanks for checking and good to hear ranks are correct. 
There are two possibilities:
1. Hardware/infiniband issue. This fluctuates from time to time.
2. Using multiple NCCL process groups at the same time--intra and inter ring groups. But the original paper's implementation seems to do the same thing (https://github.com/InternLM/InternEvo/blob/e52f2ffc9acf818e8f2b1f97dfc69ceb2f06e154/internlm/model/ops/ring_flash_attn/zigzag_ring_flash_attn_with_sliding_window.py#L189)
This is as far as I can help, plz discuss further with @ver217 🙂

wangbluo: Thank you for your debugging, don't worry, I just want to make sure the group rank is ok, We are in agreement now. NCCL error is something wrong with the group rank. When the tp size >1, there is no NCCL error, so at least I can use 'if else' to fix that.

Edenzzzz:
The illustration with DP * PP shown
image

colossalai/shardformer/layer/attn.py
```diff
@@ -471,19 +472,50 @@ def get_double_ring_groups(sp_group, inner_ring_size=None):
...
+            for i in range(num_rings):
+                start = i * num_ring_size
+                end = (i + 1) * num_ring_size
+                for idx in range(start, end):
+                    inner_rank = []
+                    for k in range(inner_ring_size):
+                        current_num = idx + k * tp_size
+                        if current_num >= end:
+                            break
+                        inner_rank.append(current_num)
```
Edenzzzz, line 492: Personally I feel it's more concise to use list comprehension here

tests/test_shardformer/test_layer/test_ring_attn.py
```diff
@@ -161,12 +164,12 @@ def check_packed_seq(seqlen, bs, nheads, d, dtype):
 def launch_single_ring(rank, world_size, port):
     colossalai.launch(rank, world_size, "localhost", port)
     check_packed_seq()
-    check_ring_attn()
+    check_ring_attn(sp_size=world_size)
 
 
 def launch_double_ring(rank, world_size, port):
     colossalai.launch(rank, world_size, "localhost", port)
-    check_ring_attn()
```
Edenzzzz, line 169: Perhaps add a single test for the complext rank generation logic(https://github.com/hpcaitech/ColossalAI/blob/703bb5c18dd3a701209017de828e9cdc8ca58356/colossalai/shardformer/layer/attn.py#L483) for `world_size = 16`?
e.g. by putting rank generation into a single function `def get_double_ring_ranks(world_size, tp_size, sp_size, inner_ring_size, dp_pp_size)` and group creation after it.
Up to you🙂

wangbluo: As the ci machine only have 8 gpus so I set world size = 8 to test the tp+sp config for the 2d-ring. Multiple nodes have already been tested locally and showed in the description of this PR below.

Edenzzzz: Sure. It's possible to test the rank creation logic alone without using any GPU in a `world_size=16` example--in case someone messes it up later--but not adding a test case is also fine.

wangbluo: What do you mean 'only rank creation logic'?

Edenzzzz: put the for loops in a function 
```
def get_double_ring_ranks(world_size, tp_size, sp_size, inner_ring_size, dp_pp_size) -> Tuple[List[int]]:
        if tp_size > 1:
            # Create inner ring groups
            ranks = []
            for i in range(num_rings):
                start = i * num_ring_size
                end = (i + 1) * num_ring_size
                for idx in range(start, end):
                    inner_rank = []
                    for k in range(inner_ring_size):
                        current_num = idx + k * tp_size
                        if current_num >= end:
                            break
                        inner_rank.append(current_num)
                    if len(inner_rank) == inner_ring_size and inner_rank not in ranks:
                        ranks.append(inner_rank)
....
```

wangbluo: CI is for not only rank creation but also accuracy test, and there are many rank test cases that I'm afraid one or two rank creation test is not enough. There are four variables impact the rank creation. So I test locally and put in the description.

colossalai/shardformer/layer/attn.py
wangbluo: Okay

wangbluo: fixed

colossalai/shardformer/layer/attn.py
```diff
@@ -444,18 +445,11 @@ def get_double_ring_groups(sp_group, inner_ring_size=None):
...
         sp_rank = dist.get_rank(sp_group)
 
-        if inner_ring_size is None:
-            if torch.cuda.device_count() >= dist.get_world_size():
-                # single node, no need to consider NICs
-                return sp_group, sp_group
-            if sp_size <= 4:
-                inner_ring_size = min(2, sp_size)
-            else:
-                inner_ring_size = min(4, sp_size)
```
Edenzzzz, line 454: We can keep this rule of thumb logic for easy topology construction? Users may have no idea what double ring means and not touch `inner_ring_size` at all

wangbluo: @ver217 said it's better to keep it flexible as we might meet various kinds of machines. The users can use double ring by setting inner_ring_size to a non-none value.

ver217 (approved):

PR #6438, opened 2026-08-20
[fp8]: Read detected local world size environment variable

### LINKED ISSUE #6437: [BUG]: FP8 intranode detection ignores OpenMPI and Slurm local size values
### Bug

`process_group_is_intranode()` checks three environment variables for the local world size, but it always reads `LOCAL_WORLD_SIZE` after finding a match. When a job exposes only the OpenMPI or Slurm variable, FP8 collective setup fails with `KeyError: 'LOCAL_WORLD_SIZE'` before it can determine whether the process group is intranode.

### Minimal reproduction

```python
import os
from unittest.mock import patch

from colossalai.quantization import fp8

os.environ.pop("LOCAL_WORLD_SIZE", None)
os.environ["OMPI_COMM_WORLD_LOCAL_SIZE"] = "4"

with patch.object(fp8.dist, "get_process_group_ranks", return_value=[0, 1, 2, 3]):
    print(fp8.process_group_is_intranode(object()))
```

The same failure occurs when only `SLURM_TASKS_PER_NODE=4` is set.

### Expected behavior

The function should read the value from the environment variable that was detected. If more than one variable is present, the existing order should remain the priority: torchrun, then OpenMPI, then Slurm.

### Environment

- ColossalAI main at `4f9953b`
- Python 3.11
- PyTorch 2.5.1
- macOS arm64 (the failure occurs before any GPU operation)

### Checks

- [x] I searched the existing issues.
- [x] I reproduced this on the latest main branch.
- [x] I included a minimal reproduction.

tandede:

- [x] I have installed pre-commit: `pip install pre-commit && pre-commit install`

## 🚨 Issue number

Fixed #6437

## 📝 What does this PR do?

`process_group_is_intranode()` previously detected the first available local-world-size environment variable but always read `LOCAL_WORLD_SIZE`. Jobs launched with OpenMPI or Slurm therefore raised `KeyError` before FP8 collectives could select the intranode path.

This change reads the detected variable and stops at the first match, preserving the existing priority order: torchrun, OpenMPI, then Slurm.

Regression tests cover:

- torchrun, OpenMPI, and Slurm environment variables independently;
- intranode and cross-node process groups;
- deterministic priority when several variables are present.

Validation:

- `pytest -q --confcutdir=tests/test_fp8 tests/test_fp8/test_fp8_utils.py` — 7 passed
- `pre-commit run --files colossalai/quantization/fp8.py tests/test_fp8/test_fp8_utils.py` — all hooks passed

The repository-level test fixture invokes GPU accelerator cache hooks during collection, so the full GPU suite is left to CI; the new regression itself is CPU-safe.

- [x] I have linked my PR to an issue
- [x] My issue clearly describes the problem and includes a minimal reproduction
- [x] I have added thorough tests
- [ ] I have added docstrings for all the functions/methods I implemented (no new functions or methods)

- [ ] 🌝 Yes, I do.
- [ ] 🌚 No, I don't.

### DIFF
--- a/colossalai/quantization/fp8.py
+++ b/colossalai/quantization/fp8.py
@@ -39,7 +39,8 @@ def process_group_is_intranode(pg):
     local_world_size = None
     for var in ["LOCAL_WORLD_SIZE", "OMPI_COMM_WORLD_LOCAL_SIZE", "SLURM_TASKS_PER_NODE"]:
         if var in os.environ:
-            local_world_size = int(os.environ["LOCAL_WORLD_SIZE"])
+            local_world_size = int(os.environ[var])
+            break
     if local_world_size is None:
         local_world_size = torch.cuda.device_count()
 
--- a/tests/test_fp8/test_fp8_utils.py
+++ b/tests/test_fp8/test_fp8_utils.py
@@ -1,0 +1,29 @@
+import pytest
+
+from colossalai.quantization import fp8
+
+
+@pytest.mark.parametrize(
+    "local_world_size_var",
+    ["LOCAL_WORLD_SIZE", "OMPI_COMM_WORLD_LOCAL_SIZE", "SLURM_TASKS_PER_NODE"],
+)
+@pytest.mark.parametrize(
+    "group_ranks, expected",
+    [([0, 1, 2, 3], True), ([0, 1, 4, 5], False)],
+)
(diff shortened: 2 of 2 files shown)


tandede:
Hi maintainers, I wanted to follow up on this PR in case it was overlooked. The pre-commit check has passed, and the FP8 local-world-size environment handling fix is ready for review. I’d be happy to address any feedback or provide further validation. Thank you!

